PR #70, opened 2012-05-12
Management of AUTOSAR compiler abstraction macros + some additional corrections/updates

rdan:
Hi Ben,

I have updated the management of TYPE_WRAP to be able to use a multiple parameters macro that wrap a type. Seems to be compatible with old management based on the execution of tests.
I also added FUNC_PTR_WRAP to be able to declare "P2FUNC" specified by AUTOSAR.

There are also some additional updates/corrections that you will find in the commit description.

Regards,

Ronan


gmaurel:
Could you push a new version?

PR #74, opened 2012-05-17
New options to manage indentation of "do/while(0)" construct in macro function definition and to manage close parenthesis alignment

rdan:
Hi Ben,

Here is a new update of the tool: some new options and some corrections (everything is explained in the commit).

Ronan


gmaurel:
Hello Ronan!

Could please push a new version of your work against the actual git version?
I would appreciate more very little pushes.
I need also test sources, config files and expected output.
Thank

PR #83, opened 2012-06-24
extracted the very minor edits from my master branch for starters; this should be real easy to vet

GerHobbelt:
- typo fixes,
- typecasts to shut up compiler warnings,
- wrapped the 'uncrustify -h' output so it looks good on 80 column consoles
- two bugfixes:
  (1) quickfix: use Win32 ANSI APIs so one can compile the code in a MS Unicode project without errors
  (2) snprintf() isn't guaranteed to add a terminating NUL sentinel when the buffer overflows; the fix is to always put a NUL at the end of the buffer no matter what.
- consistently use 'static_inline' instead of 'static inline' (some compilers don't support 'inline') 
- some alignment whitespacing

(With these out of the way, a few more 'feature/fixes backports' should be easier to do.)


gmaurel:
And now?
Could you push a new version of your work against the actual git?
Thanks

gmaurel:
merged in #597

PR #166, opened 2013-05-09
Xcode updates

JanX2:


gmaurel:
I need a new version of your work against the actual git and some test sources with config files.

mihaipopescu:
I think this change is redundant as the Xcode project will be removed because it will be generated by the CMake scripts.

PR #204, opened 2013-10-11
Sorting of Objective-c properies, function prototypes and method declara...

darvin:
...tions


gmaurel:
Let me know about work against the actual version of git. Thanks

gmaurel:
No answer.

PR #216, opened 2013-11-10
Add support for add/remove newline before and after ObjC @interface/@implementation lines

ipmcc:
Add support for add/remove newline before and after ObjC @interface/@implementation lines. For example:

```
@interface Foo : NSObject {
    int ivar;
}
@end
```

…becomes…

```
@interface Foo : NSObject 
{
    int ivar;
}
@end
```

and things like that. Unit test coverage included.

Improve typing of chunks in ObjC @interface/@implementation lines.
Admittedly, ObjC being largely whitespace agnostic, there could have been many more options
(i.e. nl before/after class colon, category name, superclass, protocol list) but this is the
option I needed, so it's the one I built.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -4356,12 +4356,39 @@ static void handle_oc_class(chunk_t *pc)
             tmp->parent_type = CT_OC_CLASS;
          }
       }
+       if (tmp->type == CT_PAREN_OPEN)
+       {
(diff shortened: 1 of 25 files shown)


gmaurel:
Looks fine.
Could you push a new version against the actual git? Thanks

gmaurel:
No news.

PR #217, opened 2013-11-12
Adds support for sorting @property in obj-c

bslack:
Adds ability to sort property semantics. E.g.

@property (nonatomic, readonly, assign) NSString *foo;
@property (readonly, assign, nonatomic) NSString *bar;

can become:

@property (nonatomic, readonly, assign) NSString *foo;
@property (nonatomic, readonly, assign) NSString *bar;

The implementation is more convoluted then I had hoped/expected. The setter=foobar and getter=barfoo property specifiers messed up my initial implementation a bit and needed to be modified.

Control over the order that items will appear in is controlled via weights. The heigher the weight, the closer to the front it will be.

Don't expect this code to necessarily be accepted, but figured I would make the pull request so its a bit more public and others might pick it up / use it if they want.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -4948,7 +4948,10 @@ static void handle_oc_message_send(chunk_t *os)
  */
 static void handle_oc_property_decl(chunk_t *os)
 {
-    if (cpd.settings[UO_mod_sort_oc_properties].b)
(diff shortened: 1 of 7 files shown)


gmaurel:
Well, Xmas holidays is over.
Could you complete your work with an actual version against actual git, please with test sources and config files. Thanks

Issue #299, opened 2014-10-19
Newline after Label?


schancel:
```D
FOO:
	i++;
```

is being transformed into:
```D
FOO: i++;
```

I'm not sure if this is a bug, or if there's a way to fix it so that it adds a newline and indents `i` properly.

schancel:
I just tried the option in PR #316.  It fixes the mentioned problem, but causes others for D code.

Template specialization in is expressions:
```D
-                               } else static if (is (type: AstExpression)) {
+                               } else static if (is (type:
+                                                     AstExpression)) {
```

Explicit symbol importing:
```D
-               import d.ir.dscope: OverloadSet;
+               import d.ir.dscope:
+                      OverloadSet;
```

blakemcbride:
I too would like to see a nl_after_label option for my C/Java code.

gmaurel:
The actual version works pretty fine.
Please test it!

PR #316, opened 2015-01-06
Add nl_after_label_colon option

npnth:
This adds the nl_after_label_colon option, allowing

    int main(void)
    {
            if (f() < 0) goto cleanup;
            if (g() < 0) goto cleanup;
     cleanup: free_everything();
            return 0;
    }

to become

    int
    main(void)
    {
            if (f() < 0)
                    goto cleanup;
    
            if (g() < 0)
                    goto cleanup;
    
    cleanup:
            free_everything();
    
            return 0;
    }

Additionally, the slightly-pathological case of a label not
preceded by a newline (as, perhaps, in `fail: cleanup: done:` ?)
is now handled.

Test cases 01070 and 01071 are added to test this behavior. No
other tests required modification.

I believe this incidentally addresses issue #299 .


bengardner:
Looks good. Thanks.

PR #329, opened 2015-02-11
Keep trailing space in C++ comments if it follows backslash to prevent '\' from turning into line continuation

mikekov:
To avoid surprises when running uncrustify on large codebases I want to to keep following lines intact:

// some path: c:\dir\ 
// start block \/
// end block /\ 

Note that there's a trailing space in two comments. This space keeps '\' from acting as a line continuation.
If space is stripped, line following the comment is swallowed (turns into a comment too).

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -369,7 +369,12 @@ void output_text(FILE *pfile)
       }
       else if (pc->type == CT_COMMENT_CPP)
       {
+         bool tmp = cpd.output_trailspace;
+         // keep trailing spaces if they are still present in a chunk;
+         // note that tokenize() already strips spaces in comments, so if they made it up to here, they are to stay
+         cpd.output_trailspace = true;
          pc = output_comment_cpp(pc);
+         cpd.output_trailspace = tmp;
       }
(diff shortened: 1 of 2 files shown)


bengardner:
I'm sure you are aware that that code is already broken with most compilers.
This won't compile in GCC or clang, even with the space after the backslash.

```C++
void foo()
{
   // some path: c:\dir\ 
   int i;
   i = 3;
}
```

mikekov:
I see. Well, this is not the case with MS VS201x C++ compilers. Documentation typically talks about ending line with a backslash, but is silent about spaces. To the GNU's credit, they are calling this out:
"If there is white space between a backslash and the end of a line, that is still a continued line. However, as this is usually the result of an editing mistake, and many compilers will not accept it as a continued line, GCC will warn you about it."
I don't know about "many compilers", but certainly MS compilers behave like that, and it is of importance to me to keep meaning of this code unchanged, even if it is not portable to begin with.
Can you suggest some better approach, short of modifying my source code files?
I guess even with GCC there would be no harm in leaving such trailing space.

PR #330, opened 2015-02-12
Implementing code on Windows to deal with relative file paths from an 'include' in a config file

scottbilas:


bengardner:
Thanks.

PR #336, opened 2015-02-13
Add a new option for indenting continued shift expressions, fixes issue #314

### LINKED ISSUE #314: Indentation in multiline stream operators
Hello,

    I am using uncrustify v0.60. I would like to reformat my C++ code
so that operator continuation lines are always indented by 4
additional spaces (using "." to illustrate a single space character):

first-line-of-operator
....next-line-of-operator
....next-line-of-operator
....next-line-of-operator
....next-line-of-operator;

There are several cases where I cannot find the right combination of
uncrustify options to get the above formatting. I will detail these
cases below. Is there a way to do what I want with uncrustify?

All of the examples below use multiple left/right shift operations (<<
and >>). Each child context should be indented 4 spaces relative to
the parent context. I found three options relevant to my needs:

  indent_columns  = 4
  indent_continue = 4
  align_left_shift = true

But "align" is not what I need because alignment is dependent on the
first line of the context being formatted, but I want to use 4-space
indentation regardless of the first line structure.

* Case #1 (with a solution):

std::out <<
...."hello " << "world " <<
...."and " <<
...."all " <<
...."others" << ";" << std:endl;

I can get the above with

  indent_columns = 4
  align_left_shift = true

It works OK, but the format does not depend on "indent_continue" and
depends on "align_left_shift". It seems strange that indention depends
on alignment.

* Case #2 (no solution):

std::out <<
...."hello " << "world " <<
(who ? "and " : "or ") <<
...."all " <<
...."others" << ";" << std:endl;

or, to simplify further:

std::out <<
...."hello " << "world " <<
("and ") <<
...."all " <<
...."others" << ";" << std:endl;

I cannot find a way to tell uncrustify to indent the line with
parenthesis in the above examples.

* Case #3 (no solution):

std::out << "hello "
.........<< "world "
.........<< (who ? "and " : "or ")
.........<< "all "
.........<< "others" << ";" << std : endl;

Uncrustify aligns (with the << on the first line) while I want it to indent.

* Case #4 (no solution):

std::in >> a
\>> b
\>> c;

No indentation or even alignment for the right-shift operator.

Is there a way to format all of the above cases to look the way I want?

clbr:
Fixes #314. Including both a new more extensive test case
as well as an adapted existing test case.

Signed-off-by: Lauri Kasanen <curaga@operamail.com>

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -476,6 +476,7 @@ void indent_text(void)
    bool               did_newline = true;
    int                idx;
    int                vardefcol   = 0;
+   int                shiftcontcol = 0;
    int                indent_size = cpd.settings[UO_indent_columns].n;
    int                tmp;
    struct parse_frame frm;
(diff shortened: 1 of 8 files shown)


clbr:
I found some issues, will revise.

PR #338, opened 2015-02-14
Introduced new options indent_min_vbrace_open and indent_vbrace_open_on_...

nicodecker:
...tabstop.

The new options allow to apply special indent after virtual braces and
newline.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -909,7 +909,12 @@ void indent_text(void)
          frm.level++;
          indent_pse_push(frm, pc);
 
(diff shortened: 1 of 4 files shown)


bengardner:
Would you please provied a test case for this?  Without a test case, features may get broken without notice.

nicodecker:
I'll try; did not check out the files for implementing a test, yet. Only saw at first glance that it's based on python, which I am not familiar in. Somehow I will figure it out most likely, though.

PR #345, opened 2015-02-18
Added package.json for npm

shazron:
http://npmjs.org

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "uncrustify",
+  "version": "0.6.1",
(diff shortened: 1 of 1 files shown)


bengardner:
Does anyone use uncrustify with javascript?

shazron:
Thanks! Not directly. But we at the Apache Cordova project (http://cordova.apache.org) actually do use it for development, and our tooling is based on node.js and npm and this makes it easier to get dev and CI environments set up for all our contributors and committers.

PR #346, opened 2015-02-23
allow indent_ctor_init to be negative

4picht:
Hi,

for my style of constructor code formatting, I need to set indent_ctor_init to a negative value. Please merge this change to support this.

Thanks,
Julian


bengardner:
This change doesn't break anything, but it doesn't have any test coverage.

Would you please add a test that shows what a negative value is supposed to do?

PR #348, opened 2015-02-25
Fix: C++ bitfields were being detected in structs, but not in classes

scottbilas:


scottbilas:
I forgot to run unit tests. This breaks 30730.

PR #350, opened 2015-02-26
Implementing support for MS's C++ extensions for C++/CLI and WinRT on the ^ character

scottbilas:
This change is to let Uncrustify differentiate the ^ as a reference markup (similar to * for pointers).

Not super sure that the way I did this is right, or the naming of "msref". In particular, when/how to test against LANG_CPP, and the new utility functions I added..


src/align_stack.cpp
```diff
@@ -236,8 +235,7 @@ void AlignStack::Add(chunk_t *start, int seqnum)
       {
          tmp = chunk_get_next(tmp);
       }
-      if ((chunk_is_star(tmp) && (m_star_style == SS_DANGLE)) ||
-          (chunk_is_addr(tmp) && (m_amp_style == SS_DANGLE)))
+      if (chunk_is_ptr_operator(tmp) && (m_star_style == SS_DANGLE))
```
bengardner: This looks like it changes the logic.  m_star_style and m_amp_style are different.
You likely want '^' to follow m_start_style.  Same for the chunk below.

src/combine.cpp
```diff
@@ -863,6 +863,10 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
       {
          set_chunk_type(pc, (prev->type == CT_ANGLE_CLOSE) ? CT_PTR_TYPE : CT_DEREF);
       }
+      if ((cpd.lang_flags & LANG_CPP) && (pc->type == CT_CARET) && (prev->type == CT_ANGLE_CLOSE))
```
bengardner: Note sure if this is correct.  You are checking for ">^", but I don't see that combo in the new tests.
Did you mean to check for "^>"?

src/align_stack.cpp
scottbilas: Derp, missed that one, thanks.

src/combine.cpp
scottbilas: You're right, I should have tested this specific case too. I'll add a couple more items to the 31730.

bengardner:
Merged. Thanks.

PR #356, opened 2015-03-23
Fix ObjC 'Methodcall in ternary expression'

### LINKED ISSUE #355: ternary operator (? : ) with method call in Objective C in 0.61: unexpected colon in col 69 n-parent=NONE c-parent=OC_MSG l=1 bl=1
Hi, 

i've stumbled over this issue with some special method calls in a ternary operator in ObjC:

This line (I know, its nonsensical, but it is valid ObjC and was the simplest line I could come up with that still reproduces the issue):

```ObjC
NSString *string = @"A" ? [[NSString alloc] initWithString:nil] : @"C";
```
causes this error:

```
Demo.m:24 unexpected colon in col 69 n-parent=NONE c-parent=OC_MSG l=1 bl=1
```

However, this line: 

```ObjC
NSString *something = @"A" ? [[NSString alloc] init] : @"C";
```

works fine. After finding a related Issue (see below), I think it is related to the `:` in the method call.

[Here is a gist with a simple complete file that reproduces the error](https://gist.github.com/iv-mexx/4af8b9c85d99296909d4) (I was initially gonna report this error in the BBUncrustify repo but then found out it was an error in uncrustify itself). The error occurs as soon as line 23 is uncommented. 

Also, here is my [uncrustify.cfg](https://gist.github.com/iv-mexx/8a3a924860d1bc36fbe1) in case its important.

This issue occurs with uncrustify 0.61 and still with commit cd1c8619321efc96afc1ab1e7d09e924f008acaa on master.

I've searched the issues and found two similar issues, but they are a little bit different:

* #319 Occurs also with method call in ternary operator, but seems to work (but incorrectly)
* #23 Occurs with ternary operator and returns a similar error message, but occurs in Java

iv-mexx:
This PR fixes #355 

In the `combine_labels` function, every colon used to reduce the `question_count`. This is wrong if the colon occurs in an ObjC method call because in that case, all parameters of the method are preceded by a colon that is not part of the ternary expression.

I have extended the ObjC `ternary.m` test input to cover this case. 

All tests (including the new case) still pass: 

```
mexx@mexxbook tests (feature/fixObjCTernaryMethodCall) $ ./run_tests.py
Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 509 / 509 tests
```

All the UNSTABLE test cases have already been UNSTABLE before my changes.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -16,6 +16,7 @@
 #include <cerrno>
 #include "unc_ctype.h"
 #include <cassert>
+#include <stack>
 
 static void fix_fcn_def_params(chunk_t *pc);
 static void fix_typedef(chunk_t *pc);
@@ -2237,9 +2238,12 @@ void combine_labels(void)
    chunk_t *prev;
    chunk_t *next;
    chunk_t *tmp;
-   int     question_count = 0;
    bool    hit_case       = false;
    bool    hit_class      = false;
+
+   // need a stack to handle nesting inside of OC messages, which reset the scope
+   std::stack<int> question_counts;
+   question_counts.push(0);
 
    prev = chunk_get_head();
    cur  = chunk_get_next_nc(prev);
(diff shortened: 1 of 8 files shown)


scottbilas:
Ok please see https://github.com/bengardner/uncrustify/pull/357. I really don't work with Obj-C very much so I wasn't able to come up with a good test case for nested messages and ternaries, but I added a case from our codebase that has a lot of extra semicolons in it, which exercises more of the problem.

bengardner:
Looks good. Thanks.

PR #357, opened 2015-03-24
Fix for SourceForge #635 GitHub #355

scottbilas:
This fixes https://sourceforge.net/p/uncrustify/bugs/635/ (also filed as https://github.com/bengardner/uncrustify/issues/355) and adds in a test case from https://github.com/bengardner/uncrustify/pull/356.

I believe what's necessary to deal with Obj-C messages is to introduce a scope stack that tracks the 'question count' per nesting level.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -2276,7 +2276,7 @@ void combine_labels(void)
          }
       }
       else if ((next->type == CT_COLON) ||
-               ((question_count > 0) && (next->type == CT_OC_COLON)))
+               ((question_count > 0) && (next->type == CT_OC_COLON) && (cur->type != CT_OC_MSG_FUNC)))
       {
          if (cur->type == CT_DEFAULT)
          {
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -16,6 +16,7 @@
 #include <cerrno>
 #include "unc_ctype.h"
 #include <cassert>
+#include <stack>
 
 static void fix_fcn_def_params(chunk_t *pc);
 static void fix_typedef(chunk_t *pc);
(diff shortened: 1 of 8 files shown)


iv-mexx:
I can try to come up with a more complicated test case that contains more colons in the ternary expression when i get home if you want to.

scottbilas:
That would be a big help, yeah. I'm not super confident I got all the cases with this stack solution.

At the top of tests/input/oc/oc_cond_colon.m is an example of a ternary inside of a message. I'm thinking we could also use a sample that combines that with nested messages. And, what the hell, combined with an outer ternary that is outside all messages. Probably very rare case but I think it would put it to bed.

iv-mexx:
Ok, I didn't know how to do this properly, so I've merged your changes into my pull request and added another commit a0086de8b5e65c828c4a45058ea040e6d9836634 with a new testcase. I see you've already picked one of my commits so I think you can just picke this one as well and add it to this merge request.

The new test case is absolutely insane and i guess something like that should not be written in real code :laughing: 

Its one ternary operator with function calls in both cases. The first one has two parameters (so two colons that would have caused problems previously) and also the parameters are dictionaries that again contain colons to separate keys and values in the dictionary. For good measure, one of the dictionary values is again a method call with a parameter. 
The other even has a ternary operator for the parameter of the method.

Still, this thing is valid ObjC Code, I have complied and tested it, it actually does something when the called methods are implemented.

I've also tired this test case with the fix from my PR and as you have predicted, my fix does not cover this case

With your fix the new test case passes, so good job! :+1: 


The Result of tests/output/oc/50016-ternary.m is (in my opinion) a little bit unexpected, i think `}) }` should not have been moved to a separate line. But maybe thats beyond the scope of this PR anyway.

PR #358, opened 2015-03-27
correct the function keywords_are_sorted

gmaurel:
1. function type is bool, but the returned value is never use at uncrustify.cpp:249
2. if an "bad sort order at idx" is found, it must stop after LOG_FMT.


gmaurel:
otherwise done.

PR #370, opened 2015-04-24
Qt translator meta data comments

Lykurg:
Qt Translator and meta-data comments in CPP style like '//:', '//=', '//~' 
where not handled properly when spaces were to be added with option 
sp_cmt_cpp_start. 

For this kind of comments http://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments.

The new option sp_cmt_cpp_qttr treats those comment leadings as a unity,
if set to TRUE.

Before sp_cmt_cpp_start=add converted '//:foo' to '// :foo'. Now it will
correctly be converted to '//: foo'.

This is basically an addition to sp_cmt_cpp_doxygen.


bengardner:
I cherry picked your 2nd commit in 15ee014c5261a26d46d3479f0f7330db719f183a.

PR #375, opened 2015-05-15
Commit vs2015 upgrade of the solution file

wjwwood:
So this is the third time the Windows machines has decided to self destruct with this license expiration:

http://54.183.26.131:8080/job/ros2_batch_ci_windows/240/console
```
         Microsoft Visual Studio 2015 RC Version 14.0.22823.1.
         Copyright (C) Microsoft Corp. All rights reserved.
         
         The license for Visual Studio has expired.
         
         The evaluation period for this product has ended.
```

This only happens when Jenkins is running the job. I cannot reproduce the problem when using `cmd.exe` from the desktop. I've tried:

- Running the Jenkins slave as my user.
- Force setting the `USERNAME` env variable to `osrf`.
- Getting a new developer license (https://msdn.microsoft.com/en-us/library/windows/apps/hh974578.aspx)
- Repairing the VS2015 installation.

None of these have worked. That last time I fixed this issue by reinstalling Windows...

The weird thing is that this failure only occurs when `uncrustify` runs `devenv` to upgrade the solution file, and not when invoking msbuild or compiling/linking. With this pull request I've committed that upgrade to the solution file and disabled the upgrade step in CMake. This at least allows the build to continue until it gets to the idl parser error https://github.com/ros2/rcl_interfaces/issues/2, see:

http://54.183.26.131:8080/job/ros2_batch_ci_windows/241/console

If I can't come up with another solution then this will have to be the work around for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,50 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(uncrustify)
+
+include(ExternalProject)
+
+if(NOT WIN32)
+  ExternalProject_Add(
+    external_uncrustify
+    SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}"
+    CONFIGURE_COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/configure"
+      "--prefix" "${CMAKE_CURRENT_BINARY_DIR}"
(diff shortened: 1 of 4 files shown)


wjwwood:
This was meant to be against our fork, sorry for the noise...

PR #380, opened 2015-06-02
Use the cmake build type

wjwwood:
This depends on https://github.com/ament/ament_tools/pull/33

By doing this we avoid some warnings about this package not using certain ament specific CMake variables.


wjwwood:
Sorry, did it again, wrong fork. I apologize for the noise...

PR #381, opened 2015-06-29
Add indent_param parameter to use as continuation indent

lasconic:
Add indent_param parameter to use as continuation indent if func_*_param are true.

The style guidelines of [MuseScore](http://github.com/musescore/MuseSCore) requires a continuation indent of 3 while the normal indent is 6. I couldn't find a way to achieve this so I added a new option.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1369,7 +1369,14 @@ void indent_text(void)
                idx--;
                skipped = true;
             }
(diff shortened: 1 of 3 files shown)


bengardner:
Would you please add a test case for this new option?

gmaurel:
Looks very fine.
Could you push some configuration and tests files with the expected output to your work?

guy-maurel:
PR #980 provides test for this proposal

gmaurel:
PR #980 implement it.

PR #412, opened 2015-11-21
nl_oc_msg_args behaviour inconsistent for dot-syntax in receiver

sobeski:
Fixes issues described in:
https://github.com/bengardner/uncrustify/issues/125


gmaurel:
Could you push some sources for the test and the expected output.
I would like to expand the tests with your work. Thanks.

gmaurel:
I asked for:
Could you push some sources for the test and the expected output.
I would like to expand the tests with your work. Thanks.

gmaurel:
Thanks

PR #416, opened 2015-12-29
Fix issue 414: The actual global spacing options must not to be applicated to the preprocessor

evandrocoan:
Fix issue 414: The actual global spacing options must not to be applicated
to the preprocessor

https://github.com/bengardner/uncrustify/issues/414

This fix is applicable/activated if the ASCII value of the string escape char 94
is selected.

It remove spaces only at the preprocessor macro functions declarations and
ignore adding arithmetic expressions spaces at the macro
function body.
Ex: 'define COOL_MACRO( % 1 , % 2 ) ( % 1 = % 2 + 1 )', which is wrong.
#instead of: '#define COOL_MACRO(%1,%2) ( %1 = %2 + 1 )' 

To accomplish it, this overrides sp_inside_paren, sp_after_comma and sp_arith
only at the macro function declaration and body.

This also fix an unreported issue where the "#emit" preprocessor directive was
receiving a space after a DOT.
Ex: '#emit CONST .pri 1911' or '#emit CONST. pri 1911', which is wrong.
instead of: '#emit CONST.pri 1911', the correct one.

Also updated the "amxmodx.cfg" config accordingly with the correct and up-to-date configuration.

This are the new test results:
Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 517 / 517 tests

This are the tests results before this work:
Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 516 / 516 tests

### DIFF
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -1215,6 +1215,24 @@ static bool parse_newline(tok_ctx& ctx)
 }
 
 
+/**
+ * PAWN #define is different than C/C++.
+ *   #define PATTERN REPLACEMENT_TEXT
+ * The PATTERN may not contain a space or '[' or ']'.
+ * A generic whitespace check should be good enough.
+ * Do not change the pattern.
+ */
+static void parse_pawn_pattern(tok_ctx& ctx, chunk_t& pc, c_token_t tt)
+{
+   pc.str.clear();
+   pc.type = tt;
+   while (!unc_isspace(ctx.peek()))
+   {
+      pc.str.append(ctx.get());
+   }
+}
+
+
 static bool parse_ignored(tok_ctx& ctx, chunk_t& pc)
 {
    int nl_count = 0;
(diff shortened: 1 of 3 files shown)


tests/input/pawn/pawn_amxmodx.sma
```diff
@@ -0,0 +1,35 @@
+#include <amxmodx>
+
+#define COOL_MACRO( %1 , %2 ) ( %1 + %2 + 1 )
+
+#define SXO( % 1 ,%2) 1 < %1 || %2 > 2
```
bengardner: As "% 1" (with the space) is invalid and won't compile, then I'm against "fixing" it.  In other words, it should be "% 1" in the output, too.

addonszz: So, just do not add spaces?

Issue #419, opened 2016-01-07
The mod_full_brace_function is putting braces inside the preprocessor and broken the code


addonszz:
The mod_full_brace_function is putting braces inside the preprocessor and broken the code. The option description is : 

#Add or remove braces on single-line function definitions. (Pawn)
```pawn
#if AMXX_VERSION_NUM < 183
public client_disconnect( player_id )
#else
public client_disconnected( player_id )
#endif
{

}
```
Became this, which is broken code:
```pawn
#if AMXX_VERSION_NUM < 183
public client_disconnect( player_id )
#{ \
    else }

public client_disconnected( player_id )
#{ \
    endif }

{
}
```
He is my debug file: [debug.txt](https://github.com/bengardner/uncrustify/files/81851/debug.txt)

gmaurel:
Could you push the output you expected to get?

bengardner:
This has been fixed.

PR #420, opened 2016-01-07
Fix issue 419: The mod_full_brace_function is putting braces inside the preprocessor and broken the code

addonszz:
Fixed issue #419 The mod_full_brace_function is putting braces
inside the preprocessor and broken the code:

https://github.com/bengardner/uncrustify/issues/419

This also fix the following bad variable definition:

The multidimensional array pointer "pcf_names" is protected by an "#ifdef DEFINE_PCF_NAMES"
which is never defined for use, but the array "pcf_names" is need at the main file.

Then, it was moved to the main file "uncrustify.cpp" without the unnecessary #ifdef tag directive.

This are the new test results:

Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 518 / 518 tests

This are the tests results before this work:

Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 516 / 516 tests


bengardner:
I checked the pawn spec and it looks like the "%1" format is a special macro variable.
Wouldn't it make better sense to pick that up in the tokenizer?
Perhaps around line 1472, where it does other PAWN stuff.

Pseudo logic would be something like:
```
if (in_preprocessor && (ctx.peek(0) == '%') && is_digit(ctx.peek(1)) {
   parse_word(ctx, pc, true);
   return(true);
}
```

What do you think?

addonszz:
Looks nice, latter on I see look into that and other things.

PR #423, opened 2016-01-18
Automatic language detection for stdin (--assume)

Lykurg:
While -f uses the simple automatic language detection with stdin
one is out of luck. With the new argument --assume one can specify
the filename Uncrustify should use to perform the automatic
language detection even for content pushed to stdin.

This argument is useful for IDEs when piping the content of an 
unsaved file to stdin. This way the result is equivalent to a normal 
command line usage on the saved file.


bengardner:
I'm not terribly fond of the option name "assume", but I can't think of anything better.

PR #431, opened 2016-02-10
Fixed issues 419 and 417

addonszz:
### Fixed issues https://github.com/bengardner/uncrustify/issues/419 and https://github.com/bengardner/uncrustify/issues/417

1. The mod_full_brace_function is putting braces inside the preprocessor and broken the code

2. Bad processing braces at '#if defined', '#else' preprocessor

To fix it was created two newer options:
```
# Disables the preprocessor source code 'mod_full_brace_...' options parsing.
pp_mod_full_brace_disable                 = true    # false/true

# Disables the preprocessor source code indent parsing.
pp_indent_parsing_disable                 = true    # false/true
```

##### This are the new test results:

Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 521 / 521 tests

##### This are the tests results before this work:

Processing c-sharp.test
Processing c.test
UNSTABLE: 00012
UNSTABLE: 00617
UNSTABLE: 02100
UNSTABLE: 02102
UNSTABLE: 02501
Processing cpp.test
UNSTABLE: 30265
UNSTABLE: 30815
Processing d.test
Processing java.test
Processing pawn.test
Processing objective-c.test
UNSTABLE: 50007
Processing vala.test
Processing ecma.test
Passed 516 / 516 tests

##### This also fix the following bad variable definition:

The multidimensional array pointer "pcf_names" is protected by an "#ifdef DEFINE_PCF_NAMES"
which is never defined for use, but the array "pcf_names" is need at the main file.

Then, it to keep related definitions next to each other in the same file, was added to the main file 'uncrustify.cpp':
```
#define DEFINE_PCF_NAMES
```
Before 'pcf_names' to be imported as the Eclipse IDE/Compiler, was unable to see it as 'DEFINE_PCF_NAMES' was never defined.


bengardner:
This pull request to too large and makes too many unrelated changes.
If you split individual changes into separate pull requests, then they are more likely to be reviewed and accepted.

src/keywords.cpp
```diff
@@ -1,6 +1,7 @@
 /**
  * @file keywords.cpp
- * Manages the table of keywords.
+ * Manages the table of keywords. This C file contains all language keywords accepted by this
```
bengardner: Why?

src/keywords.cpp
```diff
@@ -287,7 +288,10 @@ static int kw_compare(const void *p1, const void *p2)
    return(strcmp(t1->tag, t2->tag));
 }
 
-
+/**
+ * This function verifies if all language keywords as 'if', 'for', 'break', etc, are 
+ * alphabetycally sorted on the main table.
```
bengardner: Why?  And you add a spelling error.

src/logger.h
```diff
@@ -19,6 +19,13 @@
 #include <cstdio>      /* FILE */
 
 
+#define IS_DEBUG_ENABLED 1
+
+#if IS_DEBUG_ENABLED > 0
```
bengardner: Search on the right way to do 'DEBUG' with autotools.
I used something like:
```
./configure CXXFLAGS="-ggdb3 -O0 -DDEBUG" CXX="ccache g++"
```

src/uncrustify.cpp
```diff
@@ -38,10 +38,55 @@
 #include <vector>
 #include <deque>
 
+
+static const char *pcf_names[] =
```
bengardner: I don't like this move.
I prefer to have related definitions next to each other in the same file. Putting them in separate files increases the likelyhood of them getting out of sync.
I would be OK with an automated script that creates the pcf_names array based on the PCF macros.
That is done for other items.

src/uncrustify.cpp
```diff
@@ -181,10 +226,11 @@ static void usage_exit(const char *msg, const char *argv0, int code)
            "                            Detection is fairly limited.\n"
            "\n"
            "Debug Options:\n"
-           " -p FILE      : dump debug info to a file\n"
-           " -L SEV       : Set the log severity (see log_levels.h)\n"
-           " -s           : Show the log severity in the logs\n"
-           " --decode     : decode remaining args (chunk flags) and exit\n"
+           " -p FILE          : dump debug info to a file\n"
+           " -L SEV1 SEV2 ... : Use 'a' for all levels. See the log severity availibles on\n"
```
bengardner: That isn't how the logsev works.
It is a comma-separated list of numbers or ranges of numbers.
Example:  -L0-3,45,56-58

src/braces.cpp
```diff
@@ -114,12 +113,11 @@ static void examine_braces(void)
...
-   chunk_t *prev;
-
-   pc = chunk_get_tail();
-   while (pc != NULL)
+   chunk_t *backup;
+   
+   for (pc = chunk_get_tail(); pc != NULL ; pc = backup)
    {
-      prev = chunk_get_prev_type(pc, CT_BRACE_OPEN, -1);
+      backup = chunk_get_prev_type(pc, CT_BRACE_OPEN, -1);
```
bengardner: Why rename the variable?

addonszz: Sorry, no need.

src/keywords.cpp
addonszz: More information for doc.

src/keywords.cpp
addonszz: It is a documentation. It could seen obvious to you but not to everybody else. The misspelling is fixed.

src/logger.h
addonszz: I need to stop here, later on I to continue to fix.

addonszz: Sorry, I will updated it.

src/uncrustify.cpp
addonszz: I missed that, was only needed a '#define DEFINE_PCF_NAMES' before import 'uncrustify_types.h', because the Eclipse IDE was not finding it any anywhere.

src/uncrustify.cpp
addonszz: Fixed.

addonszz:
Sorry, I just realized that too late. Do you still need it to be split over different multiple branches creating the possibility to create multiple pull requests?

PR #436, opened 2016-02-19
Add support for C# interpolated strings

bengardner:
Seems to be a MS extension. Can't find it in the C# language spec.


scottbilas:
C# 6.0, bleeding edge. Apparently they only have an official downloadable language spec for 5.0 so far. But this language version is live in Visual Studio 2015 right now.

https://msdn.microsoft.com/en-us/library/dn961160.aspx

PR #461, opened 2016-03-15
sync comments and fix incorrect key name

todd-richmond:


todd-richmond:
not sure how the CI check would fail on a config file change, but Travis details claims "cannot load config", but does not say which one and manual uncrustify runs against this full file do not produce any warnings

bengardner:
There was an upstream error -- a missing test config file. That has been fixed.
I don't know how to trigger a rebuild, yet.
You can ignore the CI failure for now.

PR #479, opened 2016-05-08
small simplifications

CDanU:
4#: I am not sure why keywords_are_sorted() is tested only when not enough args are available or why it is executed in a non debug context. 
Wouldn't it be better to move it into a ifdef debug guard or add it into an assert and disabling asserts in non debug mode ? 
What is your stance on the features of c++11 and c++14 ?  It maybe could be possible to make keywords_are_sorted() a constexpr which fails at compile time if something is wrong.

### DIFF
--- a/src/compat_posix.cpp
+++ b/src/compat_posix.cpp
@@ -26,11 +26,7 @@ bool unc_getenv(const char *name, std::string& str)
 
 bool unc_homedir(std::string& home)
 {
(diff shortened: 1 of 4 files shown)


gmaurel:
Thanks.

CDanU:
3031ead305e726f028f636b3d3c0c19ba64d1d9a could be problematic since the else branch that handles the hidden config file (.uncrustify) is missing

PR #480, opened 2016-05-13
win32: Add missing file 'options_for_QT.cpp' to build

scorpion007:
Self explanatory.


gmaurel:
Already included at now. Thanks

PR #492, opened 2016-05-22
Add CMake script

jibsen:
As discussed in #470, this is an attempt to add a CMake script that works on both Windows and Linux, and can be used to create build files for Visual Studio.

Since `tests/run_tests.py` assumes the executable is in `src`, I added a hack that copies the executable there after build so the unit tests run. It might be preferable if the script took the executable path as an argument.

Example use on Linux:

~~~.sh
mkdir build
cd build
cmake ..
make
make test
~~~

Example use on Windows with a Visual C++ command prompt:

~~~.cmd
mkdir build
cd build
cmake ..
cmake --build . --config Debug
ctest -C Debug
~~~

PR #494, opened 2016-05-27
Fix testing for CMake builds

jibsen:
The hardcoded path in #485 broke testing with CMake on Windows (but great to have Python 3 compatibility!).

As mentioned in #492, I added a different hack in the CMake script to make testing work, but I believe a better solution is to allow the test script to take the path to the uncrustify executable it is supposed to test as an optional argument.

This PR adds a command line argument `--exe` which the CMake script uses to pass along the path to the uncrustify executable to test. For the Visual C++ solution files in `win32` and non-Windows builds, it defaults to the same paths as before.

I am using `argparse` which requires Python 2.7+ -- I don't know if that is a problem?

I also converted the indentation to 4 spaces (as per PEP 8), because the file contained a mixture of tabs and spaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,13 +222,10 @@ add_executable(uncrustify
 #
 if(BUILD_TESTING)
   find_package(PythonInterp REQUIRED)
-  add_custom_command(TARGET uncrustify POST_BUILD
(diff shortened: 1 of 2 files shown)


jibsen:
I updated the PR to make `run_tests.py` work if there are spaces in the path to the uncrustify folder.

PR #499, opened 2016-05-31
fix compiler warnings

dirk-thomas:
This patch fixes the compiler warnings which became visible by the added compiler flags from https://github.com/uncrustify/uncrustify/pull/492/files#diff-af3b638bc2a3e6c650974192a53c7291R18.

I tried to maintain current behavior of the code.


dirk-thomas:
The second commit suppresses the remaining compiler warning on Windows.

PR #505, opened 2016-06-02
c#: indent_cs_delegate_brace should also apply to keyword delegate

tpokorra:
the initial commit only worked for delegates with lambda expression


tpokorra:
I noticed there are more things missing, therefore withdrawing the pull request

PR #506, opened 2016-06-03
c#: indent_cs_delegate_brace should also apply to keyword delegate

tpokorra:
the initial patch for bug sf#674 only worked for delegates with lambda expression


tpokorra:
withdrawing because I realized I need it differently, with ignoring option, to keep previous mixed behaviour

PR #515, opened 2016-06-15
Added sp_func_[proto,def,class]_paren_empty

sjakub:
There is a pair of options:  sp_func_call_paren and sp_func_call_paren_empty
that control the spacing between 'func ( foo )' and 'func()' when calling,
but there are no options for declaring and defining functions/constructors without arguments.
This adds sp_func_proto_paren_empty, sp_func_def_paren_empty,
and sp_func_class_paren_empty.


mihaipopescu:
I think the changes look good but can you give me a concrete example why one would want to have different rules when the function has zero or more arguments ?

Also you should add a test for these new options.

sjakub:
There is a way to make function calls look like this: "fooBar ( x );" and "fooBar();" - spaces before and inside '( )' when there are arguments, and no spaces at all if there are no arguments. I like that style, because it has a lot of spaces when there is "something interesting" going on (when the arguments are passed), and skips those spaces when that's a simple, argument-less function call. However, it's not possible to achieve the same formatting for function declarations and definitions (and constructors/destructors, since they are handled separately). At the very least, this makes things inconsistent...

As for the tests, I'll have to figure out how to do that and then I'll update this change :)

sjakub:
I added tests to verify this functionality. I used the test from the change that added 'sp_func_call_paren_empty' support as a base.

mihaipopescu:
Tests LGTM even though I would have used `include base.cfg` instead of copy/paste because it's hard to follow which options are new/overriden etc.

sjakub:
I simplified the configs and updated the test outputs.
Now the new options are used to force brackets in those "empty" cases instead of removing them.
This way it's easier to see in the results/output file what exactly each option does.

sjakub:
I just realized that I should have probably used a separate branch in my fork for that feature, so I closed this request, and created a new one.

PR #522, opened 2016-06-21
Added sp_func_[proto,def,class]_paren_empty

sjakub:
There is a pair of options: sp_func_call_paren and sp_func_call_paren_empty
that control the spacing between 'func ( foo )' and 'func()' when calling,
but there are no options for declaring and defining functions/constructors without arguments.
This adds sp_func_proto_paren_empty, sp_func_def_paren_empty,
and sp_func_class_paren_empty.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -366,8 +366,12 @@ void register_options(void)
                   "A minimum of 1 is forced except for pointer return types.");
    unc_add_option("sp_func_proto_paren", UO_sp_func_proto_paren, AT_IARF,
                   "Add or remove space between function name and '(' on function declaration");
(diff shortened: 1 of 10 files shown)


sjakub:
There is a way to make function calls look like this: "fooBar ( x );" and "fooBar();" - spaces before and inside '( )' when there are arguments, and no spaces at all if there are no arguments. I like that style, because it has a lot of spaces when there is "something interesting" going on (when the arguments are passed), and skips those spaces when that's a simple, argument-less function call. However, it's not possible to achieve the same formatting for function declarations and definitions (and constructors/destructors, since they are handled separately). At the very least, this makes things inconsistent...

Issue #523, opened 2016-06-21
Space between '> >' in templates gets eaten inside macros


sjakub:
Current git version, using default config.
This input:
```
#define MACRO(templ_type) template <typename T> class Abc<templ_type<T> > { }
template<typename T> class Foo<Bar<T> > { };
```
Is converted to this:
```
#define MACRO(templ_type) template <typename T> class Abc<templ_type<T>> { }
template<typename T> class Foo<Bar<T> > { };
```

It looks like in the actual template, the space between '> >' got preserved, which is correct given the default settings. However, when the same thing happens inside a template, that space gets removed...

With macros like that I also saw that '>>' sometimes gets moved to the next line,
when pos_arith is set to lead. Which means it's not recognized correctly as a part of a template.

sjakub:
It looks like there is a lot of issues related to macros....
Is there a way to make it not format macros at all?

sjakub:
And it looks like this has been broken by the same change 12d474f introduced by @gmaurel that broke #509 ... undoing changes from 12d474f fixes this bug.

sjakub:
I created a pull request that reverts the original behaviour from before the 12d474f
and fixes the original bug in a different way, that doesn't cause issues with templates.

gmaurel:
Please test the actual git version and report.
Thanks for helping.

sjakub:
Just tested it: It works as expected using the current master.

PR #524, opened 2016-06-22
Fixed template handling inside macros

### LINKED ISSUE #509: Error: Unexpected '>' for 'PAREN_OPEN', which was on line
c:\tmp>C:\Users\Ingres\Downloads\uncrustify_v063.exe -v
    uncrustify 0.63.0-g44ce0f1

    c:\tmp>wget http://uncrustify.sourceforge.net/default.cfg

    c:\tmp>C:\Users\Ingres\Downloads\uncrustify_v063.exe -c uncrustify_v063.cfg  bug_form3.h
    Output suffix: .uncrustify
    Parsing: bug_form3.h as language CPP
    bug_form3.h:10 Error: Unexpected '>' for 'PAREN_OPEN', which was on line 10

Where:

    c:\tmp>type bug_form3.h
    #if defined(_MSC_VER)
        #if _MSC_VER < 1300
            #define __func__ "unknown function"
        #else
            #define __func__ __FUNCTION__
        #endif /* _MSC_VER < 1300 */
    #endif /* defined(_MSC_VER) */

    #define bug_demo        (1 > 2) ? (1 : 2)

* If the #if directives are removed, no errors.
* this appears to be benign, the fixed generated code is fine - the "error" is misleading

sjakub:
This change reverts change 12d474f that introduced
errors (#509 #523), as well as ad23416 that fixed #509, but not #523.

It also fixes the original bug (663 on sourceforge)
in a different way (that doesn't cause those issues).

### DIFF
--- a/src/tokenize_cleanup.cpp
+++ b/src/tokenize_cleanup.cpp
@@ -46,17 +46,7 @@ static chunk_t *handle_double_angle_close(chunk_t *pc)
       }
       else
       {
-         // bug #663
-         if ((pc->flags & PCF_IN_PREPROC) &&
-             (pc->flags & PCF_IN_TEMPLATE))
-         {
-            log_pcf_flags(LGUY, pc->flags);
-            // no change
-         }
-         else
-         {
-            set_chunk_type(pc, CT_COMPARE);
-         }
+         set_chunk_type(pc, CT_COMPARE);
       }
    }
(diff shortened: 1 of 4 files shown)


gmaurel:
Thanks

gmaurel:
Thamks

PR #532, opened 2016-06-28
Added mod_full_brace_if_chain_only option

sjakub:
Instead of adding or removing braces from all if/elseif/else chains that mod_full_brace_if_chain does,
the new option simply adds braces to all if-else statements with at least one 'else' or 'else if'.
It will leave simple unbraced 'if ( foo ) bar();' statements alone, but make all if-else chains fully braced.

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -30,7 +30,8 @@ static bool should_add_braces(chunk_t *vbopen);
 void do_braces(void)
 {
    LOG_FUNC_ENTRY();
-   if (cpd.settings[UO_mod_full_brace_if_chain].b)
+   if (cpd.settings[UO_mod_full_brace_if_chain].b
+       || cpd.settings[UO_mod_full_brace_if_chain_only].b )
    {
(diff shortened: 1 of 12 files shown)


src/braces.cpp
```diff
@@ -1103,6 +1104,13 @@ static void process_if_chain(chunk_t *br_start)
       {
          break;
       }
+
+      if (cpd.settings[UO_mod_full_brace_if_chain_only].b)
```
mihaipopescu: Why haven't you placed it outside the loop if it depends only on the new option ?

mihaipopescu: wait, I think it is ok.

sjakub: We need that loop to go through the entire if-else chain. We hit this point when we see an 'else', and this is the sign that this is not a simple 'if(a) b;' statement, so we want to add braces.

mihaipopescu: Yeah. I think the code in that area is a bit hard to _read_.

PR #537, opened 2016-06-29
Fixing issues with 'auto'

sjakub:
'Auto' was being treated as a qualifier,
then it was changed to be treated as a separate token type
to fix a bug which, in turn, introduced other bugs.

This reverts 502b5e1 - removes CT_AUTO, but also treats 'auto' as a type, not a qualifier.

It fixes both the original and the new bugs.


mihaipopescu:
This PR lacks tests. You could add my tests from https://github.com/Unity-Technologies/uncrustify/commits/auto-keyword-fix if you like.

pravala-jakub:
Done

sjakub:
Done

PR #549, opened 2016-07-07
Make the "indent to token following a brace" optional

pdh11:
Hi, first time pull-requester here, so please explain if I've done any of this wrong.

Full explanation in the commit message, but basically a brace that's followed by something not a newline, indents the block's entire contents to be level with the "something". The motivation appears in the provided test-case of the new option, where (without it) the class would be shoved 13-odd spaces to the right, to align it with the `namespace stm32`.

### DIFF
--- a/tests/input/cpp/namespace_namespace.cpp
+++ b/tests/input/cpp/namespace_namespace.cpp
@@ -1,0 +1,6 @@
+namespace hw { namespace stm32 {
+
+class RTC {
+}
+
+}} // namespace hw::stm32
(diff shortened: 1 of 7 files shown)


tests/input/cpp/namespace_namespace.cpp
```diff
@@ -0,0 +1,6 @@
+namespace hw { namespace stm32 {
+
+class RTC {
+}
```
mihaipopescu: missing semicolon here even if I don't think it'll change the formatting but it's good to test with valid C++.

PR #556, opened 2016-07-11
Allow Universal Character Names in identifiers

### LINKED ISSUE #555: [C++] Unable to parse UCN identifiers
Uncrustify is not able to handle `Universal Character Names` identifiers:

```cpp
int main()
{
   int IdentContainingTwoUCNCharacters\u1234\U00001234 = 0;
}
```

complaining about garbage in the console output:
```console
input/cpp/ucn.cpp:3 Garbage in col 40: 5c
input/cpp/ucn.cpp:3 Garbage in col 46: 5c
```

and generating illformed C++ code:
```cpp
int main()
{
	int IdentContainingTwoUCNCharacters \ u1234 \ U00001234 = 0;
}
```

They are somehow handled by other languages (like D) but as strings not as identifiers.

bengardner:
This handles:
    int IdentContainingTwoUCNCharacters\u1234\U00001234;

No validation of the escape sequence is done.
This change just allows '\u' and '\U' in the identifier.

Fixes #555

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -562,6 +562,9 @@ void indent_text(void)
    pc = chunk_get_head();
    while (pc != NULL)
    {
+      LOG_FMT(LGUY, "%d: [%d] pc->text() %s\n",
+              pc->orig_line, __LINE__, pc->text());
+      log_pcf_flags(LGUY, pc->flags);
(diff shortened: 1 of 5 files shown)


mihaipopescu:
A very elegant solution 👍

bengardner:
Rebased and replaced with #559

PR #559, opened 2016-07-11
Allow Universal Character Names in identifiers

### LINKED ISSUE #555: [C++] Unable to parse UCN identifiers

bengardner:
This handles:
    int IdentContainingTwoUCNCharacters\u1234\U00001234;

No validation of the escape sequence is done.
This change just allows '\u' and '\U' in the identifier.

Fixes #555

### DIFF
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -312,3 +312,5 @@
 #33080 multi_line_10.cfg                cpp/multi_line.cpp
 33081 bug_i_552.cfg                    cpp/bug_i_552.cpp
 33082 namespace_namespace.cfg          cpp/namespace_namespace.cpp
+
+33090 empty.cfg                        cpp/gh555.cpp
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -1153,10 +1153,23 @@ bool parse_word(tok_ctx& ctx, chunk_t& pc, bool skipcheck)
    pc.str.clear();
    pc.str.append(ctx.get());
 
-   while (ctx.more() && CharTable::IsKw2(ctx.peek()))
+   while (ctx.more())
    {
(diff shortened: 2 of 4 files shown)


tests/cpp.test
```diff
@@ -307,4 +307,4 @@
 33078 multi_line_8.cfg                 cpp/multi_line.cpp
 33079 multi_line_9.cfg                 cpp/multi_line.cpp
 33080 multi_line_10.cfg                cpp/multi_line.cpp
-33081 bug_i_552.cfg                    cpp/bug_i_552.cpp
+33082 empty.cfg                        cpp/gh555.cpp
```
mihaipopescu: This has to be `33081`.

bengardner: Commit 5c36767220a942f004b1b04b14181b1a231dc2c1 uses 33081.

mihaipopescu: right... a common problem of the test suite. Then we have to rename the output file to fix the mismatch.

bengardner: Oops.  I didn't notice that I didn't pull the new output file. Thanks.

PR #563, opened 2016-07-13
Fix identifier parsing when it starts with an UCN character

mihaipopescu:
Fixes mistakes done directly on master.


gmaurel:
I miss the fix.

mihaipopescu:
The fix for what?

Issue #359, opened 2015-03-29
uncrustify 0.61: `sp_case_label` config setting broken


geimer:
Applying uncrustify 0.61 to the following C++ example code
```
#include <iostream>

int main()
{
    int foo = 42;
    switch (foo) {
    case       1:
        std::cout << "1" << std::endl;
        break;
    case       2:
        std::cout << "2" << std::endl;
        break;
    default:
        std::cout << "Neither 1 nor 2." << std::endl;
    }
}
```
using this minimal configuration file
```
indent_columns      4
indent_with_tabs    False
sp_case_label       Force
```
has no effect. According to the documentation, I would have expected only a single space between the `case` and the labels `1` and `2`.

gmaurel:
The actual git version works pretty fine.
  https://github.com/uncrustify/uncrustify
Please test it!

geimer:
While I can confirm that it is now also working for numbers as `case` labels, it still doesn't work for character constants.  Maybe this is a better test case?
```
#include <iostream>

#define B_LABEL     'b'


int main()
{
    char foo = ' ';
    switch (foo) {
    case       32:
        std::cout << "space" << std::endl;
        break;
    case       'a':
        std::cout << "a" << std::endl;
        break;
    case       B_LABEL:
        std::cout << "b" << std::endl;
        break;
    default:
        std::cout << "Neither space, nor a, nor b." << std::endl;
    }
}
```
In any case, thanks for addressing this issue!

PR #568, opened 2016-07-15
Option sp_case_label consider now number as label.

gmaurel:
Issue #359 is fixed.


mihaipopescu:
It is slightly unrelated but we have noticed that there are some issues handling bitfields with enumeration constants as they were probably tokenized as labels.

something like:

```cpp
enum
{
  kEnumValue,
};

sturct foo
{
  int bar : kEnumValue;
}
```

It may be also true for switch case statements using enumerations, although the colon is on the right side.

PR #571, opened 2016-07-15
Issue with bitfields with enumeration constants

gmaurel:
bitfields with enumeration constants are tokenized as labels.
Issue #568 is fixed.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -2608,8 +2608,8 @@ void combine_labels(void)
                   set_chunk_type(next, CT_LABEL_COLON);
                }
                else if ((tmp == NULL) ||
-                        (tmp->type != CT_NUMBER &&
-                         tmp->type != CT_SIZEOF &&
+                        ((tmp->type != CT_NUMBER) &&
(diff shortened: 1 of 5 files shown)


src/combine.cpp
```diff
@@ -2592,14 +2596,26 @@ void combine_labels(void)
...
+               log_pcf_flags(LGUY, tmp->flags);
                if (next->flags & PCF_IN_FCN_CALL)
                {
                   /* Must be a macro thingy, assume some sort of label */
                   set_chunk_type(next, CT_LABEL_COLON);
                }
-               else if ((tmp == NULL) || (tmp->type != CT_NUMBER && tmp->type != CT_SIZEOF))
+               else if ((tmp == NULL) ||
+                        (tmp->type != CT_NUMBER &&
+                         tmp->type != CT_SIZEOF &&
```
mihaipopescu: If you are changing these lines could you add some extra parens to be sure we're not shooting ourselves in the foot (I don't _want_ to remember nor to rely on C++ operator precedence - magic stuff).

mihaipopescu:
Thanks for the fix !

Issue #323, opened 2015-02-07
sp_after_byref affects address-of operator in template argument list


chrullrich:
Before `sp_after_byref=add`:

```
class ATL_NO_VTABLE CProxy :
    public ATL::CComObjectRootEx<ATL::CComSingleThreadModel>,
    public ATL::CComCoClass<CProxy, &CLSID_Proxy>,
```

After:

```
class ATL_NO_VTABLE CProxy :
    public ATL::CComObjectRootEx<ATL::CComSingleThreadModel>,
    public ATL::CComCoClass<CProxy, & CLSID_Proxy>,
```

Note the space after the `&` in the last line.

gmaurel:
Help me:
What is the problem?

chrullrich:
That ampersand is the address-of operator, not declaring a reference. Spacing behind it should be controlled by sp_addr, not sp_after_byref.

mihaipopescu:
Nice catch! Could you paste the tokens map from the dump debug info file (`-p FILE` command line option).

chrullrich:
Here: [uncrustify.txt](https://github.com/uncrustify/uncrustify/files/365812/uncrustify.txt)

chrullrich:
Oh, that was from 0.61.3. Here is the same from 0.63: [uncrustify.txt](https://github.com/uncrustify/uncrustify/files/365828/uncrustify.txt)

mihaipopescu:
Cool, could you also try with the latest master ?

chrullrich:
That, too. [uncrustify.txt](https://github.com/uncrustify/uncrustify/files/366047/uncrustify.txt)

gmaurel:
Is it enough to test if prev is CT_COMMA

chrullrich:
It would have to be either "<" or comma, in case it is the first template argument.

gmaurel:
May I add: "if we are in a template"?

PR #580, opened 2016-07-21
Take care of &

gmaurel:
`class ATL_NO_VTABLE CProxy :
           public ATL::CComCoClass<CProxy, &CLSID_Proxy>`

The token & must be addr and not byref
Issue #323 is fixed

### DIFF
--- a/src/chunk_list.h
+++ b/src/chunk_list.h
@@ -214,9 +214,33 @@ bool chunk_is_star(chunk_t *pc)
 static_inline
 bool chunk_is_addr(chunk_t *pc)
 {
-   return((pc != NULL) &&
+   bool ret = ((pc != NULL) &&
           ((pc->type == CT_BYREF) ||
            ((pc->len() == 1) && (pc->str[0] == '&') && (pc->type != CT_OPERATOR_VAL))));
+   if (ret)
+   {
+      if (pc->flags & PCF_IN_TEMPLATE)
+      {
+         chunk_t *prev = chunk_get_prev(pc);
+         if ((prev->type == CT_COMMA) ||
+             (prev->type == CT_ANGLE_OPEN))
+         {
+            return(false);
+         }
+         else
+         {
+            return(true);
+         }
+      }
+      else
+      {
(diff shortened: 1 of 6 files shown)


src/chunk_list.h
```diff
@@ -214,9 +214,33 @@ bool chunk_is_star(chunk_t *pc)
 static_inline
 bool chunk_is_addr(chunk_t *pc)
 {
-   return((pc != NULL) &&
+   bool ret = ((pc != NULL) &&
```
mihaipopescu: no need to store this in a variable

src/chunk_list.h
```diff
@@ -214,9 +214,33 @@ bool chunk_is_star(chunk_t *pc)
...
           ((pc->type == CT_BYREF) ||
            ((pc->len() == 1) && (pc->str[0] == '&') && (pc->type != CT_OPERATOR_VAL))));
+   if (ret)
+   {
+      if (pc->flags & PCF_IN_TEMPLATE)
+      {
+         chunk_t *prev = chunk_get_prev(pc);
+         if ((prev->type == CT_COMMA) ||
```
sjakub: this should check whether prev is valid

sjakub:
The fix for prev=NULL actually breaks this code... if prev is NULL, then you don't hit any of the 'return' statements... Also, that function became a really big if-else-if structure, if I may suggest a more compact alternative, here it is:

```
static_inline
bool chunk_is_addr(chunk_t *pc)
{
   if ((pc != NULL) &&
       ((pc->type == CT_BYREF)
        || ((pc->len() == 1) && (pc->str[0] == '&') && (pc->type != CT_OPERATOR_VAL))))
   {
      chunk_t *prev = chunk_get_prev(pc);

      if ((pc->flags & PCF_IN_TEMPLATE) &&
           ((prev != NULL) && ((prev->type == CT_COMMA) || (prev->type == CT_ANGLE_OPEN))))
      {
         return(false);
      }

      return(true);
   }

   return(false);
}
```

Issue #322, opened 2015-02-07
Function declarations using macros


chrullrich:
Uncrustify does not correctly format a (C++ member) function declaration that is obscured by a macro (how should it ...):

    STDMETHOD(Initialize)(void);
    STDMETHOD(AddSubscription)(BSTR bsName);
    STDMETHOD(GetValues)(BSTR bsName, REFDATA** pData);

becomes

    STDMETHOD(Initialize) (void);
    STDMETHOD(AddSubscription) (BSTR bsName);
    STDMETHOD(GetValues) (BSTR bsName, REFDATA  * *pData);

STDMETHOD(x) expands to a lot of declarations, but you can think of it as `#define STDMETHOD(x) int x`.

There are two issues with this result: The space between the `) (`, and the spaces around the pointer asterisks in the argument lists. The latter come from sp_arith, but I don't know where exactly the extra space is coming from.

My config file, mostly courtesy of UniversalIndentGUI:

    tok_split_gte=false
    utf8_byte=false
    utf8_force=false
    indent_cmt_with_tabs=false
    indent_align_string=true
    indent_braces=false
    indent_braces_no_func=false
    indent_braces_no_class=false
    indent_braces_no_struct=false
    indent_brace_parent=false
    indent_namespace=false
    indent_extern=false
    indent_class=true
    indent_class_colon=false
    indent_else_if=false
    indent_var_def_cont=false
    indent_func_call_param=false
    indent_func_def_param=false
    indent_func_proto_param=false
    indent_func_class_param=false
    indent_func_ctor_var_param=false
    indent_template_param=false
    indent_func_param_double=false
    indent_relative_single_line_comments=false
    indent_col1_comment=false
    indent_access_spec_body=false
    indent_paren_nl=false
    indent_comma_paren=false
    indent_bool_paren=false
    indent_first_bool_expr=false
    indent_square_nl=false
    indent_preserve_sql=false
    indent_align_assign=true
    sp_balance_nested_parens=false
    align_keep_tabs=false
    align_with_tabs=false
    align_on_tabstop=false
    align_number_left=false
    align_func_params=false
    align_same_func_call_params=false
    align_var_def_colon=false
    align_var_def_attribute=false
    align_var_def_inline=false
    align_right_cmt_mix=false
    align_on_operator=false
    align_mix_var_proto=false
    align_single_line_func=false
    align_single_line_brace=false
    align_nl_cont=false
    align_left_shift=true
    align_oc_decl_colon=false
[...]

PR #582, opened 2016-07-22
Take care of **

gmaurel:
in STDMETHOD(GetValues)(BSTR bsName, REFDATA** pData);
Issue #322 is fixed


src/combine.cpp
```diff
@@ -1174,6 +1174,16 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
                   }
                }
             }
+            // Issue #322 STDMETHOD(GetValues)(BSTR bsName, REFDATA** pData);
+            if (pc->next->next != NULL)
+            {
+               if (pc->next->next->type == CT_STAR)
```
sjakub: if ( (pc->next->next) && (pc->next->next->type == CT_STAR) )

mihaipopescu: Not necessary as `pc->next->next` is null-checked at line `1178`.

sjakub: Yes, that's what I meant. Right now it looks like:
`if(a) { if (b) { ... } }`
and it could simply be:
`if(a && b) { ... }`

PR #587, opened 2016-07-22
Added `--if-changed` to write to stdout/FILE only if a change was detected

mihaipopescu:
New `--if-changed` option that will avoid writing and all related file operations if code was detected not to change. Caller can detect this by testing existence of '.uncrustify' file.

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -153,7 +153,7 @@ static void usage_exit(const char *msg, const char *argv0, int code)
            " --prefix PFX : Prepend PFX to the output filename path.\n"
            " --replace    : Replace source files (creates a backup).\n"
            " --no-backup  : Replace files, no backup. Useful if files are under source control.\n"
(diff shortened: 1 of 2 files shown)


sjakub:
Does this mean that '.uncrustify' file will not be created if it is no different than the original?
Also, how does that affect the --no-backup behaviour? The help text for the option should clarify that.

mihaipopescu:
The new option should not influence the behaviour of '--no-backup' or '--replace'. Why do you think otherwise ?

sjakub:
I feel like "write to" is confusing in context of files, especially if sometimes new files are not created regardless... I think it should say "create files or write to stdout".

gmaurel:
Well!
Merge or not merge?

sjakub:
Thanks! :)

mihaipopescu:
Thank you too for being watchful ! @gmaurel I think we're good to go if you don't have any other objections.

PR #594, opened 2016-07-26
Fixing a bug in nl_func_class_scope option and adding tests for it

dzenanz:


dzenanz:
In test 30070, the first time line 1882 `if ((prev != NULL) && (prev->type == CT_DC_MEMBER)` is encountered `prev->parent_type` is equal to `CT_NONE` instead of `CT_FUNC_DEF`. This is what caused the bug that this pull request is fixing. Is this a problem or expected behavior?

sjakub:
First of all, what bug are you fixing?
I just checked current master and 30070 test works just fine.
Also, please make sure that all the files have a newline at the end.

sjakub:
Also, in the original code there is no point in splitting those two 'if' statements (should be a single one), or even creating 'prev2' - unless you want to check if it's valid (which you don't at the moment).

dzenanz:
Sorry, the problematic test was 30069, not 30070. I will soon push the update addressing @mihaipopescu 's note.

sjakub:
Was that even ready? Why did it add code that is commented out???

mihaipopescu:
@dzenanz Please address the issue raised by @sjakub. Looks like you've committed something partially finished and @gmaurel just merged that in.

As a general note, please be extra careful when committing and merging stuff in. We can't rely just on the testing suite and blindly commit and merge stuff in. We have to scrutinize every commit and make use of the pull request procedure.

When contributing, please make sure we increase or at least maintain the code quality of the project as a lot of people depend on it.

mihaipopescu:
@sjakub I think @dzenanz actually fixed a bug in his own code. I think we should all try to have tests for all our changes or new features.

dzenanz:
I fixed my own code. The commented part of the code was extra safeguard which is redundant, but it does not evaluate to true for the first method implementation encountered. If that part of the code is uncommented, tests 30069 and 30067 fail.

mihaipopescu:
The confusion was why you've committed commented out code that looked like unfinished code.

sjakub:
Commented out code is pretty much useless. If it's needed, it shouldn't be commented out.
If there is something that should be explained, then there should be a comment explaining what is going on. A piece of code that is redundant and commented out is confusing, instead of clarifying things.

Also, you keep saying "first". First method encountered, first time line is encountered... What do you mean by that? Are you trying to handle the first method in a file differently? Why???

PR #599, opened 2016-07-29
Allocate some number ranges

gmaurel:
It makes much work at merging because more than one developer is working for cpp.
I propose to allocate some number ranges for these persons.

If other developer(s) wish a number range, let us know.

### DIFF
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -325,3 +325,20 @@
 33094 sp_angle_paren_empty.cfg         cpp/sp_angle_paren.cpp
 33095 bug_i_322.cfg                    cpp/bug_i_322.cpp
 33096 squeeze_ifdef_top.cfg            cpp/squeeze_ifdef.cpp
+
+# numbers upto 33199 are reserved for Ben
+# 33199
+
+# numbers from 33200 upto 33299 are reserved for Mihai Popescu
+# 33200
+# 33299
+
+# numbers from 33400 upto 33499 are reserved for Jakub Schmidtke
+# 33300
+# 33399
+
+# numbers from 33300 upto 33399 are reserved for Guy Maurel
+# 33400
+# 33499
+
+#


sjakub:
I actually thought a bit about that problem,
and maybe it would be possible to change how the tests are run?

Instead of having a single file (per language) with a list of tests, maybe it would be better
to have a directory with multiple files inside. Each file could contain one or more tests,
and tests wouldn't need to use numbers, but simply IDs (strings).

So let's say someone adds a new option, and 10 tests for it - they could all be put in
a single file, so they are all together, and yet separately from all the other tests.

This would eliminate conflicts, and the need to reserve ranges for different things.

If you think it's a good idea, I could modify the test scripts to work like that.

bengardner:
@sjakub I like your idea of breaking out to separate files.
Perhaps a subfolder per language and the tests could be identified by the language/filename/test number? The test numbers would only need to be unique per-file.
```
tests/
    languages/
        cpp/
            sometest
```
The output might be like
```
UNSTABLE: cpp/sometest/30265
```

mihaipopescu:
We were struggling with the same issues and as a workaround we have created a `staging.test` file (as a fake languge) with all tests that we want to integrate upstream. It's similar with @sjakub's idea but simpler as it's working with the current framework w/of any change.

Another idea we are considering is to have a test framework similar to Mercurial test fw or Cram https://pypi.python.org/pypi/cram. With it you can specify everything in one file and the tests can be extended to check more things like the tokens map for instance. The motivation is that uncrustify is a command line tool and the only important thing here is the user interface. This approach creates a very flexible test framework where you can test a lot of things. There won't be any intrinsic problem with merge conflicts as it won't be based on unique ids.

gmaurel:
@sjakub I like your proposal too.
Let us try it.

sjakub:
OK, I'll try to write something soon :)

gmaurel:
OK, we try another way.

PR #602, opened 2016-08-01
Failing test cases for calling convention not fixed in bug #633

scottbilas:
In January, fixes were made for this bug: https://sourceforge.net/p/uncrustify/bugs/633/

Previously, we had been using a workaround of marking "set COMMENT __stdcall" in our cfg files, so I tried removing them to validate the fix for 633. Definitely not fixed. You can see what's going on with the -p option to show the tokens. The presence of __stdcall will switch TYPE to WORD, FUNC_PROTO to FUNC_CALL, and PTR_TYPE to ARITH in many cases. We'll be continuing with our old workaround.

This PR adds more tests that demonstrate the problem more completely. Also removes a redundant test. 

Note that merging this PR will break tests, so it's meant only as a demonstration. (Though the test files updated here should be included in a real fix.)

### DIFF
--- a/tests/config/bug_633.cfg
+++ b/tests/config/bug_633.cfg
@@ -1,1 +1,10 @@
 sp_arith=add
+indent_class=true
+
+# uncomment to work around bug 633
+#set COMMENT __stdcall
+#set COMMENT __clrcall
+#set COMMENT __fastcall
+#set COMMENT __thiscall
+#set COMMENT __vectorcall
+#set COMMENT __cdecl
--- a/tests/config/stdcall.cfg
+++ b/tests/config/stdcall.cfg
@@ -1,1 +1,0 @@
-sp_arith=add
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -259,7 +259,6 @@
 33014 kdepim.cfg                       cpp/DB.cpp
 33015 kdepim.cfg                       cpp/Q_FOREACH.cpp
 33016 indent_once.cfg                  cpp/indent.cpp
-33017 stdcall.cfg                      cpp/stdcall.cpp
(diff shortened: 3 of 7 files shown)


mihaipopescu:
I think this is a very good way of submitting issues. What do you think @gmaurel @bengardner ?

gmaurel:
Right.
I create a new PR #761 to merge the workaround.

scottbilas:
@gmaurel I think you completely misunderstood this if you are creating a copy of this PR with the same commits + my local workaround.

My report here is solely that the original improvement to add stdcall etc was done incorrectly, and the tests I added demonstrate this. Taking those and then merging them along with what I've been using as a workaround is totally the wrong thing to do.

It's extra weird because you were the one to originally implement stdcall, so merging in a workaround that effectively disables it all with "set COMMENT" is baffling.

I suggest one of the following:

* Fix the stdcall code to work correctly, so that TYPE does not become WORD etc. such as my test in this PR shows.
* Remove the stdcall code entirely and hope that someone else picks it up. (I'd prefer this over incorrect behavior that we must work around).

I also suggest while you're at it to remove the hard coded variations of all calling conventions and simply provide a marker keyword that represents such a token, which we can provide as an 'include' file for Microsoft compiler users to use. Same as I recommended earlier for the qt keywords. That way we do not unnecessarily pollute Uncrustify with compiler-specific keywords. This will simplify the code a lot as well.

No matter what though - do not merge this PR in copy-pasted form https://github.com/uncrustify/uncrustify/pull/761. That's the wrong way to go.

gmaurel:
I make a alternative PR at https://github.com/uncrustify/uncrustify/pull/782
Please comment!

PR #603, opened 2016-08-02
The TODO from output.cpp make the first_len minimum configurable

gmaurel:
To make possible to change the last line of a multi-line comment from
` */`
to
`*/`


src/options.cpp
```diff
@@ -1778,6 +1783,11 @@ int load_option_file(const char *filename)
    {
       cpd.line_number++;
 
+      if (cpd.line_number == 528)
```
bengardner: Would you remove this? I assume it is left over from some test code.

PR #612, opened 2016-08-05
Build with CMake on Travis

jibsen:
This is work in progress, I am (ab)using the PR to get results from Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: cpp
-compiler: gcc
-os: linux
-before_script: ./autogen.sh
+
 branches:
   only:
   - master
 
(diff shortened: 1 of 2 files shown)


jibsen:
Okay, think this is working now.

The first commit switches to building with CMake on Travis.

The next two compile with GCC 6 and add an address sanitizer build (and some logic for other potential build types). I don't know if there is any interest in that?

mihaipopescu:
Sanitizers are ok but what about Release build ? I don't see it covered and I think that should be the default unless we want. Debug is usually run manually to debug issues locally.

jibsen:
Good point, I've added a release build.

PR #614, opened 2016-08-06
Coverity says CID 76238 (#1 of 1): Out-of-bounds read (OVERRUN)

gmaurel:
I try to fix the bug detected by Coverity

### DIFF
--- a/src/char_table.h
+++ b/src/char_table.h
@@ -10,6 +10,8 @@
 #define CHAR_TABLE_H_INCLUDED
 
 
+#define CharTable_Length 128
+
 /**
  * bit0-7 = paired char
  * bit8 = OK for keyword 1st char
@@ -17,7 +19,8 @@
  */
 struct CharTable
 {
-   static int chars[128];
+   //static int chars[128];
+   static int chars[CharTable_Length];
 
    enum
    {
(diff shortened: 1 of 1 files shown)


src/char_table.h
```diff
@@ -32,7 +35,9 @@ struct CharTable
       {
          return(0);
       }
-      if (ch < (int)ARRAY_SIZE(chars))
+      //if (ch < (int)ARRAY_SIZE(chars))
+      // Coverity: CID 76238 (#1 of 1): Out-of-bounds read (OVERRUN)
+      if (ch < CharTable_Length)
```
scottbilas: There are 29 matches of ARRAY_SIZE usage in the code, and this one looks correct to me. Why is Coverity complaining only about this one, and why is switching to a #define at all helpful?

This looks like the wrong fix to me.

bengardner: I agree that this fix is wrong.

Coverity is likely complaining that ch is an int, which means it could be negative.
Maybe try 
```
if (ch > 0 && ch < ARRAY_SIZE(chars))
```

gmaurel: Well, the coverity says about the  values the parameter ch might have.
The test ch < 0 is correct.
The test ch < (int)ARRAY_SIZE(chars) is not enough because ch might have a value (let say 300) which is bigger as the size of the array chars[128]
My proposal is also not enough because it doesn't test the values above 127, which should be
"return(0)"

gmaurel:
I take all this back.
Sorry for the noise.

Issue #604, opened 2016-08-03
CMake build issues


mihaipopescu:
@jibsen I've just tested cmake build on windows and apparently it works but:
- [x] there are no instructions on how to build with cmake
- [x] cmake fails when it tries to create a `/Testing` directory because there is another file with the same name `/TESTING`
- [x] seems like process is rather messy. It creates everything in the root and pollutes the repository with untracked files.
- [x] modify `appveyor.yml` and `.travis.yml` to switch to cmake
- [x] Add appveyor build status badge in README.md
- [x] add helper script for running tests
- [x] make_version.py is not working
- [ ] repository has to be cleaned of old build system files (gnu automake, mingw32 scripts, VS and Xcode solution files)

mihaipopescu:
Other related issues #470, #538, #600.

jibsen:
I would always recommend doing an out-of-source build with CMake, so all the files created do not end up in the source tree:

~~~.cmd
mkdir build
cd build
cmake ..
~~~

Cleaning up is just a matter of removing the build folder.

mihaipopescu:
Can't the output/temp folders be specified ? I would be very surprised if not.

jibsen:
I may be wrong, but I think CMake always creates them in the folder cmake is run from. Perhaps it would be worth adding a check to the CMake script that warns the user if attempting an in-source build?

Regarding cleaning up the other build systems, I think it is up to the authors if they feel comfortable enough with CMake as the only option.

mihaipopescu:
I have no problem with changing the `cwd` if that's the only way for cmake. I have little to none experience with CMake.

I would agree with any build system as long as it's cross platform and can create native executables for that OS.

mihaipopescu:
@jibsen I've branched the discussion into a new issue #606. We can keep this one only for CMake related issues.

jibsen:
Thanks. I've opened #605 and #608 to address some of these.

mihaipopescu:
Thanks @jibsen. I'll try them out.

mihaipopescu:
Let me know if you think there should be something else.

jibsen:
The output files from the tests should probably go into the build folder somewhere (using the `--results` option). I'll take a look at that.

mihaipopescu:
Why ? I think it's good where they are right now in the `tests` folder. The test fw just has to find the executable, that's all I guess.

jibsen:
[...]

PR #618, opened 2016-08-09
Make generating token_names.h part of the build

jibsen:
As discussed in #604, `token_names.h` was not being regenerated if the source file changed. This should hopefully fix that.


jibsen:
Btw, Travis is not using CMake yet, so the build check means nothing for this PR.

mihaipopescu:
Yes. We are waiting for @gmaurel to integrate AppVeyor.

mihaipopescu:
Btw, I've noticed that `token_names.h` is not part of the workspace. It don't think it's required but I wanted to ask if this change also links it's rebuild as a part of the regular build process. For instance, if I'm adding or removing a token from `token_enums.h` and if I do a build does it trigger the regenerate action ?

Now I have to go to command line and run `make_token_names.bat` but I wish it was added as a pre-build event.

mihaipopescu:
Now I have to go to command line and run `make_token_names.bat` and move it from `src` to `build` folder but I wish it was added as a pre-build event.

jibsen:
With this change, it should trigger a rebuild of `token_names.h` (and the files that include it) when building, if you change `token_enums.h`. You can look for the line "Generating token_names.h" in the build log.

PR #625, opened 2016-08-10
MinGW cross-compile support

jibsen:
This adds [toolchain files](http://www.vtk.org/Wiki/CMake_Cross_Compiling) for MinGW cross-compilers from the mingw and mingw-w64 packages. I have only tested it on [Fedora](https://fedoraproject.org/wiki/MinGW/Tutorial), but I believe it should work on Ubuntu as well.

Basically, you install the mingw cross-compiler packages (on Fedora these are `mingw32-gcc`, `mingw32-gcc-c++`, and `mingw32-winpthreads-static` for Win32), and then build with:

~~~.bash
mkdir buildwin
cd buildwin
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchain-mingw32.cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXE_LINKER_FLAGS="-static -s" ..
make
~~~

I hadn't used CPack before, but after a little googling around I think I got it to build a zip package close to the one `make_win32.sh` produces. After the above, just do `cpack`, and you get a zip file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,12 +250,47 @@ if(BUILD_TESTING)
 endif()
 
 #
+# Package
+#
+set(CPACK_PACKAGE_NAME "uncrustify")
+set(CPACK_PACKAGE_VERSION "${PACKAGE_VERSION}")
+set(CPACK_PACKAGE_VENDOR "Ben Gardner")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Code beautifier")
+set(CPACK_PACKAGE_DESCRIPTION_FILE "${PROJECT_SOURCE_DIR}/README.md")
+set(CPACK_RESOURCE_FILE_LICENSE "${PROJECT_SOURCE_DIR}/COPYING")
+set(CPACK_RESOURCE_FILE_README "${PROJECT_SOURCE_DIR}/README.md")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -250,12 +250,47 @@ if(BUILD_TESTING)
...
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+  install(TARGETS uncrustify DESTINATION ".")
+  install(
+    FILES
+      "${PROJECT_SOURCE_DIR}/README.md"
+      "${PROJECT_SOURCE_DIR}/BUGS"
+      "${PROJECT_SOURCE_DIR}/ChangeLog"
+    DESTINATION "."
+  )
+  install(FILES "${PROJECT_SOURCE_DIR}/documentation/htdocs/index.html"
```
jayvdb: does this only include the `index.html`, or magically all the files that are referenced by `index.html`?

jibsen: I just copied [what `make_win32.sh` does](https://github.com/uncrustify/uncrustify/blob/master/make_win32.sh#L53). If you look at the release zip on SF, it only includes `index.html`.

Of course it would be no problem including the entire folder if that's required.

Issue #410, opened 2015-11-12
C++11 lambdas curly brackets are 1 indent off, missing option?


mlt:
As of right now (uncrustify-0.61.3-gf65394e-win32), curly brackets for C++11 lambda are placed on the same indentation level as enclosing function call, i.e. making lambda's body match what otherwise would be line continuation. It would be nice to add an option to control whether general indentation rules shall apply for lambda body instead of line continuation rules.
Instead of

```
        m_ShutdownTimer.async_wait (
            [](const boost::system::error_code& ecode)
        {
            Daemon.running = 0;
        });
```

I'd like an option to have

```
        m_ShutdownTimer.async_wait (
            [](const boost::system::error_code& ecode)
            {
                Daemon.running = 0;
            });
```

Did I miss something? There are only 2 lambda options.
I have

```
indent_continue                           = 0        # number
indent_with_tabs                          = 1        # number
indent_func_call_param                    = true
sp_cpp_lambda_assign                      = ignore   # ignore/add/remove/force
sp_cpp_lambda_paren                       = remove   # ignore/add/remove/force
```

gmaurel:
I take it in the TODO list.

kankaristo:
I have a similar issue.

For example, this:
```cpp
    return std::shared_ptr<SDL_Surface>(
        image,
        [](SDL_Surface* surface) {
            LOG_DEBUG << "Freeing SDL_Surface: " << surface;
            SDL_FreeSurface(surface);
        }
    );
```

Turns into:
```cpp
    return std::shared_ptr<SDL_Surface>(
        image,
        [](SDL_Surface* surface) {
        LOG_DEBUG << "Freeing SDL_Surface: " << surface;
        SDL_FreeSurface(surface);
    }
);
```

Similarly, this:
```cpp
    auto sortListing = (
        [](const std::string &a, const std::string &b) {
            // Function body
        }
    );
```

Turns into:
```cpp
    auto sortListing = (
        [](const std::string &a, const std::string &b) {
        // Function body
    }
);
```

exploser:
I just tried the new option and noticed that it only fixes indentation for lambdas that have parentheses in the declaration. When that is not the case, however, lambdas still get indented as if indent_cpp_lambda_body is false.

For example,
```
 auto foo = (
        []() {
            // Function body
        }
        );
```
remains fine, but if I use a shorthand version:
```
 auto foo = (
        [] {
            // Function body
        }
        );
```
the result is:
```
[...]

PR #626, opened 2016-08-11
refs #410 add test for indent_cpp_lambda_body option

vyacheslavs:
test lambda intendation


gmaurel:
Please make a PR with the code you want to modify.

PR #627, opened 2016-08-11
register each test separately

dirk-thomas:
By registering each test separately you can leverage the full power of `ctest`, e.g. filter by labels (each language), by regex, rerun the test which failed on the last run, etc.

Since the `run_tests.py` script already allows passing in an optional language and number range this PR only needed to touch the CMake code.

The downside regarding speed: Python is being invoke for every test (which requires several milliseconds of startup time of the interpreter per test).

The upside regarding speed: you can use `ctest -j N` to run the tests in parallel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,4 +63,6 @@ before_script:
     esac
 
 
-script: make VERBOSE=1 && make test
+script:
+  - make VERBOSE=1
+  - ctest -V
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,14 +238,32 @@ configure_file(man/uncrustify.1.in uncrustify.1 @ONLY)
 if(BUILD_TESTING)
   find_package(PythonInterp REQUIRED)
   enable_testing()
+  # parse the .test files and create one test per line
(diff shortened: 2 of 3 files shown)


dirk-thomas:
The second commit updates the AppVeyor and Travis configuration to invoke `ctest` instead of `run_tests.py`.

mihaipopescu:
Nice ! I think the next step would be to port as much functionality out of the `run_tests.py` as possible to be able to get rid of the dependency. The ideal would be to invoke `uncrustify` directly from cmake script and take advantage of it in full.

Could you paste the console output from a `ctest` run ?

dirk-thomas:
You can see the `ctest` output in the logs of  Travis ad AppVeyor (both are running with `-V` in order to see the "unstable" messages).

I don't see how a ctest can invoke `uncrustify` directly since beside running the formatter it also needs to compare the output and potentially run the formatter another time (and compare again).

But we could remove the `.test` files and list the tests in CMake instead. After that the `run_tests.py` script could be reduced to may be a third of its size (I would keep only the logic of `run_tests` with no coloring and likely no command line options) and also rename it to `run_test`.py` since it only needs to run a single test at a time.

PR #629, opened 2016-08-14
Support optional parens for PROTO_WRAP

mihaipopescu:
Previously, for prototype macro wrap functions, parameters with `CT_AMP` and `CT_STAR` where tokenized as `CT_ARITH` when they were not in parenthesis. The change makes parens optional.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -5702,7 +5702,7 @@ static void handle_wrap(chunk_t *pc)
 /**
  * A proto wrap chunk and what follows should be treated as a function proto.
  *
- * RETTYPE PROTO_WRAP( NAME, PARAMS );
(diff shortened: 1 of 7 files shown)


src/combine.cpp
```diff
@@ -5740,7 +5739,17 @@ static void handle_proto_wrap(chunk_t *pc)
    set_chunk_parent(clp, pc->type);
 
    set_chunk_parent(tmp, CT_PROTO_WRAP);
-   fix_fcn_def_params(tmp);
+
+   if (tmp->type == CT_PAREN_OPEN)
+   {
+	  fix_fcn_def_params(tmp);
```
mihaipopescu: Here I've made an indentation mistake (tab instead of space). It will eventually get fixed with other source formatting changes.

mihaipopescu:
I guess the conflict is at tests. I can remerge if you want.

PR #630, opened 2016-08-15
Run Uncrustify on itself

mihaipopescu:
Formatted uncrustify sources with itself using `etc/dofiles.sh` which uses `etc/ben.cfg`

### DIFF
--- a/src/ChunkStack.h
+++ b/src/ChunkStack.h
@@ -22,19 +22,18 @@ public:
       {
       }
 
-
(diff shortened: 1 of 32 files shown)


gmaurel:
mmmh!
I introduce a configuration file forUncrustifySources.cfg on April-27
to make a difference to ben.cfg

mihaipopescu:
Oh, I didn't knew you'd like that. No problem. You can ignore this PR then.

Issue #633, opened 2016-08-15
typename derails variable definition alignment


dzenanz:
With `align_var_def_span` turned on
```
typename TImage::PixelType m_Label;
int                        m_Axis;
bool                       m_HeuristicAlignment;
```
gets turned into
```
typename TImage::PixelType m_Label;
int  m_Axis;
bool m_HeuristicAlignment;
```
instead of remaining as it is.

I created a test depicting this problem in my [fork](https://github.com/dzenanz/uncrustify/commit/07399f30fb846fa181d14f207320598c295e6887). Should I turn it into a pull request and reference this ticket?

mihaipopescu:
`align_var_def_span` it's not an on/off option. It is used to determine if it should align variables or not within that defined span. 

```
align_var_def_span                                Number
  The span for aligning variable definitions (0=don't align)
```

for eg.:

```cpp
extern bool           QT_SIGNAL_SLOT_found;
extern int            QT_SIGNAL_SLOT_level;
extern bool           restoreValues;
align_var_def_span    ^
```

dzenanz:
OK, I updated my [fork](https://github.com/dzenanz/uncrustify/commit/99606ba7fb72fd6278cb40e43450d9437ba3c7b7) to contain both positive and negative examples:
```
typename TImage::PixelType m_Label;
int                        m_Axis;
bool                       m_HeuristicAlignment;

aLongTypeNameButNotSpecial m_Label;
int                        m_Axis;
bool                       m_HeuristicAlignment;
```
which gets transformed into:
```
typename TImage::PixelType m_Label;
int  m_Axis;
bool m_HeuristicAlignment;

aLongTypeNameButNotSpecial m_Label;
int                        m_Axis;
bool                       m_HeuristicAlignment;
```

mihaipopescu:
Do you think the problem is the `typename` or the namespaced type `TImage::PixelType` ? What happens if you change one or the other ?

dzenanz:
typename is the problem. This code fraction gets aligned correctly:
```
TImage::PixelType m_Label;
int               m_Axis;
bool              m_HeuristicAlignment;
```
but this does not:
```
typename PixelType m_Label;
int  m_Axis;
bool m_HeuristicAlignment;
```

mihaipopescu:
Could you also dump the tokenization map using the `-p` flag ? I suspect something fishy going on there around the `typename` keyword.

dzenanz:
[...]

PR #634, opened 2016-08-17
Fixes typename var def

mihaipopescu:
A variable definition can start with the `typename` keyword so uncrustify should take this into account.

Fix for issue #633.


dzenanz:
Thanks for the fix Mihai!

PR #635, opened 2016-08-18
Separate handling of commas in enums

versusvoid:
I like my functions simple:

```cpp
void foo(int a, int b, int c);
```

but like enums with entities on separate lines:

```cpp
enum Test
{
    A,
    B,
    C,
}
```

hence new option.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -3231,7 +3231,7 @@ void newlines_chunk_pos(c_token_t chunk_type, tokenpos_e mode)
    chunk_t *prev;
    int     nl_flag;
 
-   if ((mode & (TP_JOIN | TP_LEAD | TP_TRAIL)) == 0)
+   if ((mode & (TP_JOIN | TP_LEAD | TP_TRAIL)) == 0 && chunk_type != CT_COMMA)
    {
       return;
    }
@@ -3252,6 +3252,10 @@ void newlines_chunk_pos(c_token_t chunk_type, tokenpos_e mode)
             {
                // change mode
                mode_local = cpd.settings[UO_pos_class_comma].tp;
+            }
+            else if (pc->flags & PCF_IN_ENUM)
+            {
+               mode_local = cpd.settings[UO_pos_enum_comma].tp;
             }
             else
             {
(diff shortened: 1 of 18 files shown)


src/newlines.cpp
```diff
@@ -3231,7 +3231,7 @@ void newlines_chunk_pos(c_token_t chunk_type, tokenpos_e mode)
    chunk_t *prev;
    int     nl_flag;
 
-   if ((mode & (TP_JOIN | TP_LEAD | TP_TRAIL)) == 0)
+   if ((mode & (TP_JOIN | TP_LEAD | TP_TRAIL)) == 0 && chunk_type != CT_COMMA)
```
mihaipopescu: What was the reasoning behind this change ? This changes behaviour and it's not entirely related to the new option. Basically you're saying that now all modes are valid unless chunk_type is not a comma. Can you show with a test the effect of what this does and how it was wrong previously for `CT_CLASS_BASE` ?

versusvoid: @mihaipopescu Yeep, easy. See changes in uncrustify.cpp:

```cpp
if (cpd.settings[UO_pos_comma].tp != TP_IGNORE || cpd.settings[UO_pos_enum_comma].tp != TP_IGNORE)
{
   newlines_chunk_pos(CT_COMMA, cpd.settings[UO_pos_comma].tp);
}
```

with config

```
pos_comma = ignore
pos_enum_comma = trail_force
```

it will return (because ```(mode & (TP_JOIN | TP_LEAD | TP_TRAIL)) == 0```) and leave enums unchanged.

It has produced expected results for ```pos_class_comma``` independently of ```pos_comma``` value, because class commas are separately handled in  ```newlines_class_colon_pos()```.

mihaipopescu: Well, I was just saying that we should have coverage on the original bad behaviour (i.e. to show there was a bug and we had to fix it). If that's not possible with the other option it's ok.

versusvoid: @mihaipopescu 
If I understand correctly what you're asking for, ```enum_comma-1.cfg``` and ```enum_comma-2.cfg``` should provide such coverage.

mihaipopescu: Not entirely. I was just thinking about showing how the previous code was wrong for the UO_pos_class_comma but if you say that it's a different code path (handled in `newlines_class_colon_pos` I guess that's not possible to show.

versusvoid:
So, any further suggestions for enhancement?

mihaipopescu:
Nope, I guess you're all good. :+1:

PR #641, opened 2016-08-21
The actual authors list.

gmaurel:
Please comment the list.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,5 +1,10 @@
+Author:
 2005 - 2016 : Ben Gardner
 
+Maintenance:
+Guy Maurel
+
+Other collaborators:
 Dirk Thomas
 Dženan Zukić
 Guy Maurel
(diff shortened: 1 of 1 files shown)


mihaipopescu:
I'm in favour of having this file instead of having names in the source code. The code should be as impersonal as possible.

```cpp
/**
 * @file align.cpp
 * Does all the aligning stuff.
 *
 * @author  Ben Gardner
 * @author  Guy Maurel since version 0.62 for uncrustify4Qt
 *          October 2015, 2016
 * @license GPL v2+
 */
```

```cpp
#ifdef DEBUG
   LOG_FMT(LGUY, "(%d) ", __LINE__);
#endif
   LOG_FMT(LGUY, "%s: %d:%d %s\n", __func__, pc->orig_line, pc->orig_col, pc->text());
```

mihaipopescu:
Also, let's don't forget the contributors from SourceForge.

PR #642, opened 2016-08-21
Testing framework improvements

mihaipopescu:
Optimized running tests from cmake by launching uncrustify directly by cmake and removing the dependency on the python script.

Unfortunately due to the complexity of the test and a current limitation of the cmake testing framework, ctest still needs to launch cmake to check the blessed and resulted output files thus making the process to run a bit slower.

Some benchmarks:

| cmd | before [s] | after [s] | improvement |
|:-----|:----|:----|:---|
| ctest -Q | 39.78 | 25.11 | - 36.88% |
| ctest | 68.32 | 36.26 | - 46.93% |

I'm sure it can be optimised further but it could be good as a first iteration.
@dzenanz @dirk-thomas What do you think ?


mihaipopescu:
Also, I have tried to optimize the console output to have something similar as for `run_tests.py` to hide the passed test and to output just failures.

`ctest --output-on-failure`

but that doesn't give me the desired output. I don't know if it's possible.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+        set(test_input input/${CMAKE_MATCH_4})
+
+        get_filename_component(test_dir ${CMAKE_MATCH_4} PATH)
+        get_filename_component(test_file ${CMAKE_MATCH_4} NAME)
+
+        set(test_expected ${test_dir}/${CMAKE_MATCH_1}-${test_file})
+        set(test_output output/${test_expected})
+        set(test_result results/${test_expected})
+
+        if("${CMAKE_MATCH_2}" STREQUAL "!")
```
dirk-thomas: If you want to be safe you might want to add trailing spaces to both strings to avoid double expansion (same as in line 27).

dirk-thomas:
Beside the minor comment this looks good to me and CI seems to be happy too :+1:

tests/CMakeLists.txt
mihaipopescu: That's odd. Ok, I didn't knew that, thanks.

dzenanz:
Good job, Mihai!

Issue #648, opened 2016-08-23
Group class in align_var_struct_span instead of align_var_def_span?


dzenanz:
Currently `align_var_def_span` controls alignment of variable/member declarations in both functions and classes, while `align_var_struct_span` controls alignment in structs and unions. Wouldn't it be more logical to group class member declarations together with struct/union than with functions?

I tried adding `align_var_class_span`, `align_var_class_thresh` and `align_var_class_gap` to have separate control over class member declaration alignment, but these tests fail (as reported by CTest):

```
5 - c_sharp_10010 (Failed)
267 - cpp_30018 (Failed)
272 - cpp_30023 (Failed)
436 - cpp_32003 (Failed)
522 - d_40000 (Failed)
528 - d_40006 (Failed)
529 - d_40007 (Failed)
556 - java_80030 (Failed)
```

Should I modify the tests to comply with the separation, join class with struct/union or what?

dzenanz:
Perhaps I should make it backwards compatible by using the new settings only if `UO_align_var_class_span` is non-zero? @mihaipopescu @gmaurel

mihaipopescu:
I think your PR #649 is reasonable. I think the solution is to update the tests as well to take the new option into account. We should mention that in the release notes.

dzenanz:
Done.

PR #649, opened 2016-08-23
Introducing align_var_class_span and friends

dzenanz:
#648 Adding align_var_class_span, align_var_class_thresh and align_var_class_gap
to have separate control over class member declaration alignment.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -283,7 +283,8 @@ void align_all(void)
 
    /* Align variable definitions */
    if ((cpd.settings[UO_align_var_def_span].n > 0) ||
-       (cpd.settings[UO_align_var_struct_span].n > 0))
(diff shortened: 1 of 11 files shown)


mihaipopescu:
While updating a test can you also add new tests for the new options ?

dzenanz:
This is ready to be merged.

gmaurel:
Not in this way, please.
Do not change the "ben.cfg"
As it is necessary, create a new configuration file to be used for your changes.

dzenanz:
Either ben.cfg needs to be changed, or the expected output of a few tests which use it. The third commit adds a test for the new functionality.

mihaipopescu:
@gmaurel If we change ben.cfg then we have to change all references. For what the old ben.cfg is useful ? I've already seen one in `etc` folder. I think we can make a backup of it if that's what you want (even if we can always get back from git history).

PR #654, opened 2016-08-26
Indent constructor members

agateau-g:
If one sets `indent_class`, `indent_ctor_init` and `indent_constr_colon`, initialized member variables were not indented twice.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1417,9 +1417,19 @@ void indent_text(void)
                else
                {
                   next = chunk_get_next(pc);
-                  if ((next != NULL) && !chunk_is_newline(next))
+                  if (next != NULL)
                   {
-                     frm.pse[frm.pse_tos].indent = next->column;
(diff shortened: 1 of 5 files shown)


gmaurel:
Your comment is too short for me.
Could you push more explanations, why, why not, and some more.
Thanks.

agateau-g:
Sure, here is a bit more context. Our coding style says we should use 4 space indentation and indent constructor initializers twice. Something like this:

```
MyClass::MyClass() :
        BaseClass(12),
        mSomeVariable(24) {
    constructorBody();
}
```

Looking at uncrustify doc, I expected that the combination of setting `indent_constr_colon` to true and `indent_ctor_init` to 4 would produce this result, but it did not work: initializers where indented like the constructor body, like that:

```
MyClass::MyClass() :
    BaseClass(12),
    mSomeVariable(24) {
    constructorBody();
}
```

So I made this patch which applies the value of `indent_ctor_init` to constructor initializers.

PR #655, opened 2016-08-26
Removing (123) from UniversalIndentGUI's documentation

dzenanz:


gmaurel:
I would like to keep it under the condition
#ifdef DEBUG
Changing the number 123 with the vim-command as noticed, one get the real line of the data, which can be seen using UniversalIndentGUI by staying on the command. This show the line number. In this way, it is easier to find it in the source. Only usefull by debugging.

dzenanz:
@gmaurel I made the modification.

PR #656, opened 2016-08-28
Coverity helps to find Dereference after null check

gmaurel:
Add some check of pointers

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -336,7 +336,7 @@ static chunk_t *skip_dc_member(chunk_t *start)
  * This is called on every chunk.
  * First on all non-preprocessor chunks and then on each preprocessor chunk.
  * It does all the detection and classifying.
- * This is only called by fix_symbols at line 1345.
+ * This is only called by fix_symbols.
  * The three parameters never get the value NULL.
  * it is not necessary to test.
  */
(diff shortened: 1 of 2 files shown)


src/combine.cpp
```diff
@@ -352,6 +352,10 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
    //         pc->text(), get_token_name(pc->type),
    //         next->text(), get_token_name(next->type));
 
+   if (next == NULL)
+   {
+      return;
```
mihaipopescu: This will basically skip checking the last symbol. I think we need to check for null only where it's required.

src/combine.cpp
```diff
@@ -336,6 +336,9 @@ static chunk_t *skip_dc_member(chunk_t *start)
  * This is called on every chunk.
  * First on all non-preprocessor chunks and then on each preprocessor chunk.
  * It does all the detection and classifying.
+ * This is only called by fix_symbols at line 1345.
```
mihaipopescu: > This is only called by fix_symbols at line 1345.

This statement is very likely to change.

src/combine.cpp
```diff
@@ -336,6 +336,9 @@ static chunk_t *skip_dc_member(chunk_t *start)
...
+ * The three parameters never get the value NULL.
```
mihaipopescu: Maybe you can assert this at the beginning of the function.

PR #658, opened 2016-08-30
Add assorted indent and newline options

brmqk3:
Options added:
- indent parens after function definition
- if function parameters are empty, leaves parens on same
  line
- indenting case statements, extern blocks, function
  definitions, and braces directly inside include guards

Change-Id: If9d02e8b82c1ad9e809957d261782a46d9a389a4

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1342,14 +1342,22 @@ void indent_text(void)
              (cpd.settings[UO_mod_case_brace].a & AV_REMOVE)) &&
              (prev->parent_type == CT_CASE))
          {
-            /* Affects the brace and closes the case */
-            int tmp = frm.pse[frm.pse_tos].indent - indent_size;
-
-            frm.pse[frm.pse_tos].indent     = tmp;
-            frm.pse[frm.pse_tos].indent_tmp = tmp + indent_size;
-            frm.pse[frm.pse_tos].indent_tab = tmp;
-
-            indent_column_set(frm.pse[frm.pse_tos].indent_tmp);
+		   if (cpd.settings[UO_indent_case_brace].n > 0)
+		   {
+		   	  /* Affects the brace and closes the case */
+              int tmp = frm.pse[frm.pse_tos].indent - cpd.settings[UO_indent_case_brace].n;
+              
+              frm.pse[frm.pse_tos].indent     = tmp;
+              frm.pse[frm.pse_tos].indent_tmp = tmp + cpd.settings[UO_indent_case_brace].n;
+              frm.pse[frm.pse_tos].indent_tab = tmp;
+              
+              indent_column_set(frm.pse[frm.pse_tos].indent_tmp);
+		   }
+		   else
+		   {
+              /* This only affects the 'break', so no need for a stack entry */
+              indent_column_set(prev->column);
+		   }
          }
       }
       else if (pc->type == CT_LABEL)
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -666,18 +666,44 @@ void indent_text(void)
               (pc->parent_type == CT_PP_ELSE)))
          {
             next = chunk_get_next(pc);
-            /* Hack to get the logs to look right */
-            memtype = next->type;
-            set_chunk_type(next, CT_PP_IF_INDENT);
-            indent_pse_push(frm, next);
-            set_chunk_type(next, memtype);
+            int should_indent_preproc = true;
+            chunk_t* preproc_next = chunk_get_next_nl(pc);
+            preproc_next = chunk_get_next_nblank(preproc_next);
 
-            /* Indent one level except if the #if is a #include guard */
-            int extra = ((pc->pp_level == 0) && ifdef_over_whole_file()) ? 0 : indent_size;
-            frm.pse[frm.pse_tos].indent     = frm.pse[frm.pse_tos - 1].indent + extra;
-            frm.pse[frm.pse_tos].indent_tab = frm.pse[frm.pse_tos - 1].indent_tab + extra;
-            frm.pse[frm.pse_tos].indent_tmp = frm.pse[frm.pse_tos].indent;
-            frm.pse[frm.pse_tos].in_preproc = false;
+            /* Look ahead at what's on the line after the #if */
+            while (preproc_next->type != CT_NEWLINE)
+            {
+                if ((((preproc_next->type == CT_BRACE_OPEN) ||
+                    (preproc_next->type == CT_BRACE_CLOSE)) &&
+                    !cpd.settings[UO_pp_indent_brace].b) ||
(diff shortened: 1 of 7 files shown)


gmaurel:
It might interesting...
But we are working with a little different way. Such as:
Explain what you wanted to get, what is wrong on the actual version, why.
Give us more informations about the fix you have proposed.
Let run ALL the tests to confirm there are no regression with your proposal.
Thanks for it.

brmqk3:
Do I need to put the information in the commit message? Or just comment here?

mihaipopescu:
@brmqk3 Why did you had to enable all options from config files to fix the tests ? Are you changing the default behavior ?

brmqk3:
I changed this so that when the options get set to false by default, they won't actually change the default indenting. You will have to set these options to true to change any of the formatting.

mihaipopescu:
That's not what I meant. If you're not changing the default behaviour and just add new options on top of it with other default value just add that option in the `set_option_defaults` function while preserving the default behaviour. A negative option is quite hard to comprehend. I'm sorry I wasn't clear.

brmqk3:
Oh okay, yeah I was debating between doing that as well. Wasn't sure which would be better practice. I'll switch to set in the set_option_defaults function.

src/indent.cpp
```diff
@@ -666,18 +666,44 @@ void indent_text(void)
...
-            memtype = next->type;
-            set_chunk_type(next, CT_PP_IF_INDENT);
-            indent_pse_push(frm, next);
-            set_chunk_type(next, memtype);
+            int should_indent_preproc = true;
+            chunk_t* preproc_next = chunk_get_next_nl(pc);
+            preproc_next = chunk_get_next_nblank(preproc_next);
+
+            /* Look ahead at what's on the line after the #if */
+            while (preproc_next->type != CT_NEWLINE)
```
mihaipopescu: what if `preproc_next` gets NULL ?

tests/config/pp_if_indent-1.cfg
```diff
@@ -4,4 +4,3 @@ indent_columns     = 4
 pp_if_indent_code  = true
 pp_indent_if       = 0
 pp_indent_at_level = true
-
```
mihaipopescu: you probably committed this one by mistake

src/indent.cpp
```diff
@@ -1312,11 +1338,26 @@ void indent_text(void)
       {
          prev = chunk_get_prev_ncnl(pc);
          if ((prev != NULL) &&
-             (prev->type == CT_BRACE_CLOSE) &&
+             ((prev->type == CT_BRACE_CLOSE) ||
+             (cpd.settings[UO_mod_case_brace].a & AV_REMOVE)) &&
```
mihaipopescu: maybe it's late but I don't really get this. so we do the indent only if `mod_case_brace` is true ? and why are you changing these along with adding new options ?

src/indent.cpp
```diff
@@ -666,18 +666,44 @@ void indent_text(void)
...
 
-            /* Indent one level except if the #if is a #include guard */
-            int extra = ((pc->pp_level == 0) && ifdef_over_whole_file()) ? 0 : indent_size;
-            frm.pse[frm.pse_tos].indent     = frm.pse[frm.pse_tos - 1].indent + extra;
-            frm.pse[frm.pse_tos].indent_tab = frm.pse[frm.pse_tos - 1].indent_tab + extra;
-            frm.pse[frm.pse_tos].indent_tmp = frm.pse[frm.pse_tos].indent;
-            frm.pse[frm.pse_tos].in_preproc = false;
+            if (should_indent_preproc)
+            {
+                /* Hack to get the logs to look right */
```
mihaipopescu: this doesn't sounds too good :)

src/options.cpp
```diff
@@ -1470,6 +1476,14 @@ void register_options(void)
...
    unc_add_option("pp_define_at_level", UO_pp_define_at_level, AT_BOOL,
                   "Whether to indent '#define' at the brace level (true) or from column 1 (false)");
+   unc_add_option("pp_indent_case", UO_pp_indent_case, AT_BOOL,
+                  "Overrides the indent from pp_if_indent_code for case statements");
+   unc_add_option("pp_indent_func_def", UO_pp_indent_func_def, AT_BOOL,
+                  "Overrides the indent from pp_if_indent_code for function definitions");
+   unc_add_option("pp_indent_extern", UO_pp_indent_extern, AT_BOOL,
+                  "Overrides the indent from pp_if_indent_code for extern C blocks");
+   unc_add_option("pp_indent_brace", UO_pp_indent_brace, AT_BOOL,
+                  "Overrides the indent from pp_if_indent_code for braces");
```
mihaipopescu: the new options should state their default option in the description. I know it's not great but it's what we have.

mihaipopescu:
Good update! Few issues need to be addressed. Also some tests to show how the options behave with non-default values - would be nice!

src/indent.cpp
brmqk3: I can check for NULL here as well.

tests/config/pp_if_indent-1.cfg
brmqk3: Will remove.

src/indent.cpp
brmqk3: The "hack" part? I didn't add that, it was already there.

src/options.cpp
brmqk3: Okay, I will update these.

src/indent.cpp
mihaipopescu: yeah... ok :) it just sounds nasty to hack the code just to make the logs look nice. but if you're just moving stuff around it's cool.

src/indent.cpp
brmqk3: So my thinking here is that not all case statements need braces, right? I could have:
case 3:
    doFive();
    break;
Which is a correct case statement, but this check requires there to be a CT_BRACE_CLOSE. So I guess this would be a bug fix, related to a case I was testing. Let me know if you think this makes sense.

mihaipopescu: ok cool but I guess it would be better to be separate as it doesn't depend on the newly added options. right ?

brmqk3: Sure, I can put it in a separate patch.

gmaurel:
I miss the tests

gmaurel:
Please, push some tests. Thanks.

mihaipopescu:
@brmqk3 Things are moving and conflicts are starting to appear. If you have some time adding a small test would be awesome. The sooner the better ! Thanks.

guy-maurel:
@brmqk3 Would you like to push some tests for your proposal and solve the conflicts?
If not, I would like to close the PR.
Any other message?

PR #660, opened 2016-08-30
Use a fixed size (32 chars) for option alignment

bengardner:
This prevents config files from changing every time a longer option is added.


gmaurel:
The file options.cpp defines at line 103 the maximal length of an option name.
```#define OptionMaxLength    60```

PR #757, opened 2016-09-10
Update Travis script

jibsen:
I am playing around with the Travis script a bit, so atm this PR is just to get build status info.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,58 +1,73 @@
 language: cpp
+sudo: false
 
-env:
-  matrix:
-    # - COMPILER=clang
-    # - COMPILER=clang-3.8
-    - COMPILER=gcc-6
-    - COMPILER=gcc-6 BUILD_TYPE=release
-    - COMPILER=gcc-6 BUILD_TYPE=asan
-    # - COMPILER=gcc-6 BUILD_TYPE=ubsan
-    # - BUILD_TYPE=coverage
+matrix:
+  include:
+    - env: BUILD_TYPE=debug
+    - env: BUILD_TYPE=release
+
+    # GCC 6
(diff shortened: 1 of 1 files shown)


jibsen:
Okay, seems to be working now. This adds a Win64 cross-compile using mingw-w64 to Travis.

I also switched to Clang for the address sanitizer build, and moved the package installs to the build matrix (only install the packages required for each build).

mihaipopescu:
@jibsen Awesome. Thanks for checking this out ! It's really important to have the biggest coverage we can have.

Actually I was thinking about this as I've seen you've also added a Coverage build type but it's still inactive. I think we can really make http://coveralls.io to work with CMake and then put a coverage badge on the first page.

I saw that CMake already has support for Coverage so I started writing a few things.

```cmake
if(BUILD_TESTING)
  enable_testing()
  add_subdirectory(tests)

  if(ENABLE_COVERAGE)
    include(CodeCoverage)
    # debug, no optimisation, enable coverage
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 -fprofile-arcs -ftest-coverage")
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fprofile-arcs -ftest-coverage")
    SETUP_TARGET_FOR_COVERAGE(uncrustify_coverage "ctest" "_coverage")
    target_link_libraries(uncrustify gcov)
  endif()

endif()
```

gmaurel:
Some more comments before I merge?

jibsen:
@mihaipopescu, I think adding coverage would be a good idea. But perhaps it would be better if you start a new PR for working on that.

mihaipopescu:
@gmaurel Looks good to me.
@jibsen Yeah. First I need to have something working.

PR #758, opened 2016-09-13
Add option comment after long class names. Don't add function or clas…

DVE2000:
…s headers for one liners.

You may want to add this into the main branch. I added an option for long class names (similar to long functions or namespaces), and had uncrustify ignore headers for single line function or class declarations. The latter could be made into options, but I didn't see the point.

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -760,9 +760,11 @@ void add_long_closebrace_comment(void)
    chunk_t  *tmp;
    chunk_t  *br_open;
    chunk_t  *br_close;
-   chunk_t  *fcn_pc = NULL;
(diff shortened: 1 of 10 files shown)


mihaipopescu:
I think the changes but there are no tests. Can you add some? Thanks!

DVE2000:
Will do. I need to figure out how to add them :p

DVE2000:
I modified 30265-long_br_cmt.cpp to test mod_add_long_class_closebrace_comment
I also modified the tests that output 30870-cmt_insert.cpp and 30871-cmt_insert.cpp to test that function and class headers are not added for forward declarations.

PR #769, opened 2016-09-27
Cleanup build system

jibsen:
Taking a stab at a PR for cleaning up the build system files as discussed in #604.


jibsen:
While the CI passes, `make_version.py` currently does not work:

  - print needs parenthesis for Python 3 compatibility
  - `scripts/tokenize.py` shadows a system module that `make_version.py` includes
  - it insists on looking for `.git` in the same folder as the script rather than it's working folder

Perhaps @mihaipopescu has a good way to solve these? Or I could just remove the commit that moved it to the scripts folder for now.

gmaurel:
@jibsen 
I am missing the file INSTALL

jibsen:
The [`INSTALL`](https://github.com/uncrustify/uncrustify/blob/85eb60939320a3b3eb2c46dd123f24361a94b72d/INSTALL) file contained generic instructions for building software that uses Autotools. I think the instructions in the README are more useful.

If you need a INSTALL file to comply with some rules for releasing software we could add one pointing to the instructions in the README?

nivekkagicom:
The steps in the README no longer work as autogen.sh, etc. have been removed. So how is one to run autogen.sh or configure which no long exist?

Building using Autotools

Quick start:

$ ./autogen.sh
$ ./configure
$ make
The executable is src/uncrustify. Copy that to your ~/bin/ folder or wherever you want.

jibsen:
The idea is to use CMake to configure the build instead. Work on the CMake solution has been going on for a while, which is why both options are listed in the README. The instructions for Autotools will likely be removed when cleaning up the docs and other non-source files.

A quick start for CMake could be:

~~~.sh
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
make
~~~
The executable is in the build folder, you can copy it where you like, or use `make install` to install it.

mihaipopescu:
Actually I think @nivekkagicom has a good point. I think we should've updated the information from README.md along with this PR. Now the instructions point to non-existing scripts and from the user pov is not clear that cmake has entirely replaced autotools scripts.

PR #770, opened 2016-09-28
Check if an option name is too long.

gmaurel:
This make possible to detect at the first run.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -48,6 +48,13 @@ void unc_add_option(const char *name, uncrustify_options id, argtype_e type,
                     const char *short_desc, const char *long_desc,
                     int min_val, int max_val)
 {
+#define OptionMaxLength 60
+   int lengthOfTheOption = strlen(name);
+   if (lengthOfTheOption > OptionMaxLength) {
+      fprintf(stderr, "FATAL: length of the option name (%s) is too big (%d)\n", name, lengthOfTheOption);
+      fprintf(stderr, "FATAL: the maximal length of an option name is %d characters\n", OptionMaxLength);
+      exit(EXIT_FAILURE);
+   }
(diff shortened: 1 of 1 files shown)


src/options.cpp
```diff
@@ -48,6 +48,13 @@ void unc_add_option(const char *name, uncrustify_options id, argtype_e type,
                     const char *short_desc, const char *long_desc,
                     int min_val, int max_val)
 {
+#define OptionMaxLength 60
+   int lengthOfTheOption = strlen(name);
+   if (lengthOfTheOption > OptionMaxLength) {
+      fprintf(stderr, "FATAL: length of the option name (%s) is too big (%d)\n", name, lengthOfTheOption);
+      fprintf(stderr, "FATAL: the maximal length of an option name is %d characters\n", OptionMaxLength);
+      exit(EXIT_FAILURE);
```
mihaipopescu: I wish we could have this as a compile time assert.

src/options.cpp
```diff
@@ -48,6 +48,13 @@ void unc_add_option(const char *name, uncrustify_options id, argtype_e type,
                     const char *short_desc, const char *long_desc,
                     int min_val, int max_val)
 {
+#define OptionMaxLength 60
```
mihaipopescu: I would have used an anonymous enums for constants

mihaipopescu:
IIRC I think I've seen a similar code from Ben some while ago.

src/options.cpp
gmaurel: How?

src/options.cpp
gmaurel: I don't like anonymous enums!

gmaurel:
I wait a little for you.

PR #782, opened 2016-10-06
Failing test cases for calling convention in bug #633 #602: now fixed

gmaurel:
fixed with a "MS-calling_conventions.cfg" file


tests/config/bug_633.cfg
```diff
@@ -1 +1,8 @@
+indent_columns=4
+indent_with_tabs=0
 sp_arith=add
+indent_class=true
+#
+# bug 633
+#
+include "MS-calling_conventions.cfg"
```
scottbilas: i would call this msvc.cfg instead. config file would be better to apply to the entire compiler, rather than a subset of its keywords.

gmaurel:
a better name and place for the config file

PR #787, opened 2016-10-13
Bug 643 is partially fixed

gmaurel:
The second part of the bug about Q_PROPERTY( ... ) cannot be solved


misery:
Thank you very much! Why it cannot be solved? I don't understand why it is not possible.

gmaurel:
Q_PROPERTY is a macro for Qt,
  http://doc.qt.io/qt-5/properties.html
uncrustify doesn't know what a macro is for, what the parameters are

Issue #682, opened 2016-09-09
Lambda in return statement not recognized


gmaurel:
Seems that lambda in return statement is not recognized. I will upload output/wanted output/cfg later, now just input.h. Cfg is same as default, except:
- sp_arith is set to "add", to show that pointers in lambda arguments is not recognized and treated as multiplication
- sp_assign is set to "add" to show that "[=]" capture list is mistreated as equal
- sp_cpp_lambda_assign set to "remove" as this is what it should be

Reported by: brankokokanovic

Original Ticket: [uncrustify/bugs/652](https://sourceforge.net/p/uncrustify/bugs/652)

PR #791, opened 2016-10-22
better recognize lambda statements

gmaurel:
Fix the issue #682


CDanU:
Just for better documentation - Both commits where merged with: db9478a6d4b1ab4dd82ffefddddc1f305c0ed917.

Issue #687, opened 2016-09-09
Crash when using static if without curly braces


gmaurel:
Crash when a `static if` in a struct does not contain curly braces for its body.

Test-case:

`
struct S
{
    static if (false)
    void bar()
    {
    }
}
`

Error:

`
C:\Apps\uncrustify>uncrustify.exe -c %CD%\cfg\defaults.cfg -f test.d 

Parsing: test.d as language D

test.d:7 Error: Unexpected '}' for 'VBRACE_OPEN', which was on line 3

flag_parens: no match for [{] at [2:1] [CallStack:-DEBUG NOT SET-]
`

The following works ok:

`
struct S
{
    static if (false)
    {
        void bar()
        {
        }
    }
}
`

Version:

`
C:\Apps\uncrustify>uncrustify.exe --version
uncrustify 0.63.0-g44ce0f1
`

Reported by: andrejm08

Original Ticket: [uncrustify/bugs/678](https://sourceforge.net/p/uncrustify/bugs/678)

gmaurel:
The bug is fixed

PR #792, opened 2016-10-23
Bugfix for issue #687

gmaurel:
Prevent the crash. 
Change the source test file to work.


src/combine.cpp
```diff
@@ -11,7 +11,6 @@
 #include "chunk_list.h"
 #include "ChunkStack.h"
 #include "prototypes.h"
-//#define DEBUG
```
mihaipopescu: 👍 good, we need to clean all of them.

Issue #796, opened 2016-10-27
emscripten build


CDanU:
@bengardner, @gmaurel 

Take a look at:
https://github.com/CDanU/uncrustify/commits/emscripten
https://cdanu.github.io/uncrustify_config_preview/index.html

and tell me what you think of this.

--------------------

- [x] prepare for Emscripten 0
- [x] prepare for Emscripten 1
- [x] prepare for Emscripten 2 && merge mistake fix
- [x] prepare for Emscripten 3
- [x] ambiguity fix
- [x] Emscripten Interface

mihaipopescu:
@CDanU This is awesome! I was looking for something like this for uncrustify. clang-format has something. Previously it was a bit cumbersome to play with config options. We were using [UniversalIndentGUI](http://universalindent.sourceforge.net/) but what you did is so much better and faster. Thank you!

mihaipopescu:
I'd like to see a PR (or more) to get those changes upstream. :)

gmaurel:
If the tool is usable direct on my processor (not online), it would be very fine.
I don't like to be always on line to test my sources, might be I don't like to upload all my sources.
The output on the right for "sp_arith" is fine. It would be fine to get such examples for all the options.
It is a better documentation to see what is working on. 
I use, for my debugging, the line numbers of the uigui_uncrustify.ini (with universalindentgui),
I only need to put my cursor on the area of the option.
As a developer, I haave option(s) you don't know about. I need to test that. 
I find the display  (with universalindentgui) left the options (such as your display) AND on the right a display of my (changed) source.

CDanU:
Both links are separate projects. 
The first link contains everything that is needed to compile the Uncrustify C++ code into JavaScript.
I would hope the changes that where made in this repo could be merged into the Uncrustify repo.

The second link on the other hand is just an example of what could be done with the JS port. It will not be merged into Uncrustify itself. It is just a preview for now as it is lacking examples, option dependencies, some comport functions and a more automated build process.
At some point this example could be used to aid in development but for now it is targeted at users that have no or an incomplete Uncrustify config file.

As of now the JS port is set up to work completely without the need of a webserver. Later I have to figure out how to enable higher Emscripten optimization settings to significantly reduce the filesize of the generated JS lib ( from 5mb to 1mb ). The higher optimization settings require a server ( like a local node server ) to load a 'pre-compiled' blob. This still would let the execution of the formatting run only locally.

[...]

PR #801, opened 2016-11-01
Emscripten Build

CDanU:
see issue #796


src/emscripten/libUncrustify.d.ts
```diff
@@ -0,0 +1,823 @@
+interface EmscriptenVector< T >
```
mihaipopescu: Is this generated ?

src/options.h
```diff
@@ -843,15 +843,15 @@ struct option_map_value
    argtype_e          type;
    int                min_val;
    int                max_val;
-   const char         *name;
-   const char         *short_desc;
-   const char         *long_desc;
+   string             name;
```
mihaipopescu: Why do you really need this as a string while the data is static ?

mihaipopescu:
We may have to hold on as this is quite a consistent change and we need time to digest all this.

src/uncrustify_emscripten.cpp
```diff
@@ -0,0 +1,1312 @@
...
+* @return formated string
+*/
+string uncrustify( string file )
+{
+    // overload for default frag parameter
+    return uncrustify( file, false );
+}
+
+
+EMSCRIPTEN_BINDINGS( MainModule )
```
CDanU: see above

src/emscripten/libUncrustify.d.ts
CDanU: This is a declaration file for Typescript. It helps working with the generated Emscripten library in a Typescrript environment. 

It is was also created by hand and the enum values should also be generated at some point.

src/options.h
CDanU: It is not really needed to work with strings here but it makes everything so much easier when one can avoiding pointers in emscripten.

gmaurel:
Why do you need the changes?

CDanU:
Could you specify which changes? 

- 827362c, 1dae768 and 6542fce are just some random improvements I had made, those are not needed but improve readability
- 93c9a73 Since I cannot work with the C FILE type, this allows me to parse the config file line by line as a string
- cdae69b I don't see why option_name_map is using a string as index, the enum value should be more suitable here and it also allows an better/easier interface for the js lib
- 3f358d1  Here also I don't get why a list is used, a vector should be more suitable and Emscripten has already bindings for the vector structure implemented.
- 3f0022e like 93c9a73 this is used to read a passed config file as a string, although I guess I could move that function into the Emscripten interface file
- ce2a89a will be replaced with something non intrusive

Everything else is either part of the Emscripten interface file or removes static specifiers in order to allow the Emscripten interface file to use functions.

gmaurel:
Sorry, wrong question.
I try again:
What is the relation uncrustify <--> emscripten

PR #804, opened 2016-11-08
small cleanup

CDanU:


gmaurel:
We try to have "uncrustify-ed" sources.
Use the config file ```forUncrustifySources.cfg``` before making a PR.
If necessary, make a PR to amend the config file.

PR #805, opened 2016-11-08
prepare for Emscripten 0 - extract line processing code into its own function

CDanU:
ref: #796


CDanU:
All PRs with the same issue have been rebased. They should have the right formatting.

gmaurel:
Look at line 51 of prototypes.h
the star is not at the right place
Use the config file forUncrustifySources.cfg to uncrustify the source

PR #813, opened 2016-11-11
Add a script to uncrustify the sources

gmaurel:
The uncrustify sources will be uncrustify-ed with the config file forUncrustifySources.cfg.
The comparsion of the output with the original should show "No change".
It might be necessary to amend the config file to take care of new development(s).
In this last case, a PR for the config file is to be done.

### DIFF
--- a/scripts/Run_uncrustify_for_sources.sh
+++ b/scripts/Run_uncrustify_for_sources.sh
@@ -14,9 +14,11 @@ cd ${where}
 #
 mkdir -p results
 #
+find . -name uncrustify
+ls -l ./build/uncrustify
 for file in ${list_of_H} ${list_of_C}
 do
-  .build/uncrustify -q -c ./forUncrustifySources.cfg -f ${SRC}/${file} -o results/${file}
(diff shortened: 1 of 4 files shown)


scripts/Run_uncrustify_for_sources.sh
```diff
@@ -0,0 +1,31 @@
...
+list_of_H=`ls *.h`
+list_of_C=`ls *.cpp`
+cd ${where}
+#
+mkdir -p results
+#
+for file in ${list_of_H} ${list_of_C}
+do
+  uncrustify -q -c ./forUncrustifySources.cfg -f ${SRC}/${file} -o results/${file}
+  diff ${SRC}/${file} results/${file} > /dev/null 2>&1
```
CDanU: A `cmp -s ${SRC}/${file} results/${file}` should be better here if you just want to compare both files without getting any metrics.
Additionally a check for errors should be done as well: 

> 0 if inputs are the same, 1 if different, 2 if trouble

PR #814, opened 2016-11-11
This is the first part of the PR#70, published by

gmaurel:
rdan on  12 May 2012
This affect the tests 02000, 02002 and 30023

### DIFF
--- a/tests/output/c/02000-i2c-core.c
+++ b/tests/output/c/02000-i2c-core.c
@@ -795,9 +795,9 @@ static int i2c_probe_address(struct i2c_adapter *adapter, int addr, int kind,
    return((err == -ENODEV) ? 0 : err);
 }
 
-int i2c_probe(struct i2c_adapter *adapter,
+int i2c_probe(struct i2c_adapter             *adapter,
               struct i2c_client_address_data *address_data,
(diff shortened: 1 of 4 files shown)


tests/output/c/02000-i2c-core.c
```diff
@@ -795,9 +795,9 @@ static int i2c_probe_address(struct i2c_adapter *adapter, int addr, int kind,
    return((err == -ENODEV) ? 0 : err);
 }
 
-int i2c_probe(struct i2c_adapter *adapter,
+int i2c_probe(struct i2c_adapter             *adapter,
               struct i2c_client_address_data *address_data,
-              int (*found_proc)(struct i2c_adapter *, int, int))
+              int (                          *found_proc)(struct i2c_adapter *, int, int))
```
mihaipopescu: I'm not entirely sure if this space is really what we want in this case.

gmaurel: mhhhh!
What should be correct?

mihaipopescu: I don't think the alignment should be done in this case so no change should be the correct approach here.

PR #818, opened 2016-11-15
Remove some warnings, part 2

gmaurel:


src/align.cpp
```diff
@@ -851,7 +851,7 @@ static void align_same_func_call_params()
    deque<chunk_t *>  chunks;
    deque<AlignStack> as;
    AlignStack        fcn_as;
-   int               idx;
+   size_t            idx;
```
CDanU: If you want to be 100% correct on this one it should be `deque<AlignStack>::size_type`.

src/chunk_list.cpp
```diff
@@ -207,7 +207,7 @@ void chunk_move_after(chunk_t *pc_in, chunk_t *ref)
 
    /* HACK: Adjust the original column */
    pc_in->column       = ref->column + space_col_align(ref, pc_in);
-   pc_in->orig_col     = pc_in->column;
+   pc_in->orig_col     = (UINT32)pc_in->column;
```
CDanU: A static_cast would be nicer here.

src/uncrustify_types.h
```diff
@@ -263,7 +263,7 @@ struct chunk_t
    }
 
 
-   int len()
+   size_t len()
```
CDanU: This and the return type of size() in unc_text.h should be deque<int>::size_type.

src/align.cpp
gmaurel: http://www.cplusplus.com/reference/deque/deque/
wrote:
size_type	an unsigned integral type that can represent any non-negative value of difference_type	usually the same as size_t

src/chunk_list.cpp
gmaurel: Correct. BUT
it is an interim solution, until I have done the work on all the files

Issue #815, opened 2016-11-14
package.json file


mihaipopescu:
@shazron The npm package.json file needs an update as some things have changed:

I've identified some things (can be more):
- [x] https://github.com/uncrustify/uncrustify/blob/master/package.json#L3 => update version
- [x] https://github.com/uncrustify/uncrustify/blob/master/package.json#L7 => replace with cmake
- [x] https://github.com/uncrustify/uncrustify/blob/master/package.json#L12 => update link

Also, if you can add some instructions in README.md on how to use it would be great.

CDanU:
I assume this package.json is for npm, right? I am not 100% on this but isn't npm used for JavaScript Packages?

PR #822, opened 2016-11-17
package.json --> 0.64

gmaurel:
fix for #815

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A highly configurable, easily modifiable source code beautifier for C, C++, C#, ObjectiveC, D, Java, Pawn and VALA",
   "main": "uncrustify",
   "scripts": {
-      "preinstall": "mkdir build; cd build; cmake .."
+      "preinstall": "mkdir -p build && cd build && cmake .. && make"
   },
   "bin" : "src/uncrustify" ,
   "repository": {
@@ -16,5 +16,5 @@
     "source code beautifier"
   ],
   "author": "Ben Gardner",
-  "license": "GPLv3"
+  "license": "GPL v2+"
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,20 +1,20 @@
 {
   "name": "uncrustify",
-  "version": "0.6.1",
+  "version": "0.64",
   "description": "A highly configurable, easily modifiable source code beautifier for C, C++, C#, ObjectiveC, D, Java, Pawn and VALA",
   "main": "uncrustify",
   "scripts": {
-      "preinstall": "./configure && make"
+      "preinstall": "mkdir build; cd build; cmake .."
```
CDanU: With this cmake will run even if mkdir and / or cd fail.
This should fix it: `mkdir -p build && cd build && cmake ..` 
The make step is missing too but I am not sure if this should be inside a script called preinstall in the first place.

package.json
```diff
@@ -1,20 +1,20 @@
...
     "type": "git",
-    "url": "https://github.com/bengardner/uncrustify"
+    "url": "https://github.com/uncrustify/uncrustify"
   },
   "keywords": [
     "uncrustify",
     "source code beautifier"
   ],
   "author": "Ben Gardner",
   "license": "GPLv3"
```
CDanU: I am uncertain about this. This repo states GPLv2 is used. inside the file [uncrustify.cpp](https://github.com/uncrustify/uncrustify/blob/d0d3a32b5ce1e871905305f05012554995c0b588/src/uncrustify.cpp#L6) GPLv2**_+_** is mentioned. 
Does that mean Uncrustify is licensed under `GPLv[2, ∞)` ?

PR #830, opened 2016-11-22
temporarily fixes ambiguity problems for the emscripten build

CDanU:
Please merge only either this PR or PR #829 

-------------------


mihaipopescu:
We should avoid making `#ifdef`s in our code base as much as possible - especially in places where it should be _platform_ independent code like this string class.

gmaurel:
we prefer PR 829

PR #835, opened 2016-11-23
Modifications to get less warnings.

gmaurel:


mihaipopescu:
Sorry for the late message. In general when I'm doing these sort of fixes I know they are not be regarded as trivial. Casting to is not always the best solution and also implicit integer promotions can be a real evil sometimes and can generate all sorts of strange bugs that are hard to discover and debug.

gmaurel:
Thanks for reading. I appreciate very much.
I let ```ctest``` every time running. If there where no errors, I think it is OK.
If not, I look for the bug.
We need indeed more and more tests pattern.

gmaurel:
There is at least one error at logger

PR #836, opened 2016-11-25
Less warnings, part 2

gmaurel:
some more will come soon.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -18,7 +18,7 @@
 #include <cerrno>
 #include "unc_ctype.h"
 
-static chunk_t *align_var_def_brace(chunk_t *pc, int span, int *nl_count);
+static chunk_t *align_var_def_brace(chunk_t *pc, size_t span, size_t *nl_count);
 static chunk_t *align_trailing_comments(chunk_t *start);
 static void align_init_brace(chunk_t *start);
 static void align_func_params();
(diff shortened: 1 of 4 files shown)


src/align.cpp
```diff
@@ -1418,18 +1418,19 @@ static CmtAlignType get_comment_align_type(chunk_t *cmt)
...
+   size_t       min_orig = 0;
    chunk_t      *pc      = start;
-   int          nl_count = 0;
+   size_t       nl_count = 0;
    ChunkStack   cs;
    CmtAlignType cmt_type_start, cmt_type_cur;
-   int          col;
-   int          intended_col = cpd.settings[UO_align_right_cmt_at_col].n;
+   size_t       col;
+   size_t       intended_col = cpd.settings[UO_align_right_cmt_at_col].n;
```
CDanU: All number options are ints, some options have special properties when values are below zero. Someone could set the align_right_cmt_at_col to negative values in their config and underflow the size_t. 
If size_t is better suited here a `max(0, cpd.settings[UO_align_right_cmt_at_col].n);` could be used to prevent the underflow.

A thing that I noticed is that min_val and max_val in option_map_value are not used. This values could be used to prevent out of rang values while parsing the config.

src/align.cpp
```diff
@@ -422,12 +422,12 @@ void align_right_comments(void)
          /* Change certain WHOLE comments into RIGHT-alignable comments */
          if (pc->parent_type == CT_COMMENT_WHOLE)
          {
-            int max_col = pc->column_indent + cpd.settings[UO_input_tab_size].n;
+            size_t max_col = pc->column_indent + cpd.settings[UO_input_tab_size].n;
```
CDanU: Same if: pc->column_indent < abs(cpd.settings[UO_input_tab_size].n)

src/align.cpp
gmaurel: It is still an interim.
I plan to introduced a variant, such as ```cpd.settings[UO_xyz].u``` for the options, which values are not negativ.
The work is soooo big as I only want to make little changes each PR.
My test reference are all the tests we have today.
It might be necessary to create new tests. Please create "many" PR for that, it will be fine.

PR #855, opened 2016-12-08
Disable indent in #pragma asm/endasm and #asm/#endasm sections

xeyownt:
Some compiler supports adding inline assembly code using #asm/#endasm
directives. For instance:
```
    #asm
        sll     a
        jrc     sub_1f      ; subtract $1f if A.x msb is set
        ret
    sub_1f:
        xor     a,#$1f
    #endasm
```
Another way is to use #pragma asm/endasm directives:
```
    #pragma asm
        XREF     _my_var
        PUSH     #LOW(my_func)
        PUSH     _my_var+0
        RETF
    App_CallEnd:
    #pragma endasm
```
This commit disables parsing of these section, much like when using the
keywords `*INDENT-OFF*` / `*INDENT-ON*`.

The way we detect the end of a section is limited. Since parsing is disabled, we
cannot rely on it. Instead we do a simple string search on either the word
`#endasm` or the words `#pragma` and `endasm`.

Ideally the search should use regular expression, and the section begin/end
keywords should be user configurable.

The commit adds a new test to cover nominal cases and a few corner cases.


src/tokenize.cpp
```diff
@@ -1417,6 +1417,16 @@ static bool parse_ignored(tok_ctx &ctx, chunk_t &pc)
       /* end of file? */
       return(false);
    }
+   /* HACK: turn on if we find '#endasm' or '#pragma' and 'endasm' separated by blanks */
+   if ( ( ( (pc.str.find("#pragma ") >= 0) || (pc.str.find("#pragma	") >= 0) ) &&
```
mihaipopescu: This is very hacky. :) Why do you have to do this?

src/tokenize.cpp
```diff
@@ -1913,6 +1935,21 @@ void tokenize(const deque<int> &data, chunk_t *ref)
             cpd.preproc_ncnl_count++;
          }
 
+         /* Figure out the type of pragma for enable/disable indentation */
+         if (cpd.in_preproc == CT_PP_PRAGMA)
+         {
+            if (memcmp(pc->text(),"asm",3) == 0)
```
mihaipopescu: there is an util function that does just that. you can use that one.

src/keywords.cpp
```diff
@@ -83,6 +83,7 @@ static const chunk_tag_t keywords[] =
    { "and_eq",           CT_SASSIGN,       LANG_C | LANG_CPP                                                           },
    { "as",               CT_AS,            LANG_CS | LANG_VALA                                                         },
    { "asm",              CT_ASM,           LANG_C | LANG_CPP | LANG_D                                                  },
+   { "asm",              CT_PP_ASM,        LANG_ALL | FLAG_PP                                                          },
```
mihaipopescu: I don't think this is available in all languages. Can we just start with C/C++ ?

src/tokenize.cpp
```diff
@@ -1902,6 +1912,18 @@ void tokenize(const deque<int> &data, chunk_t *ref)
          cpd.preproc_ncnl_count = 0;
       }
 
+      /* Enable/disable indentation between #asm / #endasm diretives */
+      if (pc->type == CT_PP_ASM )
+      {
+         LOG_FMT(LBCTRL, "Found a directive %s on line %d\n","#asm",pc->orig_line);
+         cpd.unc_off = true;
```
mihaipopescu: Disabling the entire processing would be very an extreme approach. Uncrustify already supports formatting of the assembly code. Wouldn't that be a better approach ?

src/tokenize.cpp
mihaipopescu: Why do you need to do this here as well ? Also, this check is not guarded by any new option so it will be called all the times.

gmaurel:
Please, before commiting a change, be sure you have use all the new tests. Look at ./scripts
Here the error link:
```https://travis-ci.org/uncrustify/uncrustify/jobs/182407603```

**Comment by xeyownt** [comment]
Hello,
Thank you for the detailed review and the CI feedback. I'll read and analyze all this as soon as possible.

**Comment by guy-maurel** [comment]
@xeyownt Any new message?

**Comment by xeyownt** [comment]
Hello,

Sorry for being off so long. Many parallel lives to manage ;-)

I checked the travis log, and it turns out the failure is not in my branch, but in a merge commit (merge with c407f41). Actually this merge commit is nowhere to be found so I had to create it myself (where I noticed that `-Xignore-all-space` is used for the merge).

The failure is not in a test case, but rather in a script that uncrustifies the sources, and check for differences. Failure indicates that there is a difference between original and uncrustified sources. Two files have differences: `src/tokenize.c` and `src/keywords.c`.

Now I don't know how to proceed because:
* For `src/keywords.c`, the file is actually correctly indented in the source branch, but not anymore in the merge commit because columns are indented differently in the target branch.
* For `src/tokenize.c`, it is correct I did not run uncrustify on that file (I did not know, albeit obvious I guess). But now, even if I do so, I still get differences because the config file `forUncrustifySources.cfg` is different between both branches.

I understand the logic of the `Run_uncrustify_for_sources.sh`, but to my opinion, a failure should not lead to reject because there are actually no way for the submitter to guarantee a non-rejection. It is like chasing a moving target. Also failure does not indicate that the code is faulty.

Also, it was not easy for me to find out all the above. I wonder if there is some "script", or recipe that allows one to reproduce exactly the travis checks. Like a "pull request procedure" or something. I checked the readme, etc, but could not find that.

Finally, I'm going to go through comments from @mihaipopescu above, and try to improve the commit.

File: src/tokenize.cpp (hunk 1 above)
Comment (xeyownt) [inline src/tokenize.cpp]: Actually we have to enable back parsing if line contains directive `#pragma endasm` or `#endasm`. Ideally we should match any valid variants like leading spaces or tabs, or additional spaces or tabs between `#pragma` and `endasm`. This could be fixed by using a regexp engine, but I found none in the code, and even then, it might have negative impact on performance. So the "hack" above is a poorman's regexp implementation.

File: src/tokenize.cpp (hunk 4 above)
Comment (xeyownt) [inline src/tokenize.cpp]: Yes, sure! If we can dynamically change the formatting language, it would be much better. Also, it might also improve detection of `#pragma endasm` and `#endasm`.
I'll investigate if/how this can be done.

File: src/tokenize.cpp (hunk 2 above)
Comment (xeyownt) [inline src/tokenize.cpp]: Sorry but I don't see the util function you are referring to.

Not sure what you meant by "here as well". The first checks (starting line 1913) are for detecting directive `#asm`. The second checks (starting  like 1936) are for detecting directive `#pragma asm`.

Now, maybe you ask why I test for the end of section (i.e. directives `#endasm` and `#pragma endasm`). Reading the code again, maybe this is in fact not needed, since only effect is to reset the flag `cpd.unc_off`, which was done already.

File: src/keywords.cpp (hunk 3 above)
Comment (xeyownt) [inline src/keywords.cpp]: Ok to limit with C/C++.

**Comment by CDanU** [comment]
You need to format your sources with a freshly build Uncrustify from the master with the newest forUncrustifySources.cfg. This forces you a bit to stay up to date with the master. 

You can either rebase your feature branch on the master format everyting normally or clone the Uncrustify repo in a temp location, build there Uncrustify and with it do the formatting on your files.
Currently you have no merge conflicts so rebaseing shouldn't be a problem.

**Comment by gmaurel** [comment]
You have only a little problem with the source:
use: diff ./src/tokenize.cpp results/tokenize.cpp to see what is to do.
After that, run the script again, to test:
./scripts/Run_uncrustify_for_sources.sh

**Comment by xeyownt** [comment]
@CDanU github does not display the merge conflicts because it uses option <code>-Xignore-all-space</code> to do the merge. Actually the difference is 1-space indent change in a file.

**Comment by xeyownt** [comment]
Cool, all checks have passed. This new commit fixes the CI failures, and also remove a redundant check to re-enable indentation, as observed in review above.

Note that this commit does not address the suggestion of switching to ASM indentation between #asm / #endasm. This does not look straightforward. So I suggest to address this point in a separate feature request.

File: src/keywords.cpp [hunk 5]
```diff
@@ -152,6 +153,7 @@ static const chunk_tag_t keywords[] =
    { "else",             CT_PP_ELSE,       LANG_ALL | FLAG_PP                                                          }, // PAWN
    { "elseif",           CT_PP_ELSE,       LANG_PAWN | FLAG_PP                                                         }, // PAWN
    { "emit",             CT_PP_EMIT,       LANG_PAWN | FLAG_PP                                                         }, // PAWN
+   { "endasm",           CT_PP_ENDASM,     LANG_ALL | FLAG_PP                                                          },
```
CDanU: I don't see this one used in the general change set . Do you still need it here?

src/token_enum.h
```diff
@@ -270,6 +270,8 @@ typedef enum
    CT_PP_FILE,
    CT_PP_LINE,
    CT_PP_SECTION,
+   CT_PP_ASM,           /* start of assembly code section */
+   CT_PP_ENDASM,        /* end of assembly code section */
```
CDanU: same

src/keywords.cpp
CDanU: Don't forget ^

src/keywords.cpp
xeyownt: No indeed, the test was not necessary, and so neither is this keyword. I guess it will become necessary if we can switch to ASM indent in these section.

xeyownt:
New commit removing unnecessary endasm keyword/token as identified by @CDanU.

xeyownt:
Hi Guy,

Thanks for the merge. 

I just noticed that I'm using a test case number that is not in range for 'c' language (10004, where range is apparently 00000-09999). If you want I can create a push request to fix that (using for instance 02511). See my commit xeyownt/uncrustify@a690e96, or you can fix that with:

``` bash
sed -ri 's/10004/02511/' tests/c.test
mv tests/output/c/10004-pragma_asm.c tests/output/c/02511-pragma_asm.c
```

PR #860, opened 2016-12-14
using void for functions without parameters

stefan-nu:
A few functions that use no parameters had an empty paramter list.
I have made this explicitely by adding void. This makes the code more consistent.

For C++ this makes no functional difference. But for people like me who program in C for ages it makes a difference as an empty parameter list in C means the function accepts any number of parameters.


CDanU:
The no function param voids seem to be remains from C code. Today Uncrustify is a C/C++ mix and I would rather remove them completely.

PR #861, opened 2016-12-14
Several minor code improvements

stefan-nu:


src/tokenize.cpp
```diff
@@ -11,6 +11,7 @@
 #include "char_table.h"
 #include "prototypes.h"
 #include "chunk_list.h"
+#include <stdbool.h>
```
CDanU: Hmm, why do you need this? 

The build fails because:
> ./scripts/Run_uncrustify_for_sources.sh; fi" exited with 1

Your code formatting somewhere does not align with the standards of this repo. Use Run_uncrustify_for_sources.sh to find out which files are missformatted and Uncrustify them with the forUncrustifySources.cfg (in the project dir) config.

gmaurel: we wanted to have sources which all are "uncrustify-ed" them self with the actual config file.

stefan-nu: @CDanU you are right the stdbool include is not necessary. I added it when I saw the linker error in the release build. I'll remove the unnecessary include.

src/align.cpp
```diff
@@ -21,13 +21,13 @@
 static chunk_t *align_var_def_brace(chunk_t *pc, size_t span, size_t *nl_count);
 static chunk_t *align_trailing_comments(chunk_t *start);
 static void align_init_brace(chunk_t *start);
-static void align_func_params();
+static void align_func_params(void);
```
bengardner: I'm curious as to why you are adding ```void``` to these prototypes.
It seems the C++ way to is leave that out.

http://stackoverflow.com/questions/7412274/why-add-void-to-method-parameter-list

todd-richmond: some c++ compilers still complain about the old C syntax when set to their highest warning level and void shuts them up

stefan-nu: @bengardner The reason I did it was just as described in the link you provided.
> In C++, foo() and foo(void) both mean "a function with zero parameters", but some people prefer the second form because it is more explicit.

I prefere the explicit way as it makes clear what the programmer's intention was.
But I understand that people who only program in C++ do not see any use for it. 
So if you prefer the empty parenthesis I am glad with that too. But I would propose to stick to one stile throughout the code.

PR #866, opened 2016-12-15
batch script to check source files with uncrustify under windows

stefan-nu:
The new batch script reproduces the functionality of "Run_uncrustify_for_sources.sh". It's useful for those who work on windows.

### DIFF
--- a/scripts/Run_uncrustify_for_sources.bat
+++ b/scripts/Run_uncrustify_for_sources.bat
@@ -1,0 +1,91 @@
+@echo off
+setlocal
+
+rem
+rem 14.12.2016
+rem
+rem Checks if all source files conform
+rem to the uncrustify configuration
+rem The check is performed with the uncrustify
+rem program that is available in the path.
+rem This might be an older version than
+rem a newly build one.
+rem 
+rem Call this batch script from the top level
+rem directory of the uncrustify project.
+
+set SRC_DIR=src
+set OUT_DIR=results
+set CFG_FILE=forUncrustifySources.cfg
+set DIFF_FILE=lastdiff.txt
+set "PWD=%cd%"
+
+rem Check if uncrustify is available in path
+set prog=uncrustify.exe
+for %%i in ("%path:;=";"%") do (
+    if exist %%~i\%prog% (
+		set found=%%i
+rem		echo found %prog% in %%i
+	)
+)
+
+rem Check if fc (file compare tool) is available in path
+set CMP=fc.exe
+for %%i in ("%path:;=";"%") do (
+    if exist %%~i\%CMP% (
+		set found=%%i
+rem		echo found %prog% in %%i
+	)
+)
+
+if %found%=="" goto PROG_MISSING
+
+rem build the list of source files
+set SRC_LIST=%PWD%\%OUT_DIR%\files.txt
+dir /b %SRC_DIR%\*.h   >  %SRC_LIST%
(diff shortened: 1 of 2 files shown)


scripts/Run_uncrustify_for_sources.bat
```diff
@@ -0,0 +1,91 @@
...
+set OUT_DIR=results
+set CFG_FILE=forUncrustifySources.cfg
+set DIFF_FILE=lastdiff.txt
+set "PWD=%cd%"
+
+rem Check if uncrustify is available in path
+set prog=uncrustify.exe
+for %%i in ("%path:;=";"%") do (
+    if exist %%~i\%prog% (
+		set found=%%i
```
CDanU: I guess here a break would be nice if _.bat_ scripts have that.

scripts/Run_uncrustify_for_sources.bat
```diff
@@ -0,0 +1,91 @@
...
+rem		echo found %prog% in %%i
+	)
+)
+
```
CDanU: Shouldn't here also be a
```
if %found%=="" goto PROG_MISSING
```
check ?

scripts/Run_uncrustify_for_sources.bat
```diff
@@ -0,0 +1,91 @@
...
+rem Call this batch script from the top level
+rem directory of the uncrustify project.
+
+set SRC_DIR=src
+set OUT_DIR=results
+set CFG_FILE=forUncrustifySources.cfg
+set DIFF_FILE=lastdiff.txt
+set "PWD=%cd%"
+
+rem Check if uncrustify is available in path
```
CDanU: I am not sure about the search in the PATH var. 
The sources should be always uncrustified with a freshly build Uncrustify or else you are uncrustifying with possibly older (faulty) releases/builds. 

Relying that there is a build dir in the project dir with a freshly build Uncrustify is also not the best solution. Could this maybe be executed with CMake after a build? CMake does know where the generated binary is.

stefan-nu: Thanks for picking to up. I was thinking about it too.

I think there are two use cases:
1. Checking if a changed uncrustify code corresponds to the expected format -> this is for uncrustify developers
2. Checking if the code of any other project corresponds to the expected format -> this is for uncrustify users

For 1. it might be more appropriate to use the newly build uncrustify binary. Even tough this might fail if in the new binary introduced a new bug. So there is also a good reason to use an older uncrustify version that is know to work.

For 2. Usually there is no uncrustify source project present but an uncrustify binary. So in this case searching the path seems more appropriate to me.

The best solution would probably be to use a command line parameter for the script to specifiy which uncrustify binary to use. If the parameter is missing the path is searched as fallback. 

I think we should propose the scripts Run_uncrustify_for_sources.bat/sh to uncrustify users that simply want to check there own projects. Maybe adding a comment to README.md?

scripts/Run_uncrustify_for_sources.bat
stefan-nu: Yes, but I didn't know how to do it. My batch programming skills are rather limited :-(

CDanU: Wow, I looked it up, bat scripts actually do not have break statements. A solution would be to use goto but apparently that could, in some way, cause errors too as MS did to intend to let loops end prematurely.

Issue #870, opened 2016-12-16
use variables with minimal scope


stefan-nu:
Throughout the code there are many variables that have unecessary large scope. In my opinion it is good practice to keep the scope of variables as small as possible. This greatly improves code readability and reduces the risk for bugs.

An example from args.cpp

 ```C
   int idx;

   if (token != NULL)
   {
      for (idx = 0; idx < m_count; idx++)
      {
         if (strcmp(token, m_values[idx]) == 0)
         {
            SetUsed(idx);
            return(true);
         }
      }
   }
 ```
 would become

 ```C
   if (token != NULL)
   {
      for (int idx = 0; idx < m_count; idx++)
      {
         if (strcmp(token, m_values[idx]) == 0)
         {
            SetUsed(idx);
            return(true);
         }
      }
   }
```
The approach to define all variables at the head of a function was necessary in the old days where compiler where stupid. But every decend compiler nowadays accepts later variable definitions.

Would you agree if I reduced the scope of the variables?

gmaurel:
It is a fine idea.

stefan-nu:
I think the objective of this issue was reached with the various commits. Thus I close the issue.

PR #871, opened 2016-12-16
#870 reduced the scope of many variables, no functional change

stefan-nu:
The intention of this commit was to reduce the scope of the variables as much as possible.
Thus all variables are as close as possible to the place where they are used.

The actual functionality was not changed. Or at least it was not my intention.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -805,13 +805,11 @@ static void align_func_params(void)
 static void align_params(chunk_t *start, deque<chunk_t *> &chunks)
 {
    LOG_FUNC_ENTRY();
-   chunk_t *pc       = start;
-   bool    hit_comma = true;
+   bool hit_comma = true;
 
    chunks.clear();
 
-   pc = chunk_get_next_type(start, CT_FPAREN_OPEN, start->level);
-
+   chunk_t *pc = chunk_get_next_type(start, CT_FPAREN_OPEN, start->level);
    while ((pc = chunk_get_next(pc)) != NULL)
    {
       if (chunk_is_newline(pc) ||
@@ -1639,8 +1637,7 @@ static void align_log_al(log_sev_t sev, size_t line)
    if (log_sev_on(sev))
    {
       log_fmt(sev, "%s: line %lu, %lu)", __func__, line, cpd.al_cnt);
-      size_t idx;
-      for (idx = 0; idx < cpd.al_cnt; idx++)
+      for (size_t idx = 0; idx < cpd.al_cnt; idx++)
       {
          log_fmt(sev, " %lu/%lu=%s", cpd.al[idx].col, cpd.al[idx].len,
                  get_token_name(cpd.al[idx].type));
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -205,15 +205,12 @@ static void align_add(ChunkStack &cs, chunk_t *pc, size_t &max_col, size_t min_p
 void quick_align_again(void)
 {
    LOG_FUNC_ENTRY();
-   chunk_t    *pc;
-   chunk_t    *tmp;
-   AlignStack as;
-
    LOG_FMT(LALAGAIN, "%s:\n", __func__);
-   for (pc = chunk_get_head(); pc != NULL; pc = chunk_get_next(pc))
+   for (chunk_t *pc = chunk_get_head(); pc != NULL; pc = chunk_get_next(pc))
    {
       if ((pc->align.next != NULL) && (pc->flags & PCF_ALIGN_START))
       {
+         AlignStack as;
          as.Start(100, 0);
          as.m_right_align = pc->align.right_align;
          as.m_star_style  = (AlignStack::StarStyle)pc->align.star_style;
@@ -223,7 +220,7 @@ void quick_align_again(void)
          LOG_FMT(LALAGAIN, "   [%s:%lu]", pc->text(), pc->orig_line);
          as.Add(pc->align.start);
          chunk_flags_set(pc, PCF_WAS_ALIGNED);
-         for (tmp = pc->align.next; tmp != NULL; tmp = tmp->align.next)
+         for (chunk_t *tmp = pc->align.next; tmp != NULL; tmp = tmp->align.next)
          {
             chunk_flags_set(tmp, PCF_WAS_ALIGNED);
             as.Add(tmp->align.start);
@@ -239,14 +236,12 @@ void quick_align_again(void)
 void quick_indent_again(void)
 {
    LOG_FUNC_ENTRY();
-   chunk_t *pc;
-   chunk_t *tmp;
 
-   for (pc = chunk_get_head(); pc; pc = chunk_get_next(pc))
+   for (chunk_t *pc = chunk_get_head(); pc; pc = chunk_get_next(pc))
    {
       if (pc->indent.ref)
       {
-         tmp = chunk_get_prev(pc);
+         chunk_t *tmp = chunk_get_prev(pc);
          if (chunk_is_newline(tmp))
          {
             int col = pc->indent.ref->column + pc->indent.delta;
@@ -344,13 +339,12 @@ void align_all(void)
 static void align_oc_msg_spec(int span)
 {
    LOG_FUNC_ENTRY();
-   chunk_t    *pc;
-   AlignStack as;
-
    LOG_FMT(LALIGN, "%s\n", __func__);
+
+   AlignStack as;
    as.Start(span, 0);
 
-   for (pc = chunk_get_head(); pc != NULL; pc = chunk_get_next(pc))
+   for (chunk_t *pc = chunk_get_head(); pc != NULL; pc = chunk_get_next(pc))
    {
       if (chunk_is_newline(pc))
       {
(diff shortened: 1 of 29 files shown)


src/options.cpp
```diff
@@ -20,6 +20,59 @@
...
+   "#\n"
+   "# all tokens are separated by any mix of ',' commas, '=' equal signs\n"
+   "# and whitespace (space, tab)\n"
+   "#\n"
+   "# You can add support for other file extensions using the 'file_ext' command.\n"
+   "# The first arg is the language name used with the '-l' option.\n"
+   "# The remaining args are file extensions, matched with 'endswith'.\n"
+   "#   file_ext CPP .ch .cxx .cpp.in\n"
+   "#\n";
+
```
CDanU: This was already merged into the master with 2d4d38eb8c40206959607ce3695ce0ed29edec6b

src/options.cpp
```diff
@@ -2064,51 +2115,7 @@ int save_option_file_kernel(FILE *pfile, bool withDoc, bool only_not_default)
 
    if (withDoc)
    {
-      fprintf(pfile,
```
CDanU: same

src/output.cpp
```diff
@@ -178,12 +178,11 @@ static bool next_word_exceeds_limit(const unc_text &text, int idx)
  */
 static void output_to_column(int column, bool allow_tabs)
 {
-   int nc;
-
    cpd.did_newline = 0;
    if (allow_tabs)
    {
       /* tab out as far as possible and then use spaces */
+	  int nc;
```
CDanU: Here the formatting is off.

src/parse_frame.cpp
```diff
@@ -84,12 +84,12 @@ void pf_copy(struct parse_frame *dst, const struct parse_frame *src)
...
 void pf_push(struct parse_frame *pf)
 {
-   static int ref_no = 1;
-
    if (cpd.frame_count < (int)ARRAY_SIZE(cpd.frames))
    {
       pf_copy(&cpd.frames[cpd.frame_count], pf);
       cpd.frame_count++;
+
+      static int ref_no = 1;
```
CDanU: Hmm why is this static in the first place? If there is some deeper meaning behind this, this should not be in the if branch.

src/align.cpp
```diff
@@ -2072,7 +2067,7 @@ static void align_oc_msg_colon(chunk_t *so)
    {
       tmp = nas.m_aligned.GetChunk(idx);
 
-      tlen = tmp->str.size();
+      int tlen = tmp->str.size();
```
jibsen: Looks like the type is perhaps `deque<int>::size_type` (from unc_text.h), which is likely, but not guaranteed, `size_t`.

src/options.cpp
stefan-nu: Sorry my fault. I am still stuggling with git / github /tortoisegit sometimes.

src/align.cpp
```diff
@@ -1638,12 +1636,10 @@ static chunk_t *scan_ib_line(chunk_t *start, bool first_pass)
...
 static void align_log_al(log_sev_t sev, size_t line)
 {
-   size_t idx;
-
    if (log_sev_on(sev))
    {
-      log_fmt(sev, "%s: line %lu, %lu)",
-              __func__, line, cpd.al_cnt);
+      log_fmt(sev, "%s: line %lu, %lu)", __func__, line, cpd.al_cnt);
+      size_t idx;
```
stefan-nu: Absolutely. I did't do it as I tried to stick to the original code as much as possible. But me too I prefer placing the definition in the for loop. 

I will walk through the code and where it seems appropriate to me, move the loop counter into the loop.

src/align.cpp
stefan-nu: Yes I think there are still places where a variable/parameter could be made const or the type might be better changed. But I didn't want to do that in this change request to avoid too many changes. 
This CR is growing rather larger anyway.

src/output.cpp
stefan-nu: thanks

src/parse_frame.cpp
stefan-nu: I agree the static variable goes better at the start of the function.

stefan-nu:
I am kind of puzzled that Run_uncrustify_for_sources.sh fails on travis but Run_uncrustify_for_sources.bat succeeds on my windows build system. Is the forUncrustifySources.cfg used for travis the same as in the source? Any other ideas?

src/newlines.cpp
```diff
@@ -397,6 +396,7 @@ chunk_t *newline_add_between(chunk_t *start, chunk_t *end)
    }
 
    /* Scan for a line break */
+   chunk_t *pc;
```
CDanU: The scopes could be limited a bit more if this definition is defined in the for loop and inside:
```
if (end->type == CT_BRACE_OPEN)
{
   pc = chunk_get_next(end);
```

src/align.cpp
CDanU: @jibsen Yes, I had the same discussion with @gmaurel about deque<int>::size_type vs size_t, in the end we decided to use size_t. I hope at some point we `auto` everything anyway.

src/newlines.cpp
stefan-nu: You are right. In this case it is actually more appropriate to use a new variable. Both variables have the same name but actually they are two different things so it is more sensible to use separate variables.
There are a few more locations where this could be done as well.

stefan-nu:
I think the PR is ready to be merged. But my last commit shows a conflict. I could not figure out why and how to resolve it. Can you help me out?

CDanU:
Looking at your repo I em impressed that github did not complain earlier and removed the commits that where already merged into the master from this PR.

Here is what you do:
* use `git stash` to stash away anything that you did not have commited yet.
* edit your .git/config file, it should have somewhere the lines:
```
[remote "origin"]
	url = https://github.com/stefan-nu/uncrustify.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
```
add:
```
[remote "unc"]
	url = https://github.com/uncrustify/uncrustify.git
	fetch = +refs/heads/*:refs/remotes/origin/*
```
and change:
```
[branch "master"]
	remote = unc
	merge = refs/heads/master
```
With this changes your master branch will not pull change from your repo, but from the uncrustify repo.
* change to you master branch: `git checkout master`
The changes that you have made in the master branch where already merged into the master so they can be removed: `git reset --hard b4a791e82cff005a554dda7bcbe12d8bacc8df2d `
This basically reverts to the last stage that your master branch had prior to diverting.
* now you can pull the new changes that the upstream master has via `git pull --rebase`
* change now into your #870 branch: `git checkout \#870`
* Here you still have the previous state of your master and the additional changes, remove the unneeded commits again. This time you have to use a different command to keep the additional changes.
Lets manipulate history for that: `git rebase -i b4a791e82cff005a554dda7bcbe12d8bacc8df2d` 
Now, based on your git config ([core] -> editor), some editor should open up if you use windows and tortoise it will most likely be vim. You could use some other editor but most need special settings only terminate the edition phase in the command line. For linux I use `kate -b`
Every line represents a commit, remove every commit which message is not starting with `#870 ....`, save and than close the editor.
* You removed the unneeded comments now but you still don't have the new changes from the master so you need to import theses and push your changes on top via: `git rebase master`
In doing so git will print out error messages for conflicts that you have to resolve. Either you do it by hand or use a tool for that. I use kdiff3 via `git mergetool -t kdiff3`. Kdiff3 will resolve most of the easy stuff for you automagically. I looked and in src/align.cpp after 

>/* find the longest args that isn't the first one */

it does not know what to do with the var `int len` that was changed in the upstream master and in your branch.
Your changes should overwrite the changes from the master branch. Manually check the formatting again. If formatting or any other changes where needed add them via `git add <filename>`. With that everything is resolved and you can continue the rebase with `git rebase --continue`.
* After the rebase finished successfully now is the time to get the `git stash`ed suff back via `git stash pop`. Since the stashed code was based on your old state of the branch here again some merge conflicts can occur than you again have to fix.
* When everything is done push the changes into your upstream branch via "git push -f". The -f is needed to forcefully overwrite your branch after you altered time. _Generally you should mess with history only in branches and not inside the master._
Done

That is in a nutshell my process of manipulating history and merge conflicts. It is most likely not the easiest/fastest/safest way. 

Don't fear that you do anything wrong while stepping through that short guide, experiment with git a bit, if you mess up read about the `git reflog`. Also I have a copy of your branch just in case. If you don't manage to fix this on your own I will make a new PR with your commits and some fixes after some days.

CDanU:
Sorry that I spoil your learning experience a bit, but I had to step in as the amount of merge conflicts kept growing. 
Please don't use the '#' char at the first position of your commit messages next time, as it comments out the commit message line. How did you actually do this? I assume you where using windows? If so this should maybe be filed as bug in the windows git version.

stefan-nu:
Thanks a lot for merging and for your extensive explanation. It will help me for the next pull request. I was trying to follow your steps but found that two tests failed after I rebased to a newer master commit. Obviously I was too slow in searching for the problem's source, so you stepped in. 

You are right I am using windows which does not complain about the # character in the commit message. I usually use it as my ticket system recognizes the issue number referring to the hashed number. But github seems to work differently.

Issue #852, opened 2016-12-07
Different results on 1st and 2nd run.


sniper-developer:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastbin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

working version: uncrustify 0.64

===ex.cpp===
```
void func
{
    if (true)
    {

        return;
    }
}
```
result after pass 1:
```
void func
{
	if (true)
	{
		return;
	}
}
```
result after step 2 (expected result)
```
void func
{
	if (true)
		return;
}
```
===config file u.cfg===
eat_blanks_after_open_brace=true
mod_full_brace_if=remove
mod_full_brace_nl=2

===command line===
uncrustify.exe -c u.cfg -q -l CPP ex.cpp & uncrustify.exe -c u.cfg -q -l CPP ex.cpp.uncrustify

CDanU:
That is a tough one. Here everything works as it should and I am not really sure if this should be solved if it is solvable in the first place.

sniper-developer:
Of course it's not the same, issue is that it becomes what it should be only after 2nd run on same file.
Different files is just to show reproduce quickly.

CDanU:
[...]

PR #875, opened 2016-12-20
uncrustify needs some more documentations.

gmaurel:
Let me know if it is the right way to help the beginners

### DIFF
--- a/documentation/htdocs/align_typedef.html
+++ b/documentation/htdocs/align_typedef.html
@@ -1,0 +1,161 @@
+<html>
+<head>
+   <meta http-equiv="content-type" content="text/html; charset=utf-8"/>
+   <title>Uncrustify: The align typedef options</title>
+</head>
+<body lang="en-US">
+<h2>align_typedef</h2>
+<h3>align_typedef_span</h3>
+The span for aligning single-line typedefs (0=don't align)
+<table border="border">
+<tr>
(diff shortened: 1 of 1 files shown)


CDanU:
The current style is a bit plain, maybe [sphinx](http://www.sphinx-doc.org/en/1.5.1/) could be used for this? 
A bad example for this is CMake, a fairly good on is [OpenCV](http://docs.opencv.org/2.4/doc/tutorials/imgproc/imgtrans/hough_lines/hough_lines.html)

In the webtool that I made (which development currently is on hold) I wanted to create for every option an example. The real complexity lies within combining options and document such issues as #852.

I see that you have examples in there for combined options, maybe we should collect [public uncrustify configs](https://encrypted.google.com/search?hl=en&q=site%3Agithub.com%20uncrustify.cfg) and see what people are using/not using to get a feeling what should maybe be documented?

mihaipopescu:
Why HTML when Markdown is so much simpler ?

gmaurel:
@CDanU OK, I wait for your tool.

CDanU:
Which one of the tools do you mean? Anyway don't wait up on me or my tools.

As @mihaipopescu mentioned markdown would be easier to start with and Github does have a wiki system with markdown support. We can set up some HTML docs or a website later if markdown is too restrictive.

PR #877, opened 2016-12-21
fix compiler warnings

nivekkagicom:
With the change to use size_t the format specification (%d) may produce warnings and should be updated to specify a length modifier. I have used (%zd) per C99 standard. Also *printf functions without a string literal format string are considered a "potentially insecure" as the compiler warns. The remedy is to use "%s" as the format string and pass the string as the argument.

### DIFF
--- a/src/chunk_list.cpp
+++ b/src/chunk_list.cpp
@@ -507,7 +507,7 @@ chunk_t *chunk_get_prev_type(chunk_t *cur, c_token_t type,
       pc = chunk_get_prev(pc, nav);
       if (pc != NULL)
       {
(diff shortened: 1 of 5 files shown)


gmaurel:
I am making lot of changes from ```int``` to ```size_t```.
I use ```%lu``` for the unsigned types.
What is better, what are the differences? ```%lu``` or ```%zu```

nivekkagicom:
AFAIK long can be 32-bit on some systems (Windows) while long long and size_t can be 64-bit on the same system. When that is the case one has to use %llu (or %lld); however, %zu (or %zd) will always work for size_t regardless. Same reasoning goes for using %td for ptrdiff_t instead of %ld or %lld.

Issue #776, opened 2016-10-02
make_version.py issues


jibsen:
As mentioned in #769, `make_version.py` does not appear to work from its new location in the scripts folder. The issues I noticed were:
- print needs parenthesis for Python 3 compatibility
- `scripts/tokenize.py` shadows a system module that `make_version.py` includes
- it insists on looking for `.git` in the same folder as the script rather than it's working folder

mihaipopescu:
I don't think this is fixed. We should reopen this one and take a look.

mihaipopescu:
@jibsen Thanks for checking this. I have few questions.

> print needs parenthesis for Python 3 compatibility

I don't think it does. I have Python 2.7 and it works. Can you elaborate why do you think otherwise?


> scripts/tokenize.py shadows a system module that make_version.py includes

I don't understand. I don't see any connection between the two scripts.

@gmaurel Can you reopen this issue ? Also pretty please don't close issues that don't have any resolution (like this one) as they can get lost. If you really want to do it you can add a comment with a reason why.

jibsen:
In Python 3, [print is a function](https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function), so it requires parenthesis. In order for the script to work with both Python 2 and 3, we need to add parenthesis.

For better or worse, Python files in the local folder may [shadow Python modules](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-name-shadowing-trap). `make_version.py` imports subprocess, which somewhere down the line inside the Python 3 libraries imports tokenize. Python 2 does not import this, but you can get the same effect if for instance you create a file `threading.py` in the scripts folder containing only the word raise. Basically, since this isn't a package with its own namespace, I think we'll need to steer clear of any filenames that might shadow Python libraries.

PR #892, opened 2016-12-31
cleans up python scripts and makes them also run with python3

CDanU:
ref: #776


mihaipopescu:
Nice work. Is it still compatible with Python 2.7 ?

CDanU:
Yes, I just made them compatible with 2.7 and 3.5 but have not checked if the scripts are still used or if they had previously generated right results to begin with.

I made the changes mostly because I wanted to use `make_version.py` in a [cmake build](https://github.com/CDanU/uncrustify/commit/39fcc4bb52c4f5022b3580b8ac1455e332f6761d) for the emscripten interface.

mihaipopescu:
Cool. Maybe we can ensure that with a CI job (either travis or appveyor) controlling it from CMakeLists.txt like this `find_package( PythonInterp 2.7 REQUIRED )`.  You could also have a job for the emscripten that requires 3.5.

Issue #868, opened 2016-12-16
Header files code layout


stefan-nu:
I would like to propose to restructure the header files. But before I do that, I would ask you to comment on my propositions to know if you like or dislike them.

1. Create an individual header file for ever C file with the same name but ending *.h Thus splitting up prototypes.h into several header files
2. Create a subdirectory "inc" and move all headers to there. Keeping the *.cpp files in the subdirectory "src"
3. Move things that are used only in one C file from the header to the C file to keep it "private"
4. Move things that are used in more than one file to the header to make it "public"
5. All headers use a standard include guard

I think point 3,4,5 are already satisfied. But I wanted to point out the basic idea.

I think this would make the code structure more logical. Maybe there are good reasons why this have not been done so far. If you don't agree with my proposition I would love to learn what is the advantage of the current header layout.

gmaurel:
Proposal 2 is OK for me, but using ```include```.
Proposal 1:
I don't understand the first part!
I don't like ```prototypes.h``` at all.

CDanU:
He wants to split the content of _prototypes.h_ into several _.h_ files that are included by their _.cpp_ files. This would be nice, I am not entirely sure tough on point 4 if there are such cases in the code.

jibsen:
I think an `include` folder is usually used to separate the public header files from the source code of libraries, so people know which headers are needed to link with the library.

PR #897, opened 2017-01-06
split prototypes.h into individual header files

stefan-nu:
This commit 
* splits the header prototypes.h in several files that correspond to their respective *.cpp source file
* moves function comments to the header
* adds a new directory named "include"

I did not move the header files into the include directory yet, as this might break the build scripts.
Could somebody please add the include directory to the compiler's/linker's include list. 
The header files can then be moved without trouble.

Implements #868


stefan-nu:
There seems to be a problem with the uncrustify format check. It adds two newlines in file detect.cpp line 479. Is this a bug in uncrustify?

mihaipopescu:
Does the test passes locally ?

stefan-nu:
No I see the same behaviour locally. Maybe I've introduce a bug, even though I did only move code.

The following lines 
'''C
// unc_add_option("sp_cond_question", UO_sp_cond_question, AT_IARF,
//                "Add or remove space around the '?' in 'b ? t : f'");
'''

get changed to
'''C
// unc_add_option("sp_cond_question", UO_sp_cond_question, AT_IARF,


//                "Add or remove space around the '?' in 'b ? t : f'");
'''

CDanU:
I checked out your branch and everything works here if I uncrustify `detect.cpp` to remove the mentioned newlines. Running uncrustify on `options.cpp` changes nothing. Did you uncrustify with a fresh build?

CDanU:
I checked your branch, the mentioned newlines in `detect.cpp` do occur also before your commit. I can not reproduce the additional newlines in `options.cpp`

stefan-nu:
Thanks for confirming. I do not see an additional newline in options.cpp though.

gmaurel:
Have a look to the travis log.
while running ```./scripts/Run_uncrustify_for_sources.sh ``` you get:
```
Problem with detect.cpp
use: diff ./src/detect.cpp ./results/detect.cpp to find why
```
This means that the file is not conform to the actual configuration file.
Please run the script to see locally.

CDanU:
The problem is that when the change above is introduced uncrustify, while formatting, adds two new newlines that should not be there.

gmaurel:
I just try today 8da4bdcdd4e2801e8821f302e6e3fb547fef5014
I cannot reproduce.

stefan-nu:
I have moved the function detect_option before the multiline comment.
This prevents uncrustify to add the newlines. Thus the actual change request passes the tests and if you find it useful it may be merged.

As far as I understand, there exists a bug in uncrustify that adds the unexpected newlines under some specific condition. Thus a test and fix for it is required. But this shall better be done in separate commit.

CDanU:
On 8da4bdcdd4e2801e8821f302e6e3fb547fef5014 I still see the same error. I will open a new issue for this.

---------------------------------------

@stefan-nu: It would be nice if you could try to squash all commits into one. Usually you would not do this, but in this case it is OK as everything that you have done is essentially having only one objective (excluding the fixup commits). 

As there are merge commits inside your feature branch you can not use `git rebase --interactive` and set the needed commits to squash but would have to do it like this:
* Reset the HEAD of your branch to the point where you branched off. You do this with `git reset 94ac85169703dc0c1f82a36cd4c245806dbe8010`. It will remove all commits up to 94ac85169703dc0c1f82a36cd4c245806dbe8010 but all the changes that you have made will still persist in your code, they will be unstaged. 
* Stage all the changes: `git add src/ tests/ scripts/`
* Make a new commit. You can copy all the metadata and commit text from another commit via 'git commit -C b3cdf93caf5853ba73426e222982a8f4ae076e8f'. Now all the changes that you have done are in one commit.
* As you had merged the master into your branch you need to filter out those imported changes. You can make this easily if you checkout to your master branch, pull in the new changes from the uncrustify repo master (see #871 where I described how to pull from the uncrustify repo via changing the .git/config) with `git pull --rebase`, checkout into your feature branch again and apply your changes on top of the master  with `git rebase master` 
* As you have effectively changed the commit structure/order you cannot just easily push the changes into your remote feature branch but have to overwrite it via: `git push -f `

stefan-nu:
@CDanU I tried what you asked for. But I am not sure if it worked out correctly.

CDanU:
@stefan-nu it does not seem like you pushed the changes onto github.

gmaurel:
Let us synchronize.
Begin is the comment: https://github.com/uncrustify/uncrustify/pull/897#issuecomment-270901827
@stefan-nu is writting: There seems to be a problem with the uncrustify format check.
This must be solved as first.
Please have a look before you introduce any change.
The actual git shows a SHA1ID: 8da4bdcdd4e2801e8821f302e6e3fb547fef5014.
With this the test ./scripts/Run_uncrustify_for_sources.sh don't change any sources.
Please test that and report. Thanks.

gmaurel:
Correct.
It isn't a bug but a problem.
Some (most) of function have some comment before the first lines of code to
comment the function (call, use, ...).
It is NOT possible to distinguish between that and the case you are showing.
As uncrustify finds a function, it look for comment before and add the NL as specified
at ```nl_before_func_body_def```.

The only solution is to introduce a (little) comment before the function. Sorry.

gmaurel:
Sorry for the bad answer.
Thanks for the PR.

gmaurel:
@stefan-nu 
Can you please clean the PR?
The "moved detect_options()" is not more necessary as the PR 904 is merged.

stefan-nu:
@gmaurel I cleaned up the pull request and rebased it to the latest master.

src/compat_win32.cpp
```diff
@@ -7,7 +7,7 @@
  */
 #if defined (_WIN32) && !defined (__CYGWIN__)
 
-#include "windows_compat.h"
+#include "compat.h"
```
CDanU: What is the reason for this change? 
Note:
* In your previous commits you had a `compat_win32.h` file which apparently wasn't used.
* Your new `src/uncrustify.h` file is including `windows_compat.h`

stefan-nu:
CDanU, You have a good eye. I appreaciate a lot.

Regarding the comment before `detec_options()`. I left it as I did not see the fix for #902 in the master where I branched off yet. The comment shall be removed as soon as possible.

The file `compat_win32.h` originally held the function prototypes `unc_getv()` and `unc_homedir()`. As those functions are used in `compat_win32.cpp` and `compat_posix.cpp` I felt it is more reasonable to move the prototype to a common header named `compat.h`. 
As far as I found the inclusion of `windows_compat.h` in `compat_win32.cpp` was not necessary. So I removed it.
Did I miss something and the inclusion of `windows_compat.h` is required?

My objective is eventually to get rid of `prototypes.h` altogether. But that needs some more work which I prefer to do after the current state was merged into the master. Otherwise I expect too much troble with merge conflicts.

CDanU:
I see you include `compat.h` in `compat_posix.cpp` as there is no `posix_compat.h`. 

The `windows_compat.h` file has a bunch of defines and includes so you should not just remove the 
include in `compat_win32.cpp`

For clarity it would be nice to have that missing `posix_compat.h`, even if it itself only includes the `compat.h` file, which is than included by `compat_posix.cpp`. The files  `posix_compat.h`, `compat_posix.cpp`, `windows_compat.h` and `compat_win32.cpp` should al have platform dependant if guards.

--------------

Could you remove all the changes that where connected to compat.h from the current commit and than move them with the suggested changes into a new one?

stefan-nu:
I have rebased the PR to the latest available master. 

Ok I tried what you asked for. 

Regarding the topic of the compat_xxx I have changed my modification back. I left the compat.h which will serve for later use. I am not sure if I fully understand your idea of the compat restructuring yet. In any case I agree that it should be changed a bit. But lets do that in a different PR.

The comment above `detect_options()` is still needed. It seems that the fix of #902 did not make its way into the master branch yet.

stefan-nu:
I have rebased the PR to the latest master. Is there still something you want to have changed? 
In my opinion the PR is fine and can be merged.

CDanU:
Sorry for the delay, I trashed up my system. I am goning to look at this today or tomorrow.

stefan-nu:
Don't worry. I'll keep practicing the rebase operation ;-)

src/uncrustify.h
```diff
@@ -0,0 +1,132 @@
...
+
+#endif /* UNCRUSTIFY_H_INCLUDED */
+/**
+ * @file uncrustify.h
+ * prototypes for uncrustify.c
+ *
+ * @author  Ben Gardner
+ * @license GPL v2+
+ */
+#ifndef UNCRUSTIFY_H_INCLUDED
```
CDanU: It seems like you have copied content two times into this file.

src/uncrustify.h
```diff
@@ -0,0 +1,132 @@
...
+#define UNCRUSTIFY_H_INCLUDED
+
+#include <stdio.h>
+#include "token_enum.h"
+#include "log_levels.h"
+#include "windows_compat.h"
```
CDanU: The windows_compat.h  before was always protected by ifguards
> #if defined (_WIN32) && !defined (__CYGWIN__)

src/sorting.cpp
```diff
@@ -1,13 +1,12 @@
 /**
- * @file braces.cpp
- * Adds or removes braces.
```
CDanU: This would need a new description. Something like: "Sorts chunks and imports" maybe?

src/sorting.h
```diff
@@ -0,0 +1,16 @@
...
+ * @author  Ben Gardner
+ * @license GPL v2+
+ */
+#ifndef SORTING_H_INCLUDED
+#define SORTING_H_INCLUDED
+
+#include "uncrustify_types.h"
+
+
+void sort_imports(void);
```
CDanU: Here some functions are missing ... or did we decide not to include some? 
Why is 'uncrustify_types.h' included here?

src/uncrustify.h
stefan-nu: Yes this was a merge error during rebase. Fixed it.

src/uncrustify.h
stefan-nu: The `windows_compat.h` has been replaced by `base_types.h`

src/sorting.cpp
stefan-nu: Yes done it

src/sorting.h
stefan-nu: It is included as some types from `uncrustify_types.h`are used in `sorting.cpp`. However in `sorting.h` `uncrustify_types.h`was included already. So it was not necessary to include it again.
Thus I've remove the inclusion in `sorting.cpp`.

Regarding missing functions:  I' not sure which functions you are missing. Prototypes for static functions shall not go to the header but into the source file. Some static declarations are missing. I address that in change request #927. I have not seen missing prototypes for public functions.

CDanU: > static functions shall not go to the header but into the source file

Ok, I missed the static.

> Some people whould rather keep the inclusion in the source file and remove it from the header where it is currently not used

Yes I personally would prefer that.

I am done reviewing as everything looks OK now.

gmaurel:
@dirk-thomas Hallo!
We have some problems with the AppVeyor. Could you have a look and report why?
Begin is Test 389
https://ci.appveyor.com/project/uncrustify/uncrustify/history
Thanks

dirk-thomas:
@gmaurel I can only see that the builds are queued but for some reason not being run. The same for the last seven commits on the master branch. But since I don't have permissions for `uncrustify` on AppVeyor I don't see any more information. Someone with access should look into the AppVeyor project to check if there are any more information available or if the builds can be cancelled / restarted. AppVeyor itself seems to be operational (https://appveyor.statuspage.io/).

CDanU:
It seems like they have this issues now an then ( ctr+f "stuck" in here: [appveyor.com/discussions/problems](http://help.appveyor.com/discussions/problems)). They are fixed on their side.

gmaurel:
Fine work!

PR #900, opened 2017-01-07
These changes transform (if possible) the options from the type AT_NU…

gmaurel:
…M to AT_UNUM

to be closer to the value range.
This to prevent some compiler warning(s).

### DIFF
--- a/src/prototypes.h
+++ b/src/prototypes.h
@@ -279,9 +279,9 @@ bool unc_homedir(std::string &home);
  * @return the next tabstop column
  */
 static_inline
-size_t calc_next_tab_column(int col, int tabsize)
+size_t calc_next_tab_column(size_t col, size_t tabsize)
 {
-   if (col <= 0)
+   if (col == 0)
    {
       col = 1;
    }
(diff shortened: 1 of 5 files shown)


src/prototypes.h
```diff
@@ -279,9 +279,9 @@ bool unc_homedir(std::string &home);
  * @return the next tabstop column
  */
 static_inline
-size_t calc_next_tab_column(int col, int tabsize)
+size_t calc_next_tab_column(size_t col, size_t tabsize)
```
CDanU: This function is a bit dangerous.

What should happen when:
* `tabsize` is 0? Right now it just explodes.
* (`cpd.frag_cols - 1`) is bigger than `col`? Here it will underflow the size_t.

gmaurel: Thanks for the questions.
1.
The function ```calc_next_tab_column``` is always use with ```cpd.settings[UO_input_tab_size].u```,
which value is always >=1 and <=32:
```
   unc_add_option("input_tab_size", UO_input_tab_size, AT_UNUM,
                  "The original size of tabs in the input. Default=8", "", 1, 32);
```
2.
This is used if ```--frag       : Code fragment, assume the first line is indented correctly.```
At this time, ```cpd.frag_cols``` gets a value > 0.
All the next column value(s) are OK. This code is older as I. Never got a problem.

PR #901, opened 2017-01-07
Emscripten interface cmake build

CDanU:
Adds a CMakeLists and needed python scripts to build the emscripten interface with cmake.

The buildsteps are similar to the normal build process:
```
mkdir -p ./build2 \
&& cd build2 \
&& cmake ../emscripten \
&& make \
&& make emscripten_test
```


emscripten/CMakeLists.txt
```diff
@@ -0,0 +1,174 @@
...
+
+find_package(PythonInterp REQUIRED)
+execute_process(
+  COMMAND ${PYTHON_EXECUTABLE} "${unc_projdir}/scripts/make_version.py"
+  WORKING_DIRECTORY ${unc_projdir}
+  RESULT_VARIABLE make_version_error
+  OUTPUT_VARIABLE make_version_output
+)
+
+if(${make_version_error})
```
mihaipopescu: Can't we reuse some of the code from the main `CMakeFiles.txt` instead of duplicating code here?

emscripten/README.md
```diff
@@ -1,9 +1,11 @@
...
 --------------------------------------------------------------------------------
-1. compile Uncrustify normally
-2. copy the generated headers into the src directory 
-   (uncrustify_version.h, config.h, token_names.h)
-3. run the build.sh in the emscripten directory to generate libUncrustify.js
+ 1. create a 'build' directory somewhere
+ 2. inside this directory call 'cmake <CMakeLists.txt directory>' and 'make'
+      the CMakeLists.txt file is located in <uncrustify_root_dir>/emscripten/ 
+      dependencies: cmake, python, clang, emscripten, node
+[3.] to test the generated libUncrustify.js call 'make emscripten_test'
```
CDanU: I wanted to highlight that this is optional. I guess a better solution would be to mention this as an extra sentence without numbering and brackets.

emscripten/CMakeLists.txt
CDanU: Well, I think the best solution would be to make two extra targets inside the root CMakeLists for this (something like _make emscripten_ and _make emscripten_test_). As far es I know (almost nothing) in cmake this is done via the `add_custom_target` command which needs to call another cmake file. With that all the other code would be inaccessible. 

As the emscripten build is essentially like cross compiling it would generate files that have the same names but are completely different. I am not sure if this would interfere with the normal build.

Essentially I took the easy/lazy route here and copied everything from the original cmake that is generating source files (token_names.h, uncrustify_version.h and config.h). This is much better solution than the old build process.

mihaipopescu: > best solution would be to make two extra targets inside the root CMakeLists 

👍

Issue #902, opened 2017-01-08
nl_before_func_body_def and cpp style comments newlines bug


CDanU:
As mentioned in #897 Uncrustify adds unnecessary newlines with this config: 

```
nl_before_func_body_def         = 2
```
example0, expected output0 :
```cpp
// unc_add_option("sp_cond_colon", UO_sp_cond_colon, AT_IARF,
//                "Add or remove space around the ':' in 'b ? t : f'");
// unc_add_option("sp_cond_question", UO_sp_cond_question, AT_IARF,
//                "Add or remove space around the '?' in 'b ? t : f'");


void detect_options(void)
{
	detect_space_options();
}


```
generated code0:
```cpp
// unc_add_option("sp_cond_colon", UO_sp_cond_colon, AT_IARF,
//                "Add or remove space around the ':' in 'b ? t : f'");
// unc_add_option("sp_cond_question", UO_sp_cond_question, AT_IARF,

//                "Add or remove space around the '?' in 'b ? t : f'");


void detect_options(void)
{
	detect_space_options();
}
```
[debug](https://gist.github.com/CDanU/54e2095b7c893742d8103eb752d4811f)

-------------------------

Everything works as expected when before the detect_options functions a c type comment is placed.

example1, expected output1, generated code:1 :
```cpp
// unc_add_option("sp_cond_colon", UO_sp_cond_colon, AT_IARF,
//                "Add or remove space around the ':' in 'b ? t : f'");
// unc_add_option("sp_cond_question", UO_sp_cond_question, AT_IARF,
//                "Add or remove space around the '?' in 'b ? t : f'");

/**/
void detect_options(void)
{
	detect_space_options();
}
```
[debug](https://gist.github.com/CDanU/e9044ac6569e48578edc686dfc6dae2d)

-------------------------

Tested on: 8da4bdcdd4e2801e8821f302e6e3fb547fef5014
This behaviour started with commit 3ea4b6d which introduced the `nl_before_func_body_def` option.

gmaurel:
Correct.
It isn't a bug but a problem.
Some (most) of function have some comment before the first lines of code to
comment the function (call, use, ...).
It is NOT possible to distinguish between that and the case you are showing.
As uncrustify finds a function, it look for comment before and add the NL as specified
at ```nl_before_func_body_def```.

The only solution is to introduce a (little) comment before the function. Sorry.

stefan-nu:
[...]

PR #904, opened 2017-01-09
nl_before_func_body_def checks now comment and func line delta (i_902)

CDanU:
The previous behaviour was to insert newlines between two comments tokens.
Now newlines are inserted after a comment if it is not directly behind a func_body_def or does not belong to the set of cpp comments that is directly behind a func_body_def.

ref: #902


CDanU:
This fixes the bug but it introduces a new instability when `nl_before_func_body_def=1`(n-1 newlines) is set. In this case comments that don't belong to a comment block are moved together and therefore form one. In a consecutive run newlines are applied behind this new block.

```cpp
void f(){
	return 0;
}
/*0*/

/*e*/
void e(){
	return 0;
}
```

------------

```cpp
void f(){
	return 0;
}
/*0*/
/*e*/
void e(){
	return 0;
}
```

------------

```cpp
void f(){
	return 0;
}

/*0*/
/*e*/
void e(){
	return 0;
}
```

PR #909, opened 2017-01-11
use more size_t declarations where possible.

gmaurel:


src/braces.cpp
```diff
@@ -193,14 +193,14 @@ static bool can_remove_braces(chunk_t *bopen)
...
    bool    hit_semi   = false;
    bool    was_fcn    = false;
-   int     nl_max     = cpd.settings[UO_mod_full_brace_nl].n;
-   int     nl_count   = 0;
-   int     if_count   = 0;
-   int     br_count   = 0;
+   size_t  nl_max     = cpd.settings[UO_mod_full_brace_nl].u;
+   size_t  nl_count   = 0;
+   size_t  if_count   = 0;
+   size_t  br_count   = 0;
```
CDanU: In this and in the `examine_brace(chunk_t *bopen)` function `br_count` can underflow because of the `br_count--`. 
What happened before the var type change when it became negative?

gmaurel: Thanks for the message!

PR #913, opened 2017-01-13
Less warnings for combine

gmaurel:


stefan-nu:
I love your change. Avoiding the magic numbers makes the code much more readable.

I also appreaciate a lot your effort to reduce the number of warnings. Those warnings disturb a lot during debugging. I hope we can achieve zero warnings in the near future.

gmaurel:
I want zero warning.
At this point it will be possible to refuse PR with warning.

Issue #916, opened 2017-01-13
Segmentation fault on single line


syrinxsean:
Uncrustify core dumps with this string passed as its input:

```for(i=0;i<10;i++)```

Example:

```
% echo "for(i=0;i<10;i++)" | uncrustify -l C -c ~/.uncrustify.cfg
Parsing: 18 bytes (18 chars) from stdin as language C
       28662 segmentation fault (core dumped)  uncrustify -c ~/.uncrustify.cfg
```

Surprisingly, using a different (or even empty) configuration file does not change the crash.

This is with uncrustify version 0.64_d, though I also see the crash with version 0.64:

```
% uncrustify --version
uncrustify 0.64_d
```

**Comment by syrinxsean** [comment]
Here's a stack trace of the crash location:

```(gdb) where
#0  0x000000000049c6ac in tokenize_cleanup() ()
#1  0x00000000004a0638 in uncrustify_start(std::deque<int, std::allocator<int> > const&) ()
#2  0x00000000004a07fd in uncrustify_file(file_mem const&, _IO_FILE*, char const*, bool) ()
#3  0x000000000049e512 in main ()
```

jibsen:
It appears to be crashing due to dereferencing `pc->prev` which is NULL at tokenize_cleanup.cpp:823.

Adding a check there reveals another NULL pointer dereference of `pc->next` at brace_cleanup.cpp:860.

syrinxsean:
Confirmed. Thanks for the fix!

PR #918, opened 2017-01-14
Avoid dereferencing NULL pointers

jibsen:
This is the fix mentioned in #916.

### DIFF
--- a/src/brace_cleanup.cpp
+++ b/src/brace_cleanup.cpp
@@ -857,7 +857,7 @@ static bool handle_complex_close(struct parse_frame *frm, chunk_t *pc)
 
    if (frm->pse[frm->pse_tos].stage == BS_PAREN1)
    {
-      if (pc->next->type == CT_WHEN)
+      if ((pc->next != NULL) && (pc->next->type == CT_WHEN))
       {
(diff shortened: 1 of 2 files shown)


mihaipopescu:
A null pointer check is more often than not a change to refrain from doing while fixing crashes. It is, as you also mentioned, a way of postponing or moving the crash to another place. A null check is the easy fix to do but that's local patching and not fixing the real issue.

In this case, if we take a step back and see where it's the real issue we could see that:

1. It appears that the crash first happens in C#-only option and we could be dealing with a non C# code
2. The issue in #916 we're dealing with partial for loop. Uncrustify does a pretty good job formatting uncompilable code most of the times but when it doesn't it crashes. We need Uncrustify to detect these cases (like the one with unmatched close brace) and complain about it (fail formatting) and modify the testing framework to support regression testing of failures as well.

jibsen:
I agree, this just fixes the symptom.

Would you prefer I close this and you can open an issue for the general problem?

gmaurel:
Both aspects are important.
As uncrustify could be call wilt a fragment ```--frag``` I find the PR is fine.

mihaipopescu:
No, I think this PR is fine but it could use a test as well.

PR #922, opened 2017-01-14
Less warnings for combine, second run.

gmaurel:
The last one, I hope.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -1399,7 +1399,7 @@ void fix_symbols(void)
       }
       else
       {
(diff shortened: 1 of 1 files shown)


mihaipopescu:
> The last one, I hope.

I don't understand why do you do separate PR for each file? It's the same type of change which now is distributed across many PRs. 

It should've been one PR to remove warnings, whitelist those you don't want to deal with them now, turn treat warnings as errors.

gmaurel:
I separate the things to not be to big. I prefer little ones.

PR #925, opened 2017-01-15
a few code smells fixed or marked for review

stefan-nu:
I have marked two instructions (see \todo) which seemed a bit strange to me. Could someone please review them.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -1543,7 +1543,6 @@ static bool mark_function_type(chunk_t *pc)
    size_t    star_count = 0;
    size_t    word_count = 0;
    chunk_t   *ptrcnk    = NULL;
-   chunk_t   *varcnk    = NULL;
    chunk_t   *tmp;
    chunk_t   *apo;
    chunk_t   *apc;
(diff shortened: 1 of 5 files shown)


src/combine.cpp
```diff
@@ -3243,7 +3242,7 @@ static bool can_be_full_param(chunk_t *start, chunk_t *end)
          }
          LOG_FMT(LFPARAM, " <skip fcn type>");
          tmp1 = chunk_get_next_ncnl(tmp3, CNAV_PREPROC);
-         tmp2 = chunk_get_next_ncnl(tmp1, CNAV_PREPROC);
+         tmp2 = chunk_get_next_ncnl(tmp1, CNAV_PREPROC); /* \todo where is tmp2 used? */
```
mihaipopescu: if not used, then remove

src/output.cpp
```diff
@@ -1533,7 +1533,7 @@ static bool kw_fcn_class(chunk_t *cmt, unc_text &out_txt)
    if (tmp)
    {
       out_txt.append(tmp->str);
-      if (cpd.lang_flags & cpd.lang_flags)
+      if (cpd.lang_flags & cpd.lang_flags) /* \todo strange condition ? */
```
mihaipopescu: Maybe languages where `CT_DC_MEMBER` is defined for, i.e. `LANG_C | LANG_CPP | LANG_CS | LANG_D | LANG_VALA`

Issue #921, opened 2017-01-14
Placement of static function comments


stefan-nu:
I would like to ask your opinion about where a function comment shall be located.

In my opinion a function comment goes always right before a function declaration.
The function definition gets no comment.

When implementing #897 I have moved several function comments from the function definition to the function declaration in the header. As far as I understood that was agreed to by the team members.

Now I would also like to propose to move function comments of static functions to their declarations at the beginning of the respective *.cpp files. 

To give an example referring to file brace_cleanup.cpp

Current state:

```
/* function declarations */
...
static void parse_cleanup(struct parse_frame *frm, chunk_t *pc);
...
/* function definitions */
...
/**
 * At the heart of this algorithm are two stacks.
... long comment removed ...
 * When a #define is exited, the frame is popped.
 */
static void parse_cleanup(struct parse_frame *frm, chunk_t *pc)
```

Shall be changed to 
```
/* function declarations */
...
/**
 * At the heart of this algorithm are two stacks.
... long comment removed ...
 * When a #define is exited, the frame is popped.
 */
static void parse_cleanup(struct parse_frame *frm, chunk_t *pc);
...
/* function definitions */
...
static void parse_cleanup(struct parse_frame *frm, chunk_t *pc)
```

The reason for this change is that imho the code and the documentation become both easier to read.

Please let me know if you agree or if you prefer another approach.

mihaipopescu:
Your efforts in trying to find a good place for the comments is well received. I think that we should keep in mind that when documenting a function (at least in C++) we can differentiate the function interface/contract to it's implementation. The details of the implementation are not relevant for the function's clients. Also, let's try to avoid documenting what is obvious.

stefan-nu:
Yes I agree. In my opinion too the function comment is intendet for the caller of the function. Thus it shall only describe how to use a function. 
Any comment that describes the implementation shall go directly into the function.

stefan-nu:
Another related thing I would like to propose is to always have a function prototype. Either public i.e. in the header or private/static i.e. in the source file. The reason for this is that this will evenutually lead to a fully documented API.

stefan-nu:
Regarding your remark "let's try to avoid documenting what is obvious."
In general I agree with that. The problem is just, what is "obvious" is quite different between people.
I think the best approach is to write code that is self explanatory.

stefan-nu:
The objective of this ticket was reached by the various commits listed above. I close the ticket.

[...]

PR #927, opened 2017-01-15
adds missing declarations for static functions and moves static funct…

stefan-nu:
This modification aims to better separate the API with comments from implementation code.
Implements #921.
The functionality of the code was left unchanged.


stefan-nu:
Could someone with access to a Linux build system tell me to result of 
`use: diff ./results/31.txt.sed scripts/Output/31.txt`

The script `/scripts/Test_more_Options.sh; ` is not happy with my changes.

gmaurel:
Sorry, I have to modify the test, to prevent such a noise (next week!)
The problem are the different number width between your output and the reference.
Have a look to:
https://travis-ci.org/uncrustify/uncrustify/jobs/192133573
begins with line 4647
a difference indent_text(xxxx) vs. indent_text(xxx)
Make a copy of your output by using the ./scripts/Test_more_Options.sh to ./scripts/Output/31.txt
and push to the PR.

mihaipopescu:
> I am working under Windows where the named script does not run.
That's why we need those scripts in python or something else.

Until then, did you tried Cygwin or Ubuntu Bash on Windows (anniversary update) ?

gmaurel:
I have made some changes for the script.
Could you test again?

stefan-nu:
Yes I did so, but the test still fails...

Issue #935, opened 2017-01-18
Sorting order of usings


Leopotam:
Expected behaviour with mod_sort_using=true:

```
using System;

using System.Collections.Generic;
```

Actual behaviour with mod_sort_using=true:  

```
using System.Collections.Generic;

using System;
```

So, sorting direction option required at config for proper behaviour of sorting process.

uncrustify version: 0.64.

mihaipopescu:
No, I think it's a bug if you look at it like that.
Last time I've checked this feature it was only sorting within groups. An empty line is what splits the includes/usings in groups.

Leopotam:
Now its looks like reverse sorting. Btw, VS / R# has special option - put System.xxx usings first (like at 
[screenshot at this question](http://stackoverflow.com/questions/12831038/how-to-tell-resharper-to-sort-usings-with-system-namespaces-not-at-the-top). Is it possible to add similar (optional) behaviour?

mihaipopescu:
Can you check now ? @gmaurel have made some changes.

Leopotam:
0.64 version:
```
using Client.Common;
using LeopotamGroup.Common;
using LeopotamGroup.Serialization;
using System;
using UnityEngine.EventSystems;
using UnityEngine.Events;
using UnityEngine;
```

patched version:
```
using Client.Common;
using LeopotamGroup.Common;
using LeopotamGroup.Serialization;
using System;
using UnityEngine;
using UnityEngine.EventSystems;
using UnityEngine.Events;
```
Better, but order still wrong, check 2 last positions.

Btw, what about option for sort SystemXXX separately and put them first?
Expected:
```
using System;
using Client.Common;
using LeopotamGroup.Common;
using LeopotamGroup.Serialization;
using UnityEngine;
using UnityEngine.Events;
using UnityEngine.EventSystems;
```

gmaurel:
Is it correct to sort without comparing lower letter vs. upper letter?

Leopotam:
No, case important. Expected order:
```
using nameSpace;
using Namespace;
using NameSpace;
using NameSpacEveryday;
```

Leopotam:
[Example of proper sorting (C#)](https://gist.github.com/Leopotam/a165b09004ad4361b34ce1c404727b43)

mihaipopescu:
yes, we have a similar thing on our side:

```cs
using (var ex = x.GetEnumerator())
            using (var ey = y.GetEnumerator())
            {
[...]

Issue #857, opened 2016-12-13
Use standardised exit code values for Uncrustify


mihaipopescu:
As discussed in #842 and noted here https://github.com/uncrustify/uncrustify/blob/fef2b74b9678e34fd99eaff610da3b1c2f0bef08/src/uncrustify.cpp#L510 we need to clean our exit code as other tools may depend on them. Uncrustify is returning `EXIT_SUCCESS` and `EXIT_FAILURE` from the C/C++ standard and some other exit code that apparently match the ones from FreeBSD `sysexits.h`.

Even though there is no such thing globally accepted exit codes values, we could use the ones from `sysexits.h` and when not available just redefine them.

stefan-nu:
Me too I'd prefer some human readable error codes.

I looked through the code and found the following return codes as magic numbers when calling exit()

Magic number - function location - meaning
0 - version_exit - EXIT_SUCCESS
1 - chunk_dup - NULL pointer
2 - main - option 'nl_func_var_def_blk' too big with respect to 'nl_max'
56 - usage_exit - cannot open stdout
56 - usage_exit - Unable to load the config file
58 - usage_exit - config file missing
64 - usage_exit - Error while parsing --set
66 - usage_exit - Cannot use --replace with --prefix or --suffix
66 - usage_exit - Cannot use --replace or --no-backup with -f or -o
67 - usage_exit - Cannot use --check with output option
67 - usage_exit - Cannot specify both the single file option and a multi-file option
68 - usage_exit - Cannot specify -o with a multi-file option

In my opinion the used magic numbers do not really  correspond to the meaning of sysexits.h

When using sysexits.h this is nice for BSD. But on Linux or Windows the values may not be recognized as expected. Also sysexits.h may not be available when not building for BSD. 

I think we need a portable solution. 
1. Either using different error codes depending on the platform. Seems a bit too complicated to me.
2. Or using some known error codes like sysexits.h and providing them to other systems as well. 
3. Or using or own error codes.

What do you think?

CDanU:
I am for option 2, microsoft has defined _some_ [error codes](https://msdn.microsoft.com/library/ms681381(v=vs.85))  too.

mihaipopescu:
I'm happy with the result. Thanks @stefan-nu !

PR #937, opened 2017-01-18
Bugfix for #935

gmaurel:
If the next chunck is a CT_MEMBER then the next chunck must be token.


stefan-nu:
In your commit 36d6a1407 you add an error code (2). Maybe you could consider my pull request #857 first. It introduces standard error codes. I would prefer if you used those instead of another magic number.

PR #939, opened 2017-01-18
eliminates prototypes.h (implements #921, #868)

stefan-nu:
This PR continues the refactoring work to make the code clearer.
All C functions have now a declaration. The C++ functions need some more attention.
The function `prototypes.h` has been eliminated as all source files now have their own headers.
The source code that was remaining in prototypes.h has been move to the new file `tabulator.cpp`
Public headers have been moved to the include directory.

There was an issue with the automatically generated file `punctuators.h`. I have renamed it to `punct_table.h` as `punctuators.h` is the header for `punctuators.c`. Is this acceptable for you? Where is the name of the generated file defined?


gmaurel:
It is written "Automatically generated" but the automate is not part of the distribution.
Some manually changes where also done by TP.

CDanU:
"scripts/punc.py" is generating "punctuators.h"

jibsen:
It was not part of the previous build process, which is why it was not included in the CMake script.

There are instructions at the top of [punctuators.cpp](https://github.com/uncrustify/uncrustify/blob/master/src/punctuators.cpp) that say to run the script if that file is changed.

stefan-nu:
I have changed `scripts/punc.py` a bit. It does now include a standard include guard. In the future it shall be run like this: 
```python scripts/punc.py > src/punc_table.h```

mihaipopescu:
This is ok but I don't necessarily see the benefits of doing it. It makes a lot of sense for a library to do this separation, and only there for it's public API functions. For a standalone applications I don't think it adds a lot of value. Am I missing something ? Let's discuss.

stefan-nu:
In my opinion it is good practice to keep public headers in a separate directory. True uncrustify is no library. Still sticking to a standard source layout eases working with the code. It allows new users to concentrate on the API to get a first impression. It makes it easy to run scripts on headers only...

When I asked for moving the headers in #868 I got the impression that nobody disagreed.
But I admit that this is kind of a personal preference. I wount insist if general opinion is to keep source and headers in the same directory.

CDanU:
Don't forget about #1 :)

stefan-nu:
I have changed CMakeLists.txt to reflect the moved files. I don't like the current approach to list each *.cpp and *.h file individually. Wouldn't it be better to use a wildcard? Maybe a cmake-specialist could improve that.

jibsen:
There are some issues with globbing in CMake, for instance it is done while running CMake, so if you later add another file, simply building will not pick it up. The [CMake documentation](https://cmake.org/cmake/help/v3.7/command/file.html?highlight=glob) recommends against using globbing for selecting source files.

Also see [this discussion](https://stackoverflow.com/questions/1027247/specify-source-files-globally-with-glob) for further details.

stefan-nu:
Thanks for your explanation. Me personally I see the advantage of globbing to be easier, more newcomer friendly and to remove redundancy. This holds especially if the source files are nicely separated into directories. 
According to your link the problem with CMake not picking up new files immediately can be fixed by touching the CMakeLIst File during the build process. So I don't see that as a real problem.
But as long as there are no real problems with the current approach I am fine with it.

mihaipopescu:
Interesting to see that globbing poses such a big problem for CMake... They could've fixed it with checking checking the result of glob every time against a previous run saved in a cache file (but I think due to their architecture the cache file is already generated at that point) instead of just recommending not to do it. I think this simply adds maintenance cost.

And don't get me wrong, I'm not against moving the files... but I'm not in favor of it either.

The only thing that I see it's that it's much more convenient to have the .h as a peer to the corresponding .cpp file to make it easier to be included. Otherwise, you will have to use support from your build system to define that include folder and probably your code editor won't be aware of that.

gmaurel:
The Travis CI build is not present, I don't know why.
I would like to see it.
Could you please make a little change and push your PR again to get a new run.
Thanks.

stefan-nu:
I've rebased the commit. The Travis Check succeeded.

gmaurel:
Some more meanings?
Before I merge

stefan-nu:
I think before merging we need a decision if the headers shall be moved or not. 
From the comments I am not sure if this is really wanted.

mihaipopescu:
Don't know what to say. I mean, the cost of doing the work was already spent but I'm just worried of the additional cost this will add later because of this big refactoring.

mihaipopescu:
You already have conflicts merging this which means that it's very likely that all forks which are behind will have similar issues.

There were a lot of code refactorings done lately upstream which will make it even harder for other forks to bring their changes upstream.

And just to give you a bit of context from our side... our fork is around 6 months behind because 3 months ago I've checked and I already started seeing regressions. There were some good fixes as well but they didn't over-weighted the regressions we saw. Therefore, the effort of bringing our fork up-to-date was postponed because the gain was insignificant (effort of fixing all the bugs was greater than the bugfixes we received). That's why I think the primary focus should be on fixing issues. I'm not saying that refactorings aren't good but they really have to worth it as they have to be substantial and improve the stability, functionality, or testability of uncrustify.

My rant is not about this particular PR. It's just because lately (after 0.64 was released) the main development focus moved towards fixing warnings, formatting of the code itself before committing and similar changes.

@stefan-nu I'm sorry to bring all this on you and I hope you understand my concerns as I really believe in this project and want to contribute back to it. I don't want to see our fork diverging more than it already did...

I really like the work and the effort that you put into this project and I think you can do a lot of good things.

stefan-nu:
I understand the concerns. All I ask for is a final decision. If it is agains this PR I will close it.

CDanU:
I for one welcome this changes. They need to be done at some point anyway.

Mihaipopescu I get your concerns but you guys really created this problem on your own by not putting in PRs as fast as possible and keeping up with the upstream master branch. I noticed November that you where behind significantly. I see that the tree is messy. During this time frame not much has changed on the unity **fork**. But what are we going to do about it? The problems will pile up further but we can't just stop developing. You guys have to finally start submitting PRs.

mihaipopescu:
@CDanU After last experience, I think the entire merge process took us a couple of months to finish. From our point of view, the version we were using was quite stable and suited our needs. After that, there were other priorities on our side and that stopped us from keeping up the pace. There are still at least 30 issues that need to be addressed plus the new regressions that I've noticed.

But for this PR, renaming of files is still handled smartly by git/hg. So, go ahead @stefan-nu. I'll see what I can do. I think I'll try to advance it at least up until 0.64 and just report the regressions we've seen so far.

stefan-nu:
I've rebased the commit to fix the merge conflicts.

gmaurel:
Having a look to your 'https://github.com/Unity-Technologies/uncrustify' one can see 2 important files:
regression.test and staging.test.
"Regression test suite. These are staging tests which were fixed."
"Staging test suite. These are tests that fail and need to be fixed."
The problems are real.
We cannot ignore these bugs! We need to work on them as soon as possible!
It might be difficult to merge the bugfixes, but we only need to "copy" the ideas to solve the bugs.
We must test the second part and try to solve the problems together.

mihaipopescu:
@gmaurel exactly. We've split them like that. We need to work on the staging tests and I'll be adding some more as they appear. I also think that having a test which fail is a good way to report bugs as well.

gmaurel:
I had a look to the list regression and make a test with it.
There are only 3 Unknown symbols.
This represent only 12 lines to be modified.
If you agree I would make a PR for that the next days.

Issue #940, opened 2017-01-19
Detecting and fixing all "Dereferencing NULL pointer" errors


stefan-nu:
During the last weeks, were reportet three bugs caused by derefererencing a Null pointer (#938, #916, #859).
@mihaipopescu pointed out, that most often the Null pointer dereferencing is not the root cause of the bug. This might well be the case and often the fix is more complicated than just adding a check for a NULL pointer.

Nevertheless I feel that any such crash is one too much. For a user it is disappointing to experiance such behaviour. We are lucky that some users take the effort to report the issue. I'd assume for every reported issue there are several other users that just stop using uncrustify because they think it is a buggy tool.
That said, I think it is important to prevent such crashes.
Thus in my opinion a pointer that cannot be garanteed to be valid has always to be checked. It is then up to the programmer to either print a warning, exit the program with a useful error message or slilently continue the program exection.

When I check the sourcecode with pclint I get thousands of errors/warnings and many of those refer to possible Null pointer dereferencing. My experiance is that pclint is rather precise and most of those warnings are real bugs. That said I feel we have to spend some time fixing those issues before each line gets reported by a disappointed user.

I am willing to go into this task even though I know it will be a big one and it will take quite some time to terminate.

pclint is a commercial tool that it is presumably not available to every developer. Anyway would you agree that I add a configuration for pclint to the test directory? Everyone who has access to the tool may then check his code. For all others its just a directory to ignore.

mihaipopescu:
`cppcheck` can be an alternative. Also, I think @gmaurel has already done a pass with a static analysis tool (unsure if it was coverity or something else).

Later maybe we could consider fuzz testing. And what about a crazy idea of formatting source code from github ? It could help to reveal a lot of issues.

stefan-nu:
I am using `cppcheck`. It is a nice tool. And in my opinion it should be used as an entry level tool by everyone. But by no means it is comparable to `pclint`. 

To show you the difference I attache the check output for commit 6315cb3b5d of `cppcheck` and `pclint` respectively:
[cppcheck-results.txt](https://github.com/uncrustify/uncrustify/files/716738/cppcheck-results.txt)
[pclint-results.txt](https://github.com/uncrustify/uncrustify/files/716737/pclint-results.txt)

`Coverity` is more or less comparable to `pclint`. But I don't see where it (or another tool) is used to check uncrustify so far.

[...]

PR #941, opened 2017-01-19
pclint check added (implements #940)

stefan-nu:
This PR shows how a static code check can be implemented using `pclint`. Refer to ticket #940 for a discussion of it. 
To run the check you need a license for `pclint`. The results can either be saved to a text file or be displayed on top of the source code when using `Eclipse`.
The configuration used is ment for Windows using Mingw as toolchain. To run the check with another toolchain the paths in run-pclint-eclipse.bat need to be change. To run the check under Linux another tool called `flexelint`is required.

Maybe it would be more useful to use `Coverity` as it seems to be available for Github projects. The results should be more or less equal. This PR may still serve as an example how `Coverity` could be included to the project.

Before considering this PR for merging we need to decided if this is a wanted feature or not.


gmaurel:
every effort is welcome

stefan-nu:
@gmaurel could you give me read access to Coverity so I can check it against pclint. I have send a request already.

PR #943, opened 2017-01-19
Fix include guards

jibsen:
Names starting with an underscore followed by an uppercase letter are reserved.


stefan-nu:
Thanks for your correction.

PR #944, opened 2017-01-20
adds checks for Null pointers and some more code clean up

stefan-nu:
implements #940 

This PR aims to make the code more robust against unexpected situations. Hopefully it will prevent some crashes of uncrustify.

* added some Null pointer checks
* some variable types changed
* some implicit casts with loss of precision made explicit
* added a few \todos for review

To make the code clearer
* removed unused header includes
* a few unused assignments deleted

I kindly ask you to review not only my changes but especially the instructions I have marked with \todo.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -16,7 +16,6 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
-#include <cerrno>
 #include "unc_ctype.h"
 #include "uncrustify.h"
 #include "indent.h"
@@ -503,9 +502,9 @@ void align_right_comments(void)
           (pc->type == CT_COMMENT_CPP) ||
           (pc->type == CT_COMMENT_MULTI))
       {
-         bool skip = false;
          if (pc->parent_type == CT_COMMENT_END)
          {
+            bool    skip  = false;
             chunk_t *prev = chunk_get_prev(pc);
             if (pc->orig_col < (prev->orig_col_end + cpd.settings[UO_align_right_cmt_gap].n))
             {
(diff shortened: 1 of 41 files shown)


src/align.cpp
```diff
@@ -518,7 +517,7 @@ void align_right_comments(void)
                        pc->orig_col, prev->orig_col_end, cpd.settings[UO_align_right_cmt_gap].n);
                skip = true;
             }
-            if (!skip)
+            if (skip == false)
```
gmaurel: !skip is better

stefan-nu: Why is it better? I know many people use it but in my opinion it is bad style. It requires two thoughts my version just one. And the compiler doesn't care.

mihaipopescu: I think there are some arguments against it. First of all, the first one is faster to write. 

Secondly, we all know that if expressions are inherently verifying the truth of expressions so then why you wouldn't go as far as saying `if (skip == true)` ?

Also, the naming of the bool is a bit wrong as it should represent something positive, like `can_change` for instance.
So then, instead of having something like _if skip is false then_ you'll have `if can change then`.

It's nothing wrong with the code. It'll compile but it's a bit surprising. I've not seen this very often in production code. You don't have to update the PR just for this.

stefan-nu:
Regarding "mixing different changes". You are right. All changes are the result of a static code check. I searched for issues that might indicate a bug and fixed some of them. But I should have separated the different type of changes in several commits. 

Regarding the `if(!skip)`: I agree with your remark about naming the flag. In this case I took the existing code and tweaked it a bit to make it easier to read. But it is still far from self explanatory. In general I want code that reads like a book. Well there is still some work to be done...

src/align.cpp
gmaurel: I have the same opinion.
Please, change it back. Thanks.

gmaurel:
It is too many changes for me.
As I try to do it with the last changes "LessWarningFor...", let us have smaller PR.
If we latter detect a problem, it is easy to repare.
Please, make this PR for only a part of it, or create some new PR.

stefan-nu:
Ok I will split the PR in several commits.

gmaurel:
It will be fine.

stefan-nu:
The PR has been split in several smaller ones. Therefore this PR is closed.

PR #954, opened 2017-01-21
added checks for null pointers

stefan-nu:


src/output.cpp
```diff
@@ -1832,7 +1841,7 @@ static void do_kw_subst(chunk_t *pc)
                   unc_text nl_txt;
                   nl_txt.append("\n");
                   nl_idx++;
-                  while ((nl_idx < (size_t)idx) && !unc_isalnum(pc->str[nl_idx]))
+                  while ((nl_idx < idx) && !unc_isalnum(pc->str[nl_idx]))
```
mihaipopescu: maybe my memory is failing but didn't @gmaurel did a lot of PRs switching from `int` to `size_t` and now it appears that you reverted his changes?

src/uncrustify.cpp
```diff
@@ -1288,7 +1288,10 @@ static void do_source_file(const char *filename_in,
    {
       for (deque<UINT8>::const_iterator i = cpd.bout->begin(), end = cpd.bout->end(); i != end; ++i)
       {
-         fputc(*i, pfout);
+         if (pfout != NULL)
```
mihaipopescu: this null check should be done outside of the loop.

src/combine.cpp
```diff
@@ -2334,6 +2334,12 @@ static void fix_casts(chunk_t *start)
    }
    paren_close = pc;
 
+   if (last == NULL)
+   {
+      /* \todo can we be sure that last is never NULL here? */
```
mihaipopescu: if this should never happen, why not `assert` that?

src/output.cpp
stefan-nu: This is a merge error. size_t is correct. I'll fix it.

src/uncrustify.cpp
stefan-nu: Yup

src/combine.cpp
stefan-nu: If we can be sure it is never NULL an assert is fine.  But I am not sure about it. Thus my question.

mihaipopescu: if before it was crashing otherwise, then we can safely add an assert.

stefan-nu: When I look through the function `fix_casts` it seems to me that it is well possible that last is NULL. The value of last depends on the chunk list which again depends on the input source file. We can't predict what source file a user presents to uncrustify. And we also have to expect that a source file holds invalid C code. Thus in my opinion the Null check is correct, the assert not.

stefan-nu:
@mihaipopescu 
> Null checks are usually delaying/moving the crash to another place. Maybe sometimes it's better to have a crash as it's easier to debug.

I do not really agree to that. In my opinion a crash is never acceptable. It is much better to tell the user what went wrong and to stop willingly.

stefan-nu:
This PR got broken during repeated rebasing. As it was not considered important I'll close it.

PR #956, opened 2017-01-21
added todo markers to instructions that need clarification

stefan-nu:
while walking through the code I found some instructions where I had a doubt if they are correct.

Maybe someone could have a look on those and give some clarification.

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -857,7 +857,7 @@ void add_long_closebrace_comment(void)
                {
                   nl_min = cpd.settings[UO_mod_add_long_switch_closebrace_comment].u;
                   tag_pc = sw_pc;
-                  xstr   = sw_pc ? sw_pc->str : NULL;
+                  xstr   = sw_pc ? sw_pc->str : NULL; /* \todo NULL is no unc_text structure */
                }
                else if ((br_open->parent_type == CT_FUNC_DEF) ||
                         (br_open->parent_type == CT_OC_MSG_DECL))
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -1117,12 +1117,14 @@ static bool file_content_matches(const string &filename1, const string &filename
       }
       if ((len1 <= 0) || (len2 <= 0))
       {
-         break;
+         break; /* reached end of either files */
+         /* \todo what is if one file is longer
+         * than the other, do we miss that ? */
       }
       int minlen = (len1 < len2) ? len1 : len2;
       if (memcmp(buf1, buf2, minlen) != 0)
       {
-         break;
+         break; /* found a difference */
       }
       len1 -= minlen;
       len2 -= minlen;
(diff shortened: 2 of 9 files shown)


src/braces.cpp
```diff
@@ -858,7 +858,7 @@ void add_long_closebrace_comment(void)
                {
                   nl_min = cpd.settings[UO_mod_add_long_switch_closebrace_comment].u;
                   tag_pc = sw_pc;
-                  xstr   = sw_pc ? sw_pc->str : NULL;
+                  xstr   = sw_pc ? sw_pc->str : NULL; /* \todo NULL is no unc_text structure */
```
CDanU: It is, though I am not really sure if this results in the correct behavior here. Basically the [operator=](https://github.com/uncrustify/uncrustify/blob/8004bdac1d6d9b495668daa9b502c11f99879cdd/src/unc_text.h#L82) function of `unc_text`  is called. This sets the content of `xstr` to `0`.

src/uncrustify.cpp
```diff
@@ -1416,8 +1418,8 @@ static void add_func_header(c_token_t type, file_mem &fm)
 
       if (ref->type == CT_FUNC_DEF && ref->parent_type == CT_NONE && ref->next)
       {
-         int found_brace = 0; // Set if a close brace is found before a newline
-         while (ref->type != CT_NEWLINE && (ref = ref->next))
+         int found_brace = 0;                                   // Set if a close brace is found before a newline
+         while ((ref->type != CT_NEWLINE) && (ref = ref->next)) /* \todo is the assignment of ref wanted here?, better move it to the loop */
```
CDanU: This was done on purpose, it is one of those _tricks_ that allows to save space for decreased readability as a tradeoff. 
After the type check, `ref` is assigned to `ref->next` and then, because there is a assignment inside braces, the assigned to var (`ref`) is bool evaluated. As `ref` is a pointer it is checked for **not** being `NULL`.
It would be better to clear things up a bit.

todd-richmond: the new, excess () actually makes this harder to read as the brain has to parse through () that don't actually do anything.
while (ref = ref->next) is pretty standard syntax, but combining it with another bool is much less so so moving ref = ref-> next as the 1st line of the loops is clearer

todd-richmond:
if you are trying to work through bad memory references or leaks, I highly recommend cppcheck. Free, no real setup effort and not too bad with false positives. Other scan tools will find different issues, but are usually more of a pain to both get working and wade through the results

cppcheck --enable=all --force  --inline-suppr -I incude_path  *

stefan-nu:
You are right. See #940 for some more discussion on that subject.

src/uncrustify.cpp
mihaipopescu: Yeah, I agree with @todd-richmond here. On the other hand, when the expression is not clear and it requires knowledge about C++ operator precedence then I would add parentheses.

PR #957, opened 2017-01-22
reduce redundancy of module chunk_list

stefan-nu:
the functions chunk_get_next_nvb and chunk_get_prev_nvb do almost the
same thing. They have been combined into a common function chunk_get_nvb()

### DIFF
--- a/src/chunk_list.cpp
+++ b/src/chunk_list.cpp
@@ -9,75 +9,295 @@
 #include <cstring>
 #include <cstdlib>
 
+#include "uncrustify_types.h"
 #include "ListManager.h"
 #include "prototypes.h"
 #include "uncrustify.h"
 #include "space.h"
 
 
-typedef ListManager<chunk_t> ChunkList;
+/** use this enum to define in what direction or location an
+ * operation shall be performed. */
+enum loc_t
+{
+   before, /**< indicates a position or direction upwards (=pref) */
+   after   /**< indicates a position or direction downwards (=next) */
+};
+
+
+/***************************************************************************//**
+ * @brief prototype for a function that checks a chunk to have a given type
+ *
+ * @note this typedef defines the function type "check_t"
+ * for a function pointer of type
+ * bool function(chunk_t *pc)
+ ******************************************************************************/
+typedef bool (*check_t)(chunk_t *pc);
+
+
+/***************************************************************************//**
+ * @brief prototype for a function that searches through a chunk list
+ *
+ * @note this typedef defines the function type "search_t"
+ * for a function pointer of type
+ * chunk_t *function(chunk_t *cur, nav_t nav)
+ ******************************************************************************/
+typedef chunk_t * (*search_t)(chunk_t *cur, nav_t nav);
+
+
+typedef ListManager<chunk_t> ChunkList_t;
+
+
+/* \todo if we use C++ we can overload the following two functions
+ * and thus name them equally */
+
+/**
+ * \brief search for a chunk that satisfies a condition in a chunk list
+ *
+ * A generic function that traverses a chunks list either
+ * in forward or reverse direction. The traversal continues until a
+ * chunk satisfies the condition defined by the compare function.
+ * Depending on the parameter cond the condition will either be
+ * checked to be true or false.
+ *
+ * Whenever a chunk list traversal is to be performed this function
+ * shall be used. This keeps the code clear and easy to understand.
+ *
+ * If there are performance issues this function might be worth to
+ * be optimized as it is heavily used.
+ *
+ * @retval NULL    - no requested chunk was found or invalid parameters provided
+ * @retval chunk_t - pointer to the found chunk
+ */
+static chunk_t
+*chunk_search(chunk_t       *cur,           /**< [in] chunk to start search at */
+              const check_t check_fct,      /**< [in] compare function */
+              const nav_t   nav = CNAV_ALL, /**< [in] code parts to consider for search */
+              const loc_t   dir = after,    /**< [in] search direction */
+              const bool    cond = true     /**< [in] success condition */
+              );
 
 
 /**
  * \brief search a chunk of a given category in a chunk list
  *
- * follows a chunk list until it finds a chunk of a given category.
- * The search direction is determined by parameter dir.
+ * traverses a chunk list either in forward or backward direction.
+ * The traversal continues until a chunk of a given category is found.
  *
- * @retval NULL    - no object found, or invalid parameters provided
- * @retval chunk_t - pointer to the found object
+ * This function is a specialization of chunk_search.
+ *
+ * @retval NULL    - no chunk found or invalid parameters provided
+ * @retval chunk_t - pointer to the found chunk
  */
-static chunk_t *search_chunk(chunk_t         *pc, /**< [in] chunk list to search in */
-                             const c_token_t cat, /**< [in] category to search for */
-                             const bool      dir  /**< [in] search forward=true, backward=false */
-                             );
+static chunk_t
+*chunk_search_type(chunk_t         *cur,           /**< [in] chunk to start search at */
+                   const c_token_t type,           /**< [in] category to search for */
+                   const nav_t     nav = CNAV_ALL, /**< [in] code parts to consider for search */
+                   const loc_t     dir = after     /**< [in] search direction */
+                   );
+
+
+chunk_t
+*chunk_search_typelevel(chunk_t   *cur,           /**< [in] chunk to start search at */
+                        c_token_t type,           /**< [in] category to search for */
+                        nav_t     nav = CNAV_ALL, /**< [in] code parts to consider for search */
+                        loc_t     dir = after,    /**< [in] search direction */
+                        int       level = -1      /**< {in]  */
+                        );
+
+
+/**
+ * \brief searches a chunk that is non-NEWLINE, non-comment and non-preprocessor
+ *
+ * traverses a chunk list either in forward or backward direction.
+ * The traversal continues until a chunk of a given category is found.
+ *
+ * @retval NULL    - no chunk found or invalid parameters provided
+ * @retval chunk_t - pointer to the found chunk
+ */
+static chunk_t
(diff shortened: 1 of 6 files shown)


src/chunk_list.cpp
```diff
@@ -51,6 +51,21 @@ chunk_t *chunk_get_tail(void)
...
 
+/**
+ * \brief Searches for a non-vbrace chunk
+ *
+ * follows a chunk list until it finds a non-vbrace chunk.
+ * The search direction is determined by parameter dir.
+ *
+ * @return pointer to the found chunk or NULL of none found
+ */
+static chunk_t *chunk_get_nvb(chunk_t           *cur,           /**< [in] chunk to start with*/
```
CDanU: Wouldn't it be better to describe the params with `@param` inside the description comment?

mihaipopescu: I'd prefer like that as well.

src/chunk_list.cpp
```diff
@@ -721,32 +736,26 @@ void chunk_swap_lines(chunk_t *pc1, chunk_t *pc2)
...
+{
+   return(chunk_get_nvb(cur, nav, false));
 }
 
 
-/**
- * Gets the prev non-vbrace chunk
- */
-chunk_t *chunk_get_prev_nvb(chunk_t *cur, chunk_nav_t nav)
+static chunk_t *chunk_get_nvb(chunk_t *cur, const chunk_nav_t nav, const bool dir)
```
mihaipopescu: I don't think`dir` as a boolean is a good choice here. Bools have very small semantic value in general anyways.

src/chunk_list.cpp
```diff
@@ -721,32 +736,26 @@ void chunk_swap_lines(chunk_t *pc1, chunk_t *pc2)
...
 {
    chunk_t *pc = cur;
 
    do
    {
-      pc = chunk_get_prev(pc, nav);
-   } while (chunk_is_vbrace(pc));
+      pc = (dir == true) ? chunk_get_next(pc, nav) : chunk_get_prev(pc, nav);
```
mihaipopescu: So for each token, instead of calling a function directly we're also going to check a condition ? Do you think the code is still going to remain performant ?

src/chunk_list.cpp
```diff
@@ -721,32 +736,26 @@ void chunk_swap_lines(chunk_t *pc1, chunk_t *pc2)
...
 
 
-/**
- * Gets the next non-vbrace chunk
- */
-chunk_t *chunk_get_next_nvb(chunk_t *cur, chunk_nav_t nav)
+chunk_t *chunk_get_next_nvb(chunk_t *cur, const chunk_nav_t nav)
 {
-   chunk_t *pc = cur;
+   return(chunk_get_nvb(cur, nav, true));       // \todo check this
```
mihaipopescu: If I'm reading this I won't understand a thing.

src/chunk_list.cpp
```diff
@@ -721,32 +736,26 @@ void chunk_swap_lines(chunk_t *pc1, chunk_t *pc2)
...
 
-   do
-   {
-      pc = chunk_get_next(pc, nav);
-   } while (chunk_is_vbrace(pc));
-   return(pc);
+
+chunk_t *chunk_get_prev_nvb(chunk_t *cur, const chunk_nav_t nav)
+{
+   return(chunk_get_nvb(cur, nav, false));
```
mihaipopescu: ditto

src/chunk_list.cpp
stefan-nu: The problem with @param is that the comment and the code will diverge with time and thus the documentation gets wrong. Besides it adds unnecessary redundancy. But my solution works much better if the parameter may be written on a new line which is currently prohibited by the uncrustify rules.

src/chunk_list.cpp
stefan-nu: I agree. I think I should become an enum.

src/chunk_list.cpp
stefan-nu: In my opinion performance will be no problem. Optimizing for performance should only be done as last step if there are performance issues. To decide the places where optimization is necessary a proper profiling should be done.

Code redundancy is imho a much bigger problem. Currently the same code is present in about 10 places. This makes the code error prone and difficult to extend.

stefan-nu:
Regarding code comprehension I am convinced that the redundancy free code will be much easier to understand. But what you see now is just a first step. There is much more to do.

src/chunk_list.cpp
CDanU: I don't think that the idea is bad, it is just not going the full circle.
* `chunk_get_prev_nvb` and `chunk_get_next_nvb` should be completely removed and replaced with the new `chunk_get_nvb` (or better `chunk_get_non_vbrace`, I still would have too look up what a vbrace is, but at least I will know is some kind of brace)
* the direction check could be a switch that has loop bodies for every case

> Optimizing for performance should only be done as last step if there are performance issues

I hear that all the time but there never actually is a planned last step with open source. Is either improved on or dead forever. 
Unless we know that there is some compiler magic happening that fixes this it will be slower (even if no one will notice it).

stefan-nu: I have already a version on my computer using function pointers. But I showed you a preliminary version first. Give me some more time...

mihaipopescu: Sure, but why not doing it now ?

src/chunk_list.cpp
mihaipopescu: Yes, could be better.

src/chunk_list.cpp
mihaipopescu: > make the code more understandable than faster

the proposed solution is neither.

mihaipopescu:
I really appreciate your efforts to contribute. There are just some changes to be done.

src/chunk_list.cpp
mihaipopescu: > the direction check could be a switch that has loop bodies for every case

this 👍

stefan-nu: Just a bit of patiance please.

stefan-nu:
I pushed the next version which shows already much clearer the direction I am heading.
In my opinion it it still far from finished. But as my last PR was refused due to too many changes I wanted to show you an early version this time. I appreciate every hint for improvement.

stefan-nu:
The latest version I pushed shows rather well what I mean with "reduced redundancy". In my opinion this is a fundamental requirement to achieve error free code. I assume you recognize that my changes are extensive and affect almost all of the original code in the file chunk_list.cpp.

To avoid unnecessary work I kindly ask you to tell me if you want this kind of modification or if you prefer to stick with the current code.

CDanU:
I like that this changes moves the logic from multiple functions into fewer functions, even if the fewer functions get more complicated. 

So far everything looks sane but I would like to have a second look after the single huge commit is broken up into multiple ones. Please separate the new additional comment descriptions for old functions (if possible), minor renames and explicit checks for `NULL` and so on from the actual code logic changes.

I noticed that you sometimes inserted a newline after a function return type. I would like to have several opinions on that. I don't like it that much (including the newline after `static_inline` which was already available in the code). 
@gmaurel should take notice that with the new **cleaner** `chunk_log_msg` some messages now have additional newlines (I don't know how to generate this logs).

mihaipopescu:
In general I'm ok with the idea of doing this kind of refactoring but this PR is very hard to review (even github refuses to display the diff) and if a mistake slips through bisecting won't help without paying the full cost of debugging.

> what I mean with "reduced redundancy"

I think you're talking about code duplication. For me, _code redundancy_ is dead code which I'm 100% in favor of removing.

> In my opinion this is a fundamental requirement to achieve error free code.

Sure, but there is no such thing as "error free code". :)

mihaipopescu:
I was surprised to see `static_inline` as well. I don't like it as well and I don't understand why should we change the code to make some code editor happy ?

Also, on the other hand, they are two completely different concepts and they should be used sparingly. I saw them almost everywhere...

stefan-nu:
There are different topics to discuss:

1. The huge size of the commit: In fact I did not want to put all the changes in one commit. But you forced me to do so with your earlier comments. Honestly I don't like to split the commit as this will most likely indroduce new problems. And I doubt it is worth the effort.

2. The layout of function comments: There is the general question where to put the parameter comments, after the parameter or in a @param line. In my opinion my approach is more reliable and error prove but it works only well if the uncrustify rules allow a newline after the function name and before the parameters which is not the case currently. I added the newline before the function name as this was not prevented by the uncrustify rules. But I do not like it either. I understand if you prefer the old @param approach. It's just that I don't like to commit code which I am not convinced of.

3. Obviously there is no "error free code". But sticking to some general rules like DRY reduces the likelyhood for errors strongly. And I am rather convinced that I did not interoduce a new error. But obviously nobody is perfect.

4. "If a mistake slips through": as all tests pass this is not very likely. And if so it means there is a test missing.

5. Regarding "static inline": I believe it can be removed safely. The compiler usually knows when to inline.

stefan-nu:
I close this PR as it needs some fundamental rework to be merged.

CDanU:
Did you find new problems with this PR? The current ones don't seem to be that fundamental. I would offer split the commit into several ones it this is an issue.

stefan-nu:
I tried to do a test merge locally to see if it worked. And basically I scewed up my code. So I think the approach to do such large modifications is not feasable. At the moment I am not sure how to proceed. You are welcome to take the PR and split it / merge it / or do whatever you like.

CDanU:
Ok I have split your comment into several smaller ones. While doing so I moved some functions in order to have a simple as possible commits. Every commit in its own builds. The last commit is there so you can verify that both branches( your remove_redundancy2 and my remove_redundancy3) are exactly the same (except that I removed the newlines after function return types and static_inline). After I get the message that everything is OK I would rebase the everything and open up a new PR.

https://github.com/CDanU/uncrustify/commits/remove_redundancy3

CDanU:
Here is a cleaned and rebased version: https://github.com/CDanU/uncrustify/commits/refactor-chunk_list

PR #959, opened 2017-01-24
Implementation of sort includes

mihaipopescu:
The implementation introduces 3 prioritised include categories defined by a configurable regex. This idea is borrowed from clang-format.

The sorting order is controlled by the priority of the category first followed by a string comparison with a higher priority on lowercase versus uppercase. We needed a different string comparison algorithm to control case sensitivity in a stable way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ include(CTest)
 if(MSVC)
   add_definitions(/D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_WARNINGS /wd4267)
 elseif(CMAKE_COMPILER_IS_GNUCXX)
+  add_definitions(-std=gnu++0x)
   set(gcc_warning_flags
     -Wall
     -Wextra
@@ -30,6 +31,7 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
     -Wpointer-arith
     -Wcast-qual
     -Wcast-align
+    -Wc++11-extensions
   )
   foreach(flag ${gcc_warning_flags})
     string(REGEX REPLACE "[^a-zA-Z0-9]+" "_" flag_var "CXXFLAG_${flag}")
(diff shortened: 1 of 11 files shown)


mihaipopescu:
The code is based on a draft implementation done around 6 months ago with sorting C++ code in mind. So one of the things that I don't entirely like is the name of the options.

This is a preliminary PR to and receive feedback on code review. We can figure out together what needs to be done next but it can serve as a first implementation that address issue #935.

mihaipopescu:
@gmaurel Do we have a script I should run to format my sources ?

CDanU:
There is a script([scripts/Run_uncrustify_for_sources.sh](https://github.com/uncrustify/uncrustify/blob/master/scripts/Run_uncrustify_for_sources.sh)) that checks if the formatting adheres to this the coding style of this repo. The sources must be uncrustified with [forUncrustifySources.cfg](https://github.com/uncrustify/uncrustify/blob/master/forUncrustifySources.cfg).

Also there is the [scripts/Test_more_Options.sh](https://github.com/uncrustify/uncrustify/blob/master/scripts/Test_more_Options.sh) Script that tests the several options of uncrustify. In your example this script fails because it stll assumes that "There are currently **567** options and minimal documentation" instead of "There are currently **570** options and minimal documentation". So I guess you maybe have to update the [scripts/More_Options_to_Test/help.txt](https://github.com/uncrustify/uncrustify/blob/master/scripts/More_Options_to_Test/help.txt) file?

gmaurel:
The comment is correct. Please use the two tools at home and push again.

mihaipopescu:
Why isn't this mentioned anywhere ? Developers should have instructions what scripts to run and how.

If `Run_uncrustify_for_sources` just does a verification is not going to do any changes. Then why isn't called `Test_sources_are_uncrustified` ?

Maybe we should then have another script (or this one) that applies the new formatting.

gmaurel:
Well, I prefer a script which says me what is not correct.
I prefer to make the necessary changes my self.
Run the script, notice the problem(s) (they are show with diff, meld is a proper to see what is to do),
make the changes until the script  doesn't found any more differences.
This is the way I am working with.

mihaipopescu:
Well, I would prefer the script to generate a patch and if I agree I would apply it.

mihaipopescu:
This problem is related with #898. Let's continue over there.

Issue #899, opened 2017-01-06
Contribute code of conduct


mihaipopescu:
As I was talking in the previous issue that I've opened #898 to have a contribute section in our readme file, we need to add more conventions to enable us to work together more smoothly.

Rules:
- branches should be named after what the work is about
- branches should not be named after the issue number, developer name, etc.
- branch name should contain keywords like `fix`, `feature`, etc.
- it is recommended to do one thing on the branch and get that to master, splitting work if necessary.
- the PR title should represent _what_ is being changed (could be a rephrasing of the branch name if set correctly)
- the PR description should document the _why_ not the _what_
- try to avoid merges, always rebase your work on top of master before creating a PR
- no PR should be approved to be merged in master if the work is incomplete (no fix-up commits). That's why core reviews are for, not for automatic testing by a CI server. Be very critical but not try to over-engineer things.
- the changes should always be accompanied by regression tests (if not possible, why not ?)

Please comment and let's contribute to this list.

gmaurel:
Thanks for the proposals.

CDanU:
I also like this conventions (even when I most likely had broken every one of those at some point). 

One thing that I don't get is why we use merge commits when they are not really needed:
* Merging the master branch into a feature branch: Why not just rebase on the master instead?
* Merging PRs that have no conflicts: There is a rebase option in github.

The other thing that I don't like are PRs with fix-up commits. Split them up and include them into their respective commits.

I get that most of this means one has to do more work while additionally having to have a slightly deeper understanding of git but that would produce a cleaner tree: 
* No fix-up commits that are fixing issues for commit n, n-2, n-5 together in one extra commit.
* Less looking at merge commits to see if and/or which changes where made to resolve conflicts. 
* most likely some other benefits ....

At this point in time there aren't that many PR

mihaipopescu:
@CDanU thanks for your proposals. I've updated the list. I think it needs more iteration until we go and modify README.

Let me know if rebasing is possible in git/github after pushing some changes. What feature/option is that in github. Is it something for the maintainer to care about ?

CDanU:
Yes this is mostly only for the maintainer: https://github.com/blog/2243-rebase-and-merge-pull-requests
I have not used this but as I understand int works like this: If there are no merge conflicts caused by the pr github can instead of pushing the changes with a merge commit do a rebase and just put the commits on top of the tree without introducing a new merge commit.

mihaipopescu:
[...]

PR #960, opened 2017-01-25
Add contributing file

mihaipopescu:
The main motivation is to help define a common working framework for uncrustify contributors.
The need of a contributing file was also mentioned in #898 but the main discussion happened in the comments for issue #899.


jibsen:
Should "That's why core reviews are for" be "That's what code reviews are for"?

mihaipopescu:
Thanks @jibsen, that is correct.

I think we also have to mention `Run_uncrustify_for_sources`.

PR #973, opened 2017-01-28
Refactor chunk_list

CDanU:
Here is another try for #957. I split up the one big comment from @stefan-nu into smaller ones and made some changes here and there.

### DIFF
--- a/src/chunk_list.cpp
+++ b/src/chunk_list.cpp
@@ -137,10 +137,8 @@ chunk_t *chunk_get_prev(chunk_t *cur, nav_t nav)
 
 chunk_t *chunk_dup(const chunk_t *pc_in)
 {
-   chunk_t *pc;
+   chunk_t *pc = new chunk_t; /* Allocate a new chunk */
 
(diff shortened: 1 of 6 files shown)


src/chunk_list.cpp
```diff
@@ -9,34 +9,171 @@
...
 
 
-typedef ListManager<chunk_t> ChunkList;
+typedef ListManager<chunk_t> ChunkList_t;
+
+/** use this enum to define in what direction or location an
+ * operation shall be performed. */
+enum loc_t
+{
+   before, /**< indicates a position or direction upwards (=pref) */
```
mihaipopescu: I like the enum approach. I think the name it's a bit confusing though being in lowercase I thought it was a local variable. What do you say if we use Google naming style for enums, thus we have here `kBefore`, `kAfter` ?

CDanU: Yes, the enums should be improved, but I don't like adding pointless prefixes so I will make them all caps like the majority of the enums in the codebase.

PR #974, opened 2017-01-29
Add currently used C++ standard as badge

CDanU:
preview: [here](https://github.com/CDanU/uncrustify/blob/f826b07eeaf1430412632806c54c4184a421eb52/README.md)

ref: [#935](https://github.com/uncrustify/uncrustify/issues/935#issuecomment-275778512)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@ README for Uncrustify
 
 [![Travis CI](https://img.shields.io/travis/uncrustify/uncrustify/master.svg?style=flat-square&label=Linux)](https://travis-ci.org/uncrustify/uncrustify)
 [![AppVeyor](https://img.shields.io/appveyor/ci/uncrustify/uncrustify/master.svg?style=flat-square&label=Windows)](https://ci.appveyor.com/project/uncrustify/uncrustify)
+<a href="#"><img src="https://img.shields.io/badge/C++-11-blue.svg?style=flat-square"></a>
 
 Post any bugs to the issue tracker found on the project's GitHub page:
   https://github.com/uncrustify/uncrustify/issues


mihaipopescu:
It's ok but I now have second thoughts about this as it would portray the code as an exemplary use of C++11 features rather than as a side-note for the developers that they can use it.

CDanU:
Yes that could be an issue. Maybe something like <a href="#"></a> then ? It is a bit long though.

gmaurel:
Is uncrustify conform to C++11 ?
Are there any tests suite to be sure it is conform?

CDanU:
C++11 is backwards compatible to C++ code. Old C++ code cannot link to new C++11 code but that will not be an issue as Uncrustify is not a library.
Features from C99 and C11 on the other hand could be a problem. `%zu` for the printf like function is a C99  feature that C++11 is missing.

gmaurel:
We are still far from C++11.
It is not appropriate to suggest we are conform.

CDanU:
Well I don't agree that we are not conform because we don't use C++11 everywhere. Old code style code is allowed. I am not saying that this won't cause problems here and there.

I have not made really clear why I wanted to have that badge in the first place so here we go: We use the compiler flag that allows C++11 features. My idea was to make that visible so that new contributors start submitting C++11 code. 
They now can use `auto`, smart pointers, range based loops, enum classes, lambda functions and (almost) all the other goodies that C++11 has to offer.

But I have to admit that we have other problems that somewhat hinder this:
* `%zu` in the log function 
* old gcc version
* missing formatting options (that would need to be used to keep Uncrustify well formatted)

guy-maurel:
Sorry for my confusion. I think there are two aspects about C++11:
1. uncrustify is written with C++11
2. uncrustify is able to parse and format C++11 sources (even if written with FORTRAN)
Which aspect do you speak about?

CDanU:
I speak about the uncrustify source code which __now can__ be written in C++11. I haven't event considered that this can be confused with the ability of Uncrustify to format C++11 code correctly. 
The badge is just too confusing, lets scrap that idea.

mihaipopescu:
Yes, there are two things. Uncrustify already supports formatting some C++11 code features (like auto) but not all of them. Not even all compilers support the full specifications of C++11 standard but I strongly feel that we should move towards supporting more and more features as possible to stay on top. Slowly we should then move towards supporting C++14, whether or not uncrustify is using it or not (even if don't see any reason why we should not start using C++14 already).

On the other hand, as I said before, we don't really have to create the impression that Uncrustify code is an exemplary use of C++11 but as @CDanU said maybe this will encourage contributors to use C++11 features.

PR #976, opened 2017-01-30
Adopt UT tests.

guy-maurel:
The are OK without any change.


mihaipopescu:
One of the reasons why we haven't got rid of the old python test framework is that it can run these _untracked_ tests.

PR #981, opened 2017-02-01
Add the imported list to the CMakeLists.txt

guy-maurel:
To complete the tests

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -30,7 +30,7 @@ foreach(test_lang c-sharp c cpp d java pawn objective-c vala ecma imported)
     if(NOT "${first} " STREQUAL "# ")
       if("${line}" MATCHES "^([0-9]+)([!]?)[ ]+([^ ]+)[ ]+([^ ]+)$")
         set(test_number "${CMAKE_MATCH_1}")
-        set(test_name "${test_number}")
+        set(test_name "${test_lang}_${{test_number}")
 
         set(test_config config/${CMAKE_MATCH_3})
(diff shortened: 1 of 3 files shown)


tests/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ function(add_uncrustify_test name lang config rerun_config input result output)
   )
 endfunction()
 
-foreach(test_lang c-sharp c cpp d java pawn objective-c vala ecma)
+foreach(test_lang c-sharp c cpp d java pawn objective-c vala ecma imported)
```
CDanU: I haven't noticed it in the earlier already closed PRs: Why are we making an extra category for this tests?

gmaurel: I like to show from where the tests are coming.

gmaurel:
cmake generates some errors:
```
CMake Error at tests/CMakeLists.txt:4 (add_test):
  add_test given test NAME "10001" which already exists in this directory.
Call Stack (most recent call first):
  tests/CMakeLists.txt:51 (add_uncrustify_test)
```
Why is it necessary to have distinguish number over *all* the categories?

mihaipopescu:
No and actually we already have made that change on our side https://github.com/Unity-Technologies/uncrustify/commit/1f6de8ca629e8d4c58512f7df60d63fa0a399a65

tests/CMakeLists.txt
CDanU: This is confusing if one expects every cpp test to be in the cpp.test file. 
If you really need this new category than it would be good to have a comment in the cpp.test file that explains that there are additional tests inside imported.test.

PR #988, opened 2017-02-03
Lesswarnings for tokenize

guy-maurel:


src/tokenize.cpp
```diff
@@ -72,28 +72,28 @@ struct tok_ctx
...
+   size_t peek(size_t idx)
    {
       idx += c.idx;
-      return((idx < (int)data.size()) ? data[idx] : -1);
+      return((idx < data.size()) ? data[idx] : 0);
    }
 
 
-   int get()
+   size_t get()
```
stefan-nu: The function `get()` returns an element of data which has type `int`. Is it really wanted to convert the data to `size_t`?
In this case why is the data array not also changed to type `size_t` too?
Would it not be better to use `UINT32` for the data array?

guy-maurel: This is correct.
Even if the tests are OK and the letters, punctuators and digits are always positives
I would like to discuss what could be the better solution.

PR #990, opened 2017-02-04
Refactor enums

CDanU:
Now that we use C++11 I would like to refactor the enums so that they become strongly typed enums. This is also a good opportunity to rename and make them more standardized.

Tell me what you think about this. 
The ones I pushed are the easy cases. It gets more complicated when the enum values are used for calculations. For this cases the enum values need to be casted with the help of std::underlying_type.

### DIFF
--- a/src/chunk_list.cpp
+++ b/src/chunk_list.cpp
@@ -42,9 +42,9 @@ typedef bool (*check_t)(chunk_t *pc);
  *
  * @note this typedef defines the function type "search_t"
  * for a function pointer of type
- * chunk_t *function(chunk_t *cur, nav_t nav)
+ * chunk_t *function(chunk_t *cur, nav_t scope)
  ******************************************************************************/
-typedef chunk_t * (*search_t)(chunk_t *cur, nav_t nav);
+typedef chunk_t * (*search_t)(chunk_t *cur, scope_e scope);
 
 
 /**
@@ -64,13 +64,13 @@ typedef chunk_t * (*search_t)(chunk_t *cur, nav_t nav);
  *
  * @param  cur       chunk to start search at
  * @param  check_fct compare function
- * @param  nav       code parts to consider for search
+ * @param  scope     code parts to consider for search
  * @param  dir       search direction
  * @param  cond      success condition
  * @retval NULL      no requested chunk was found or invalid parameters provided
  * @retval chunk_t   pointer to the found chunk
  */
-static chunk_t *chunk_search(chunk_t *cur, const check_t check_fct, const nav_t nav = CNAV_ALL, const loc_t dir = AFTER, const bool cond = true);
+static chunk_t *chunk_search(chunk_t *cur, const check_t check_fct, const scope_e scope = scope_e::ALL, const loc_t dir = AFTER, const bool cond = true);
 
 
 static void chunk_log(chunk_t *pc, const char *text);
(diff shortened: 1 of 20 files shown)


jibsen:
Is there any benefit to specifically choosing uint8_t as the underlying type rather than the default int?

CDanU:
Well it is smaller and because it is unsigned, negative values can not be used. For this specific enums that does not matter. 
Are there disadvantages when using this type in this case? I was planing to use uint16_t and uint32_t for the bigger enums.

gmaurel:
I have a question:
As example ```enum class brace_stage_e : uint8_t```
Do we have any options to control this line? nl_xx, sp_xx or some more?
Do we need some new ones?

CDanU:
It does not seem so.
C++11 added a ton of new features, C++14 added some on top of that and sometime this year C++17 will also add a significant amount of new stuff.

Other than some formatting options for lambda functions Uncrustify does not have options for most of the new features. I am surprised that there aren't that much complains about this yet.

src/chunk_list.h
```diff
@@ -24,17 +24,17 @@
...
+ * nav_e::ALL (default)
  *  - return the true next/prev
  *
- * CNAV_PREPROC
+ * nav_e::PREPROC
  *  - If not in a preprocessor, skip over any encountered preprocessor stuff
  *  - If in a preprocessor, fail to leave (return NULL)
  */
-enum nav_t
+enum class nav_e : uint8_t
```
mihaipopescu: If we're renaming, then maybe we can give it a more meaningful name.

src/chunk_list.cpp
```diff
@@ -20,7 +20,7 @@ typedef ListManager<chunk_t> ChunkList_t;
 
 /** use this enum to define in what direction or location an
  * operation shall be performed. */
-enum loc_t
+enum class loc_e : uint8_t
```
mihaipopescu: what about `location_e`, `position_e`, `direction_e` ?

src/combine.cpp
```diff
@@ -500,11 +500,11 @@ static chunk_t *get_d_template_types(ChunkStack &cs, chunk_t *open_paren);
 static bool chunkstack_match(ChunkStack &cs, chunk_t *pc);
 
 
-enum PLBfound
+enum class plb_found_e : uint8_t
```
CDanU: I don't know about this one. Maybe @gmaurel can help.

gmaurel: It is really a bad name.
I found do_pl declared at line 5092 as int, with some "enum" values.
P stand for Protocol.
May be you could propose a better name. It would be fine.

CDanU: It keeps the state of the sequence of found angle brackets inside an objective-c class deceleration/definition  (not sure if both or just one of them). I would call it oc-class_angle_state.
`PLBfound` is defined in _combine.cpp_ but only used for `handle_oc_class()`. I would __move__ it into that function and name it `angle_state_e`.

jibsen:
> Are there disadvantages when using this type in this case? I was planing to maybe use uint32_t for the lang enum, uint16_t for the other bigger enums and their fast counterparts for the enums where the values are used to calculate values.

To me, the fixed-size integer types are for when you need the specific sizes and semantics, like processing data, or handling file formats. I doubt there is much space to be saved due to alignment, and 8-bit values may be slower to use, though it's unlikely to make any difference in this case.

Also, the fixed-width integer types are not guaranteed to be available on all platforms.

CDanU:
Ok I played around a bit with small examples and almost every time for a unsigned char example the assembly output had one additional random operation. 
I guess unless using a bunch of densely packed data values the smaller types are not improving anything. I`ll switch everything to int.

stefan-nu:
Using unsigned values has the benefit that enums can be bitwise manipulated without warnings.

gmaurel:
My question is still present:
What style do we wanted to have?
As example ```enum class brace_stage_e : uint8_t
enum class brace_stage_e:uint8_t
enum class brace_stage_e : uint8_t
enum class brace_stage_e :                       uint8_t
enum class brace_stage_e :
   uint8_t
enum class brace_stage_e
   : uint8_t
```

**Comment by CDanU** [comment]
Ahh ok, I would like 1 space before and after the colon and a newlines __after__ the type. 
So this one: 
```cpp
enum class brace_stage_e : uint8_t
```

src/chunk_list.h
CDanU: I changed it to `scope_e`.

src/chunk_list.cpp
CDanU: I changed it to `direction_e`.

PR #1014, opened 2017-02-09
Replace typedef with using

CDanU:
clang-tidy -fix -checks="-*, modernize-use-using"


CDanU:
So it turns out that gcc 4.6 can't handle type/template aliases. I am going to close this one for now.
https://gcc.gnu.org/gcc-4.6/cxx0x_status.html

PR #1015, opened 2017-02-09
Use range based loops

CDanU:
clang-tidy -fix -checks="-*, modernize-loop-convert"


mihaipopescu:
@CDanU Thanks for the effort ! Very much appreciated.

Issue #1008, opened 2017-02-07
Newline lost in templated cast member function


jlee975:
New line deleted before operator, but not in equivalent
(named) member function 

```
	template< class T >
	operator T*() const
	{
		return 0;
	}
```

See operator2.cpp, operator2.cpp.uncrustify, operator2.debug in https://github.com/jlee975/uncrustify-tests

Using uncrustify v0.64_d 1f538888ad9b3adc8a46828ebe2ad6578629df29

CDanU:
Thanks for the repo of uncrustify failures. Is going to take some time to go through all of those :)

jlee975:
I appreciate you looking into it. Sorry to dump 10 bugs on you all at once. I was hoping to fix them myself, but haven't found the time.

I guess I would say nl_func_type_name shouldn't apply here because the "return type" and "function name" are essentially the same thing.

guy-maurel:
May we close the call?

CDanU:
No, my PR mentioned this issue but did not fix it.

CDanU:
@jlee975 Could you test if the changes do what you need?

jlee975:
Seems to work now. Thanks!

PR #1017, opened 2017-02-10
Add gen_config_combinations_uniq_output

CDanU:
Creates from a given set of options all possible option settings combinations, formats files with those and displays how much non equal formatted outputs have been created.


gmaurel:
@CDanU Could you please explain (documentation) what for and to use it?

CDanU:
Ok, I got this idea while checking #1008. Sometimes more than one option is involved in a bug. Going through all possible settings combinations is time consuming so I made a script.
It needs a [config json file](https://github.com/CDanU/uncrustify/blob/e3e4e984573e008e270609b1f2c8b5b81b84cea0/scripts/gen_config_combinations_uniq_output_example.json) to run. 

In this config script one can specify Uncrustify options (`options[]`) with their names and setting types.
The setting types (`option_settings.AT_NUM[-2, -1, 0, 1, 2, 3]`) specify which values will be used while generating the combinations. 

Let say we want to generate settings combinations of `UO_string_replace_tab_chars` and `UO_sp_num_before_tr_emb_cmt`:
```
    "options": [{
        "name": "string_replace_tab_chars",
        "type": "AT_BOOL"
    }, {
        "name": "sp_num_before_tr_emb_cmt",
        "type": "AT_UNUM"
}],
```
we keep the `option_settings` as they are: 
```
{
         "AT_BOOL": ["False", "True"],
         "AT_IARF": ["ignore", "add", "remove", "force"],
         "AT_POS": ["ignore", "join", "lead", "lead_break", "lead_force",
                    "trail", "trail_break", "trail_force"],
         "AT_LINE": ["auto", "lf", "crlf", "cr"],
         "AT_NUM": [-2, -1, 0, 1, 2, 3],
         "AT_UNUM": [0, 1, 2, 3]
     },
```
With this settup following setting combinations will be generated and stored into an specified `out_dir`:


| string_replace_tab_chars 	| sp_num_before_tr_emb_cmt 	|
|--------------------------	|--------------------------	|
| false                    	| 0                        	|
| false                    	| 1                        	|
| false                    	| 2                        	|
| false                    	| 3                        	|
| true                    	| 0                        	|
| true                    	| 1                        	|
| true                    	| 2                        	|
| true                    	| 3                        	|


With those configs uncrustify is called an executed upon the specified `in_files` ( output also written into `out_dir`).
The output than is compared to generate a list with groups of file/setting combinations that produce the same result.

You can run the script with python 2.7 and 3.X. just call 
```
gen_config_combinations_uniq_output.py ./gen_config_combinations_uniq_output_example.json
```
and it will print results and generate the mentioned files. 
Keep in mind that I have set up a cleaning process that removes files based on the set `cleanup_lvl`. With level `1` it will auto remove unnecessary generated files. If you want to see all generated files set it to `0`.

PR #1019, opened 2017-02-14
Remove unnecessary else branches

CDanU:
clang-tidy -fix -checks="-*, readability-else-after-return"

Also removes unreachable code in options.cpp convert_value()


gmaurel:
Every one should have a look to
http://llvm.org/docs/CodingStandards.html

PR #1021, opened 2017-02-15
Refactor casts

CDanU:
clang-tidy -fix -checks="-*, google-readability-casting"

### DIFF
--- a/src/keywords.cpp
+++ b/src/keywords.cpp
@@ -309,8 +309,8 @@ void init_keywords()
 
 static int kw_compare(const void *p1, const void *p2)
 {
-   const chunk_tag_t *t1 = (const chunk_tag_t *)p1;
-   const chunk_tag_t *t2 = (const chunk_tag_t *)p2;
+   const chunk_tag_t *t1 = static_cast<const chunk_tag_t *>(p1);
+   const chunk_tag_t *t2 = static_cast<const chunk_tag_t *>(p2);
 
    return(strcmp(t1->tag, t2->tag));
 }
@@ -318,7 +318,7 @@ static int kw_compare(const void *p1, const void *p2)
 
 bool keywords_are_sorted(void)
 {
-   for (int idx = 1; idx < (int)ARRAY_SIZE(keywords); idx++)
+   for (int idx = 1; idx < static_cast<int> ARRAY_SIZE(keywords); idx++)
    {
       if (kw_compare(&keywords[idx - 1], &keywords[idx]) > 0)
       {
@@ -408,8 +408,9 @@ c_token_t find_keyword_type(const char *word, size_t len)
    key.tag = ss.c_str();
 
    /* check the static word list */
-   const chunk_tag_t *p_ret = (const chunk_tag_t *)bsearch(&key, keywords, ARRAY_SIZE(keywords),
-                                                           sizeof(keywords[0]), kw_compare);
+   const chunk_tag_t *p_ret = static_cast<const chunk_tag_t *>(
+      bsearch(&key, keywords, ARRAY_SIZE(keywords), sizeof(keywords[0]), kw_compare));
+
    if (p_ret != nullptr)
    {
       p_ret = kw_static_match(p_ret);
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -333,8 +333,8 @@ void quick_align_again(void)
          AlignStack as;
          as.Start(100, 0);
          as.m_right_align = pc->align.right_align;
-         as.m_star_style  = (AlignStack::StarStyle)pc->align.star_style;
-         as.m_amp_style   = (AlignStack::StarStyle)pc->align.amp_style;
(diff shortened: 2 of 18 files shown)


src/keywords.cpp
```diff
@@ -309,16 +309,16 @@ void init_keywords()
 
 static int kw_compare(const void *p1, const void *p2)
 {
-   const chunk_tag_t *t1 = (const chunk_tag_t *)p1;
-   const chunk_tag_t *t2 = (const chunk_tag_t *)p2;
+   const chunk_tag_t *t1 = reinterpret_cast<const chunk_tag_t *>(p1);
```
bengardner: Shouldn't static_cast<> work here?  p1 is a (void *).
This comment applies to all conversions to/from (void *).

http://www.cplusplus.com/doc/tutorial/typecasting/

bengardner:
The .gitignore commit shouldn't be combined with the cast commit.  They are unrelated.

gmaurel:
Is it possible to add such a test with clang-tidy to the test define by .travis.yml

src/keywords.cpp
CDanU: Yes you are right, I was a bit concerned about the amount of reinterpret_cast but didn't check if they where actually needed. 
While making the changes I looked closer upon MD5.cpp but couldn't really decide which cast are really needed. I am therefore going to remove all changes made in this file.

CDanU:
I am not sure about this, but surely there is got to be a way to integrate cland-tidy into the travis build process. 
What I do is basically to add `CMAKE_EXPORT_COMPILE_COMMANDS` 
```
cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON 
```

while building Uncrustify to let it generate a `compile_commands.json` file which I then copy/link into the source folder. Clang-tidy needs this file to know how to compile the source files.

Then I run 
```
clang-tidy -fix -checks="-*, google-readability-casting" *.cpp
```
in the source directory.

Travis should not try to fix code so the `-fix` need so be removed. 
With 
```
-checks="-*, google-readability-casting"
``` 
First all checks are disable with `-*` and then specific tests after that can be enabled.
To get a list of all available checks use" `clang-tidy -list-checks -checks="*"`

Clang-tidy only generates warnings for issues that it finds that do not influence the return code. Apparently the option `-warnings-as-errors` should be used to change that but I still get a 0 return code. 
It should be however be possible to redirect the output of clang tidy into a file and check if it stays empty in addition to the return code check.

gmaurel:
Could we begin to put *one* test in one of the travis build?

CDanU:
Yes, a good option to start with should be the `modernize-use-nullptr` check.

Ironically I introduced one NULL while rebasing sorting.cpp:57 with a commit after 7cd5dc44737147243ab4db186f7531f33a18daf2. 

Keep in mind that not all NULL are found by clang tidy. There are plenty of them still in the source code. Especially in headers as I didn't had searched for a solution on how to also run clang tidy on them.

guy-maurel:
Using:
```
clang-tidy '-checks=-* -header-filter=.*' ./src/align.cpp
```
I get
```
1 error generated.
Error while processing /home/guy/Software/uncrustify/src/align.cpp.
/usr/include/stdlib.h:32:10: error: 'stddef.h' file not found [clang-diagnostic-error]
#include <stddef.h>
         ^
```
What is to do?

CDanU:
Sounds like you are missing the `compile_commands.json` file inside the src dir and clang therefore does not know how to compile the file:

> What I do is basically adding the CMAKE_EXPORT_COMPILE_COMMANDS option to cmake
> 
> cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON 
> 
> while building Uncrustify to let it generate a compile_commands.json file. I then copy/link it into the source folder. Clang-tidy needs this file to know how to compile the source files.

gmaurel:
No. I make it and link it.

CDanU:
Ohh your clang tidy call is wrong:

> '-checks=-* -header-filter=.*'

Checks is a arg so it should be -checks="-*", it uses wildcards not regex. With that all filters are be disabled and clang prints the help output. 
The same applies to header-filter, additionally you should be more precise with the headers.

That one should work:
```
clang-tidy -checks="-*, modernize-use-nullptr" -header-filter="./src/*" ./src/align.cpp
```

guy-maurel:
I can now see the messages, but the error is still present:
```
...
/home/guy/Software/uncrustify/src/uncrustify_types.h:257:23: warning: use nullptr [modernize-use-nullptr]
      prev          = 0;
                      ^
                      nullptr
/usr/include/stdlib.h:32:10: error: 'stddef.h' file not found [clang-diagnostic-error]
#include <stddef.h>
         ^
```

CDanU:
Are you on Ubuntu? Maybe you have something related to this: http://stackoverflow.com/questions/20587228

PR #1038, opened 2017-02-23
Prepare the use of clang-tidy.

guy-maurel:
Some more checks will follow.
First use with space.cpp

### DIFF
--- a/scripts/Run_clang-tidy.sh
+++ b/scripts/Run_clang-tidy.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+#
+# 22 februar 2017
+#
+SRC="./src"
+#
+where=`pwd`
+#
+# build the lists
+cd ${SRC}
+list_of_C=`ls *.cpp`
+list_of_H=`ls *.h`
+cd ${where}
+#
+RESULTS="./results"
+#
+rm -rf ${RESULTS}
+mkdir ${RESULTS}
+#
+cp build/compile_commands.json ${SRC}
+#
+list_of_Check="readability-else-after-return"
+#
+for file in ${list_of_H} ${list_of_C}
+do
+  echo "test for "${file}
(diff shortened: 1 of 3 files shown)


scripts/Run_clang-tidy.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+#
+# 22 februar 2017
+#
+SRC="./src"
+#
+where=`pwd`
```
CDanU: Most shell scripts in the scripts dir assume that they are always executed from the root dir. 
I think that this is a bit unflexible.

A solution which I would prefer is to base all relative paths from the dir in which the script itself lies. 
Getting the script directory, unfortunately, is astoundingly non trivial (see [here](https://stackoverflow.com/questions/59895)), especially if systems that either use really old bash versions or don't have it a all need to be supported (bsd?, illumos?, QNX?, apple os). 

Do we actually want to support those systems? I know that one dev of illumos posted some issues here. There also seem to be a bunch of apple users using uncrustify.

I use
```sh
script_dir="$(dirname "$(readlink -f "$0")")"
```
which also is not perfect, but at least not bash dependent.

----------------

Just to be safe I would quote all variables that save paths.

----------------

The German words should be replaced.

scripts/Run_clang-tidy.sh
```diff
@@ -0,0 +1,55 @@
...
+#
+list_of_Check="readability-else-after-return"
+#
+for file in ${list_of_files}
+do
+  echo "test for "${file}
+  OUTPUT="${RESULTS}/${file}.txt"
+  for check in ${list_of_Check}
+  do
+    clang-tidy -checks="-*, ${check}" -header-filter="./src/*" ${SRC}/${file} \
```
CDanU: Here the -header-filter should also use ${SRC}

PR #1044, opened 2017-02-28
all warnings should be treated as errors

guy-maurel:


guy-maurel:
I will rewrite it soon.

PR #1050, opened 2017-03-03
The change is necessary because the word "BYTE" is already defined at

guy-maurel:
cciss_defs.h

No change of functionality

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -875,7 +875,7 @@ static bool read_stdin(file_mem &fm)
 
    fm.raw.clear();
    fm.data.clear();
-   fm.enc = char_encoding_e::ASCII;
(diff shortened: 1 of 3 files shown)


CDanU:
I don't really get why this one was necessary, even if something defines BYTE somewhere this BYTE (in uncrustify) has its own namespace provided by the enum class. The enum class type has always to be provided. There are no ambiguities unless something else also defines `char_encoding_e::BYTE`.

guy-maurel:
Have a look to:
  https://travis-ci.org/uncrustify/uncrustify/jobs/207418009
at line 323:
/home/travis/build/uncrustify/uncrustify/src/uncrustify_types.h:63:4: warning: declaration of ‘BYTE’ shadows a global declaration

PR #1051, opened 2017-03-04
Use a better cast for the enum

guy-maurel:
to not get a warning with the CXX_COMPILER=g++-6


src/output.cpp
```diff
@@ -354,8 +354,13 @@ void output_text(FILE *pfile)
                      int orig_sp = (pc->orig_col - prev->orig_col_end);
                      if ((int)(cpd.column + orig_sp) < 0)
                      {
-                        fprintf(stderr, "FATAL: negative value.\n   pc->orig_col=%zu prev->orig_col_end=%zu\n",
-                                pc->orig_col, prev->orig_col_end);
+                        {
+                           char *outputMessage;
+                           outputMessage = make_message("FATAL: negative value.\n   pc->orig_col=%zu prev->orig_col_end=%zu\n",
+                                                        pc->orig_col, prev->orig_col_end);
+                           fprintf(stderr, "%s", outputMessage);
```
CDanU: I see that printf is used for the fatal messages and not the logging capabilities. Why don't you use std::cerr here? 
The extra pair of braces is weird.

guy-maurel: It is history. We could change it to a latter time.

guy-maurel:
the helper is introduced because the compiler /usr/bin/x86_64-w64-mingw32-g++
 produces a warning for the lines:
```
            fprintf(stderr, "Unmatched BRACE_CLOSE\nat line=%zu, column=%zu\n",
                    pc->orig_line, pc->orig_col);
 "unknown conversion type character ‘z’ in format [-Werror=format]"
```

CDanU:
Yes I know that's why I was confused why you implemented the helper instead of ditching the fprintf for cerr.

CDanU:
It should look somewhat like this:

```cpp
std::cerr << "Unmatched BRACE_CLOSE\nat line=" << pc->orig_line 
          << ", column=" << pc->orig_col << "\n";
```

gmaurel:
Fine. Let us introduce this soon.
I would like to merge the current changes now.

Issue #1030, opened 2017-02-20
alignment of = sign in enum typedef


stleusc:
I'm sure I saw an option that produced the following output. Can't find it anymore :-(

```C
typedef enum
{
    XXXXXXXXXXXXXXXXX      = 0U,
    ZZZZZZZZZZZ            = 1U,
    TTTTTTTT               = 2U
} MyType;
```

Any advice?

stleusc:
I was playing with

align_enum_equ_span=4
align_enum_equ_thresh=4

and different values, but no alignment at all. Maybe a global overall switch missing?

stleusc:
playing with

align_assign_span=1
align_assign_thresh=1

turns on the above formatting but also aligns = in normal code.

There has to be a way to turn it ON in enum but OFF in code, why else have these options?

CDanU:
Please actually read the [ISSUE_TEMPLATE](https://github.com/uncrustify/uncrustify/blob/master/.github/ISSUE_TEMPLATE)
Additionally to your expected result I also need your input.

stleusc:
I'll see that I update the description better.

Example input:

```C
typedef enum
{
    XXXXXXXXXXXXXXXXX = 0U,
    ZZZZZZZZZZZ = 1U,
    TTTTTTTT = 2U
} MyType;
```

I understand the need for some global flag but then there should be a specific flag for EACH suboption.

Right now I don't see a way how I can align enum and NOT align code.

Best result is achieved so far using

```
align_assign_span=1
align_assign_thresh=1
align_enum_equ_span=1
align_enum_equ_thresh=0
```

But it will still align this code as well

```C
x = 3;
xy = 4;
```

will become

```C
x  = 3;
xy = 4;
```

[uncrustify.cfg](http://pastebin.com/Gk3qvjnH)

CDanU:
Yes I agree that we should need more special options for the align_assign functionality which can also operate on their own. 

I tried to achieve proper results with this little prototype commit here: 
https://github.com/CDanU/uncrustify/commit/aa6e24811a84ea2a7ab15032fba69dae1c7110d9

That one breaks some test cases. I would have to understand what the AlignStack class is actually doing but I currently don't have the time for that.

ostraat:
This issue is fixed in PR #1055.  Please take a look and comment.

guy-maurel:
@ostraat Could you create a new PR with it?

PR #1055, opened 2017-03-09
Allow align_enum_equ* without align_assign_span

ostraat:
Fix issue #1030 to allow align_enum_equ_span and
align_enum_equ_threshold to be set without setting align_assign_span.

Remove align_assign_span from enum.cfg to effectively make 33110 a
regression test.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -404,10 +404,14 @@ void align_all(void)
    }
 
    /* Align assignments */
-   align_assign(chunk_get_head(),
-                cpd.settings[UO_align_assign_span].u,
-                cpd.settings[UO_align_assign_thresh].u,
-                nullptr);
+   if ((cpd.settings[UO_align_enum_equ_span].u > 0) ||
+       (cpd.settings[UO_align_assign_span].u > 0))
+   {
+      align_assign(chunk_get_head(),
+                   cpd.settings[UO_align_assign_span].u,
+                   cpd.settings[UO_align_assign_thresh].u,
+                   nullptr);
+   }
 
    /* Align structure initializers */
    if (cpd.settings[UO_align_struct_init_span].u > 0)
@@ -664,11 +668,6 @@ chunk_t *align_assign(chunk_t *first, size_t span, size_t thresh, size_t *p_nl_c
       return(nullptr);
    }
    size_t my_level = first->level;
-
-   if (span == 0)
-   {
-      return(chunk_get_next(first));
-   }
 
    LOG_FMT(LALASS, "%s[%zu]: checking %s on line %zu - span=%zu thresh=%zu\n",
            __func__, my_level, first->text(), first->orig_line, span, thresh);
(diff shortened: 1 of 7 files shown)


src/options.cpp
```diff
@@ -1348,8 +1348,7 @@ void register_options(void)
    unc_add_option("align_assign_thresh", UO_align_assign_thresh, AT_UNUM,
                   "The threshold for aligning on '=' in assignments (0=no limit)", "", 0, 5000);
    unc_add_option("align_enum_equ_span", UO_align_enum_equ_span, AT_UNUM,
-                  "The span for aligning on '=' in enums (0=don't align)\n"
-                  "Note: align_assign_span must be set also.", "", 0, 5000);
+                  "The span for aligning on '=' in enums (0=don't align)\n", "", 0, 5000);
```
ostraat: Oops, I left "\n" at the end.  I'll fix this along with others after comments are made.

CDanU:
Thanks for the PRs, you opened up 3 of them but this one contains the other two. Looking at them I assume that they are not independent, right?

src/options.cpp
CDanU: You can fix your last commit (974d8e6) and force push the changes into you branch 
(`align-enum-equ-bugfix`). GitHub will integrate the changes and fire up a new testrun.

ostraat:
Correct, they depend on each other.  They each do something unique, so I tried making each based on master, but some other tests failed.  Should the other two PRs be abandoned?  Not sure what the right process is...

src/options.cpp
ostraat: So that will tack a 4th commit on this PR, right?  My experience with Git has been with Gerrit, so I'm used to having all my fixes squashed into one commit.  But that's not how it works here, since my changes are public in my fork, and it looks like this project merges from others' forks.  Is that correct?

CDanU: Github tracks changes on the branch of the PR, if you add an additional commit it will be added here.
Since we don't do often squashes here (if at all) I recommend a different approach where you fix you mistakes, and integrate them into your old commits.
The change that will be necessary here are done in the last commit so you can stage the changes and use `git commit --amend --no-edit`.
This will integrate the staged changes into you last commit, as this changes the history of your commits and you will need do a force (`-f`) push.

CDanU:
Ok, keep them for now, this will be decide later.

src/options.cpp
ostraat: Okay, fixed it.  Thanks!

src/align.cpp
```diff
@@ -654,7 +655,7 @@ void align_preprocessor(void)
 } // align_preprocessor
 
 
-chunk_t *align_assign(chunk_t *first, size_t span, size_t thresh)
+chunk_t *align_assign(chunk_t *first, size_t span, size_t thresh, size_t *p_nl_count)
 {
```
guy-maurel: What is pl_nl_count for?
You use it only with NULL

ostraat: It is for storing the number of new lines found in a sub-level.  The first level gets NULL since it has no parent level.  Then &sub_nl_count is passed to the sub-level's (recursive) call to align_assign.  When it returns, this value is passed to AlignStack::NewLines functions to help determine if the configured span has been exceeded.

I ported this functionality from the existing align_var_def_brace function.

Issue #1005, opened 2017-02-07
Lose the space before global scope operator in friend


jlee975:
Space is kept elsewhere in the code:

```
friend void ::test::swap< >(future< T >&, future< T >&);
```

See friend.cpp, friend.cpp.uncrustify, friend.debug in https://github.com/jlee975/uncrustify-tests

Using uncrustify v0.64_d 1f538888ad9b3adc8a46828ebe2ad6578629df29

CDanU:
This also seems to be caused by `sp_before_dc = remove`. 
Additionally `sp_angle_paren_empty = remove` also does not work.

gmaurel:
The bug is fixed

jlee975:
Thanks

PR #1063, opened 2017-03-19
Fix the bug #1005

guy-maurel:


tests/output/cpp/34141-bug_1005.cpp
```diff
@@ -0,0 +1 @@
+friend void ::test::swap< >(future< T >&, future< T >&);
```
CDanU: Shouldn't this be 
```cpp
friend void ::test::swap<>(future< T >&, future< T >&);
```
(no space between angles) with 
```sql
sp_angle_paren_empty = remove
```
?

src/space.cpp
```diff
@@ -487,11 +487,19 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
...
 
-      if ((first->type == CT_WORD) || (first->type == CT_TYPE) || (first->type == CT_PAREN_CLOSE) ||
-          CharTable::IsKw1(first->str[0]))
-      {
-         log_rule("sp_before_dc");
-         return(cpd.settings[UO_sp_before_dc].a);
+      // Issue #1005
+      if (!((first->type == CT_TYPE) &&
+            (first->prev != nullptr) &&
+            (first->prev->type == CT_FRIEND)))
```
CDanU: That is a bit convoluted, the `! ` needs some De Morgan here. 
It should be: 
```cpp
first->type != CT_TYPE || first->prev == nullptr || first->prev->type != CT_FRIEND
```
right?

CDanU: What happens if there are two or more spaces between `void ::`? Remove and force should reduce this down to one space.

tests/output/cpp/34141-bug_1005.cpp
guy-maurel: No, here is ```sp_inside_angle``` the option you need.

CDanU: Yes , you are right, should have read the description.
> sp_angle_paren_empty: Add or remove space between '<>' and '()' as found in 'new List<byte>();'

src/space.cpp
guy-maurel: Right. Thanks.

src/space.cpp
```diff
@@ -488,18 +488,22 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
...
-         }
+      /* '::' at the start of an identifier is not member access, but global scope operator.
+       * Detect if previous chunk is a type and previous-previous is "friend"
+       */
+      if ((first->type == CT_TYPE) &&
+          (first->prev != nullptr) &&
+          (first->prev->type == CT_FRIEND))
+      {
+         log_rule("FORCE");
+         return(AV_FORCE);
```
CDanU: I don't think that this is 100% as it should be. 
Sorry, I should have mentioned this earlier, the problem is that in only in certain cases (the one you are checking here) combined with `UO_sp_before_dc = remove` that that remove needs to be changed to force. In every other case the settings should keep on working, including the case when `UO_sp_before_dc = add`. 
Certainly there will be someone out there who likes to keep his `n > 1` spaces.

PR #1067, opened 2017-03-23
more logging as a define

guy-maurel:
Thanks to @CDanU for the proposal

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,43 +4,43 @@ sudo: false
 matrix:
   include:
     - env: BUILD_TYPE=debug
-#    - env: BUILD_TYPE=release
+    - env: BUILD_TYPE=release
 
-#    # GCC 6
-#    - env: C_COMPILER=gcc-6 CXX_COMPILER=g++-6
-#      addons:
-#        apt:
-#          sources:
-#            - ubuntu-toolchain-r-test
-#          packages:
-#            - gcc-6
-#            - g++-6
-#
-#    # Clang 3.8 address sanitizer
-#    - env: BUILD_TYPE=asan C_COMPILER=clang-3.8 CXX_COMPILER=clang++-3.8
-#      addons:
-#        apt:
-#          sources:
-#            - ubuntu-toolchain-r-test
-#            - llvm-toolchain-precise-3.8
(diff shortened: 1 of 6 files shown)


CDanU:
Nope, the proposal was from @stefan-nu.

guy-maurel:
Sorry.

stefan-nu:
Cool. 

One more remark: You could write log_indent(); instead of log_indent;
Currently this makes no difference at all. But it makes the thing look like a function call even though it is a define.  If you later want to change the define into a function it is not necessary to touch all calls but only the define.

guy-maurel:
AppVeyor produces some problems with the last change.
Can anybody explain it to me?

CDanU:
Looking at the code I also don't see what would cause issues. 
The only thing that worries me is the static `frm` inside the macro, it should be passed as string param.

It would be nice if @stefan-nu could debug this under windows.

stefan-nu:
Unfortunately I do not use Visual Studio. I am running Eclipse with MinGW. I would try to compile your commit but I don't know how I can clone it?

CDanU:
MinGW is already tested by travis, so I don't know if that would help.

Here is one way how you could fetch a branch from another repo without having to fight with merges:
make a new branch && checkout into it
reset it to a to a commit prior to forking
pull --rebase remoteUrl remote branch

For guys repo that would be:
git pull --rebase https://github.com/guy-maurel/uncrustify moreLogging

With this approach all you changes are kept apart from the stuff that you are pulling in.

guy-maurel:
Removing the lines where log_indent() and log_indent_tmp() are used modify the appVeyor run.
This is a stack overflow, which occurs at different part of the test file.
This occurs before indent_text is called.
It might be the use of format %zu at other places.

CDanU:
I have tested 98c46415e2c68afaf8d07c6c1361641f61d1a994 with my student version of  VisualStudio 2015 (Version 14.0.24720.00 Update 1) and everything passes. Could this be a problem with AppVeyor?

guy-maurel:
Could be. Thanks for the message.
But how to go on?
I try some more little changes to find where it occurs.

PR #1069, opened 2017-03-25
Compute the uncrustify version

guy-maurel:


CMakeLists.txt
```diff
@@ -116,27 +116,24 @@ endif()
...
   else()
-    if (${make_version_output} MATCHES "full version: (.*)\n tag version: (.*)")
-      set(full_version ${CMAKE_MATCH_1})
-      set(tag_version ${CMAKE_MATCH_2})
-      message(STATUS "Uncrustify full version: ${full_version}")
-      message(STATUS "Uncrustify  tag version: ${tag_version}")
-    endif()
-    set(PACKAGE_VERSION "${full_version}")
+    set(FULL_VERSION "undefined")
+    set(PACKAGE_VERSION "undefined")
```
CDanU: I have looked at this, the make_version.py and git describe a bit more and come to the conclusion that this approach is problematic.

Something is going wrong when the else branch is triggered, and I personally would rather prefer to throw an error message there which aborts everything:

```cmake
message(FATAL_ERROR "message to display" ...)
``` 
> FATAL_ERROR    = CMake Error, stop processing and generation

&nbsp;&nbsp;&nbsp;than having someone out there with an undefined Uncrustify version.

Dropping the `--long` in make_version.py inside :
```ruby
git describe --long --always --dirty
```
&nbsp;&nbsp;&nbsp;would produce short version strings for tags and long version strings on non tagged commits:

release: `uncrustify-0.64`
current: `uncrustify-0.64-511-gde474b8c`

I would prefer to __always__ have this amount of minimal information. If we only need that we maybe could drop the `make_version.py` file altogether, assuming that we don't need to also manage _hg_. As far as I am aware there is no official _hg_ repository of Uncrustify, right? 

Contrary to this idea (removing the `make_version.py` file) would be to do the version string processing entirely in that file. This would:
- get rid of the `#ifdef DEBUG` checks
- create the possibility to add a `_d` postfix to `uncrustify` to differentiate between debug and non debug builds
- allow to remove the `g` prefix from the commit hash 
(it is there to differentiate between different SCM systems, g = git)

release: `uncrustify_d-0.64`
current: `uncrustify_d-0.64-511-de474b8c`

guy-maurel: Thanks for the comment.
My idea is to get some **more** informations from the user who create an issue message
and join the file created by the -p option.

Issue #1070, opened 2017-03-25
New option request


demothorg:
I'm sorry if the duplicate and for my english.

I use this options (and indent size = 4):
```
mod_full_brace_if=remove
mod_full_brace_nl=2
```

But for code like this (multiline condition) I want to have braces, because otherwise condition merges with statement.
```
    if (a < 0 || a >= 10 ||
        b < 0 || b >= 100 ||
        c < 0 || c >= 1000 ||
        d < 0 || d >= 1000)
    {
        return -1;
    }
```

Can you add an option (something like `mod_full_brace_cond_nl=2`) that means:
"Don't remove braces around statements if condition (in `if` and `while` operators) span N newlines"?

CDanU:
I am working on this.

CDanU:
Do you actually need an unum option here? It looks like determining if the condition is a mulit line condition should be enough. I don't see the point in blocking brace removal only after lets say for example 4 lines.

demothorg:
> It looks like determining if the condition is a mulit line condition should be enough.

Do you mean the option "Don't remove braces if condition is multiline" { True, False } ?
If so, it sounds good.

demothorg:
Sorry, I just noticed the merge request...
I tested this - works fine. Thank you!

PR #1075, opened 2017-03-29
Add option mod_full_brace_nl_block_rem_mlcond

CDanU:
Blocks removal of braces if the parenthesis of if/for/while/.. span multiple lines.

Affected options:
- mod_full_brace_for
- mod_full_brace_if
- mod_full_brace_if_chain
- mod_full_brace_if_chain_only
- mod_full_brace_while
- mod_full_brace_using

Not affected options:
- mod_full_brace_do
- mod_full_brace_function

ref: #1070

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -16,6 +16,7 @@
 #include "uncrustify.h"
 #include "combine.h"
 #include "newlines.h"
+#include "chunk_list.h"
 
 
 /**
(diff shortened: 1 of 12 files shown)


src/braces.cpp
```diff
@@ -98,6 +99,42 @@ static chunk_t *mod_case_brace_add(chunk_t *cl_colon);
...
+           && brace->parent_type != CT_FUNC_CLASS_DEF
+           && brace->parent_type != CT_FUNC_DEF))
+   {
+      return(false);
+   }
+
+   // find paren pair of the if/for/while/...
+   auto paren_close = chunk_get_prev_type(brace, paren_t, brace->level, scope_e::ALL);
+   auto paren_open  = chunk_skip_to_match_rev(paren_close, scope_e::ALL);
+
```
CDanU: I like Herb Sutters [almost-always-auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) approach.

src/newlines.h
```diff
@@ -35,4 +35,6 @@ chunk_t *newline_force_after(chunk_t *pc);
 void newline_del_between(chunk_t *start, chunk_t *end);
 chunk_t *newline_add_between(chunk_t *start, chunk_t *end);
 
+int newlines_between(chunk_t *pcS, chunk_t *pcE, scope_e scope = scope_e::ALL);
```
CDanU: Yes you are right I really should had add some documentation for the new code, I'll make a PR for that.

src/braces.cpp
stefan-nu: I didn't know the "almost-always-auto" approach. But when looking a the resulting code I think this approach misses the point and almost certainly leads to bugs. As I said before reading the code gets much more difficult as the types are not clear anymore. But there are situations where it gets even worst. Imagine
`auto  var = 0`
Is this signed or unsigned. What length does var have? Only the compiler knows. The programmer might assume something different and writes broken code.

Anyway I don't want to start a concept discussion here. But me personlly I stick to the fact that it is more important that the human understands the code than the computer.

Issue #1081, opened 2017-03-31
C: The 3 expressions within a for loop do not align together on the three separate lines


andrewhit:
Today I downloaded the latest uncrustify package and installed on my Linux VM.

Here is the link to the 4 files, debug.files, ngx_http.c (the orignal), ngx_http.c.unc(the generated result) and ngx_http.c.exp(the expected result).

[4 files](https://gist.github.com/andrewhit/722ebb8c11f789642c4143233de84bd6)
[debug.txt](https://gist.github.com/andrewhit/722ebb8c11f789642c4143233de84bd6#file-debug-txt)
[ngx_http.c](https://gist.github.com/andrewhit/722ebb8c11f789642c4143233de84bd6#file-ngx_http-c-the-original-file)
[ngx_http.c.unc](https://gist.github.com/andrewhit/722ebb8c11f789642c4143233de84bd6#file-ngx_http-c-unc-the-generated-result)
[ngx_http.c.exp](https://gist.github.com/andrewhit/722ebb8c11f789642c4143233de84bd6#file-ngx_http-c-exp-the-expected-result)

I have descried the issues within ngx_http.c.exp in detail.

Additionally there is no version info that was displayed.

azhuang@pek2-office-04-dhcp52:~/test$ uncrustify --version
uncrustify

andrewhit:
Could you tell me what cmake command can be executed for your below statement?

"CMake calls make_version.py which now can generate the following version string formats:"

CDanU:
When the referenced commit is merged, building Uncrustify the usual way (inside uncrsutify/build, cmake .. && make && make test) will call `make_version.py` which generates a propper version string. 
If you want to you can just manually call [make_version.py](https://raw.githubusercontent.com/CDanU/uncrustify/7ac8e54198faa1fb54ec86db4b748c280d075873/scripts/make_version.py) and it will print out the version string.

andrewhit:
azhuang@pek2-office-04-dhcp52:~/tmp/uncrustify$ ./scripts/make_version.py
current: 0.64.547-g4fdddba
release: 0.64

andrewhit:
Previously I downloaded uncrustify zip file directly so there is no .git file and no version info output when executing # uncrustify --version.

After checking make_version.py, I have to use git clone to download and .git exists under uncrustify directory. And # uncrustify --version works fine.

azhuang@pek2-office-04-dhcp52:~/test$ uncrustify --version
[...]

Issue #1074, opened 2017-03-29
align_var_def_star_style and align_var_def_amp_style ignored


Massimo-B:
Hi, switching these 2 options doesn't make a difference:
```sh
# How to align the star in variable definitions.
#  0=Part of the type     'void *   foo;'
#  1=Part of the variable 'void     *foo;'
#  2=Dangling             'void    *foo;'
align_var_def_star_style        = 2        # unsigned number
# How to align the '&' in variable definitions.
#  0=Part of the type
#  1=Part of the variable
#  2=Dangling
align_var_def_amp_style         = 2        # unsigned number
```

So setting both to 2 this works as expected:
```cpp
(cfg="https://bitbucket.org/massimo/test/raw/61a0b91ee51726239c87566467f3935fb38b41a5/crusty-test_011.cfg" ; input="https://bitbucket.org/massimo/test/raw/61a0b91ee51726239c87566467f3935fb38b41a5/crusty-test_011.h" ; uncrustify -v ; curl -s "$input" ; curl -s "$input" | uncrustify -l cpp -c <(curl -s "$cfg") )
uncrustify 0.64_d

class MyClass
{
    public:
        MyClass ( const QString &_message );
        QMainWindow     *mainWindow;
        const char*     tag;
};

Parsing: 144 bytes (144 chars) from stdin as language CPP
class MyClass
{
public:
    MyClass(const QString &_message);
    QMainWindow *mainWindow;
    const char  *tag;
};
```
Changing both options to 0, the result is different but not as expected and not consistent:
```cpp
(cfg="https://bitbucket.org/massimo/test/raw/d9fd68158db1e88315cc0bbe5907fcccd266e2f0/crusty-test_011.cfg" ; input="https://bitbucket.org/massimo/test/raw/61a0b91ee51726239c87566467f3935fb38b41a5/crusty-test_011.h" ; uncrustify -v ; curl -s "$input" ; curl -s "$input" | uncrustify -l cpp -c <(curl -s "$cfg") )
uncrustify 0.64_d

class MyClass
{
    public:
        MyClass ( const QString &_message );
        QMainWindow     *mainWindow;
        const char*     tag;
};

Parsing: 144 bytes (144 chars) from stdin as language CPP

class MyClass
{
public:
MyClass ( const QString &_message );
QMainWindow     *mainWindow;
const char*     tag;
};
```
My current build is from
```git
*   commit de474b8 (HEAD -> master, origin/master, origin/HEAD)
```
[...]

PR #1083, opened 2017-03-31
Simplify version string generation and handling

CDanU:
__Depends upon:__ #1102 

-------------------

CMake calls `make_version.py` which now can generate the following version
string formats:

| Case | Output |
|---|---|
| last commit has a tag | Uncrustify-0.64 |
| last commit is preceding a tagged commit | Uncrustify-0.64-518-0dda18c9 |
| last _N_ commits are pulled and no tags are available in the repository | Uncrustify-0dda18c |
| __E:__ make_version.py error | Uncrustify-0.64_f |

If the repository is in a dirty state a `-dirty` suffix is attached
```
Uncrustify-0.64-dirty
```

If CMAKE_BUILD_TYPE is set to Debug a `_d` suffix is attached to Uncrustify:
```
Uncrustify_d-0.64
```

__Note__ that the `_d` suffix is handled while CMake is running, Visual Studio builds will always have this generated prefix, even if the build type is changed in the .sln file.

Also fixes these Windows problems when executing `make_version.py`:
- os.EX_* is cannot be imported
- return error codes with sys.exit()

### __NOTE__:  If this is merged all _new_ tags must adhere the _current_ syntax, adding differently formatted tags <s>will break the regex and therefore the build.</s> __E:__ will use the fallback version string.
---------------
ref: [#1069](https://github.com/uncrustify/uncrustify/pull/1069#discussion_r10804150), #1081, #1074, #776


CDanU:
This kinda gets annoying, I am fairly certain that the changes I have made in `make_version.py` should work, they do locally on my windows and linux machines.

For some reason on __AppVeyor__ `check_output` and `Popen` return some random numbers instead of strings:
>74: Regex version match failed on: __.4d26814__

CDanU:
@andrewhit posted:
> Previously I downloaded uncrustify zip file directly so there is no .git file and no version info output when executing # uncrustify --version.

What do you guys think about this? 
Personally I'd obviously would like to have as few dependencies as possible, but seeing that we already need CMake and Python to build Uncrustify I don't think that git would increase to burden. My impression is that there is no one out there who is working with source code without having the  possibility to get access to git.

Making the build depending on the git history is unusual though. I previously didn't consider this.
I think the solution to this would be to (re)introduce a fallback string into the CMakeLists.txt file.
Something like:
```cmake
if( NOT CMAKE_BUILD_TYPE )
  set( CMAKE_BUILD_TYPE Release )
endif()
set( FALLBACK_VERSION "Uncrustify-0.64_f" )

find_package(PythonInterp REQUIRED)
execute_process(
  COMMAND ${PYTHON_EXECUTABLE} scripts/make_version.py ${CMAKE_BUILD_TYPE}
  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
  RESULT_VARIABLE make_version_error
  OUTPUT_VARIABLE make_version_output
)

message(STATUS )
if (make_version_error)
    message(WARNING "${make_version_error}: ${make_version_output}, scripts/make_version.py doesn't work properly")
    set( CURRENT_VERSION ${FALLBACK_VERSION} )
else()
    string(STRIP ${make_version_output} CURRENT_VERSION)
endif()
```

PR #1085, opened 2017-04-01
Fix save_option_file_kernel

CDanU:
Now params 'withDoc' and 'only_not_default' work together.


gmaurel:
Could you add a (or more) tests in the script Test_more_Options.sh ?

CDanU:
I use this for the emscripten interface. 

Inside the normal Uncrustify build the `save_option_file_kernel` function is only used with either:
```cpp
save_option_file_kernel(pfile, false, true); 
```
or
```cpp
save_option_file_kernel(pfile, true, false);
```
I will add test for both of those cases.

PR #1099, opened 2017-04-06
Fix mod_full_brace_nl_block_rem_mlcond

CDanU:
Previously this option blocked brace removal if the parenthesis spanned
only one line.

As suggested by @setfan-nu (https://github.com/uncrustify/uncrustify/pull/1075), documentation is added and the functions where restructured to be more concise.


CDanU:
@gmaurel Is it somehow possible for me to see what AppVeyor is creating during the test run?

gmaurel:
Yes:
```
-- Found PythonInterp: C:/Python35-x64/python.exe (found version "3.5.3") 
  File "scripts/make_version.py", line 8, in <module>
    from os import EX_OK, EX_USAGE, EX_IOERR
ImportError: cannot import name 'EX_OK'
CMake Warning at CMakeLists.txt:129 (message):
```
But there is no message.
At the end of the test phase:
```
The following tests FAILED:
	 42 - c-sharp_12102 (Failed)
Errors while running CTest
```

CDanU:
I meant the Uncrustifyed files generated in the testrun that differ from the expected result.

The first error you posted is the currently broken make_version.py and for the second error I'd like to see the difference of generated and expected file.

gmaurel:
No.
We have to modify the test to run a diff command

CDanU:
Well I don't know what to do here, I commented out every change that I have done. The functions just return false and I still somehow get `Uncrustify error: Access violation`

CDanU:
I finally found the bug, see 6d35fb5. I am surprised that `gcc` and `clang` did not complain about this.

PR #1120, opened 2017-04-24
Test the default values.

guy-maurel:
The default values for ```code_width``` and ```cmt_width``` are not in the declared interval.
As we need the value zero, change the interval.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1693,6 +1693,7 @@ void indent_text(void)
             }
             if (cpd.settings[UO_indent_func_param_double].b)
             {
+               // double is: Use both values of the options indent_columns and indent_param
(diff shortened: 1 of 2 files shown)


CDanU:
I had the impression that those min/max values where not used, and basically are dead code, do you intend to change that?

gmaurel:
Am 24.04.2017 um 19:18 schrieb CDanU:
> I had the impression that those min/max values where not used, and basically are dead code, do you intend to change that?
> 
The values are used to control the value(s) used by the "user" who writes a configuration file.
It is also used with universalindentgui
What do you mean with "dead code"?

CDanU:
I did not know those variables where used anywhere.

PR #1132, opened 2017-04-29
Add check for CT_DC_MEMBER CT_WORD CT_STAR sequence

CDanU:
to convert CT_WORD into CT_TYPE
and CT_STAR into CT_PTR_TYPE

ref:#1112


CDanU:
I am actually currently not sure if set_chunk_type(pc, CT_TYPE); is the right thing to do

PR #765, opened 2016-09-23
Ordering of property declaration attributes (ObjC)

Krizai:
Based on existing pull request ( https://github.com/uncrustify/uncrustify/pull/217 )
I've extracted the topic related stuff from it and added support for a few new attributes.

PR #1136, opened 2017-05-05
Fixes for obj properties

Krizai:
Based on the comments to https://github.com/uncrustify/uncrustify/pull/765:

Added missing properties
Fixed work of mod_sort_oc_properties with sp_after_comma

### DIFF
--- a/scripts/More_Options_to_Test/help.txt
+++ b/scripts/More_Options_to_Test/help.txt
@@ -70,6 +70,6 @@ Note: Use comments containing ' *INDENT-OFF*' and ' *INDENT-ON*' to disable
       processing of parts of the source file (these can be overridden with 
       enable_processing_cmt and disable_processing_cmt).
 
-There are currently 581 options and minimal documentation.
+There are currently 582 options and minimal documentation.
 Try UniversalIndentGUI and good luck.
 
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -5827,12 +5827,13 @@ static void handle_oc_property_decl(chunk_t *os)
       chunk_t                 *next       = chunk_get_next(os);
       chunk_t                 *open_paren = nullptr;
 
-      std::vector<ChunkGroup> thread_chunks;      // atomic/nonatomic
+      std::vector<ChunkGroup> class_chunks;       // class
+      std::vector<ChunkGroup> thread_chunks;      // atomic, nonatomic
       std::vector<ChunkGroup> readwrite_chunks;   // readwrite, readonly
       std::vector<ChunkGroup> ref_chunks;         // retain, copy, assign, weak, strong, unsafe_unretained
       std::vector<ChunkGroup> getter_chunks;      // getter
       std::vector<ChunkGroup> setter_chunks;      // setter
-      std::vector<ChunkGroup> nullability_chunks; // nonnull/nullable
+      std::vector<ChunkGroup> nullability_chunks; // nonnull, nullable, null_unspecified, null_resettable
 
 
       if (next->type == CT_PAREN_OPEN)
@@ -5847,63 +5848,28 @@ static void handle_oc_property_decl(chunk_t *os)
          // now.
          while (next != nullptr && next->type != CT_PAREN_CLOSE)
          {
-            if (next->type == CT_WORD)
+            if (next->type == CT_OC_PROPERTY_ATTR)
             {
-               if (chunk_is_str(next, "atomic", 6))
+               if (chunk_is_str(next, "atomic", 6) ||
+                   chunk_is_str(next, "nonatomic", 9))
                {
                   ChunkGroup chunkGroup;
                   chunkGroup.push_back(next);
                   thread_chunks.push_back(chunkGroup);
                }
-               else if (chunk_is_str(next, "nonatomic", 9))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  thread_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "readonly", 8))
+               else if (chunk_is_str(next, "readonly", 8) ||
+                        chunk_is_str(next, "readwrite", 9))
                {
                   ChunkGroup chunkGroup;
                   chunkGroup.push_back(next);
                   readwrite_chunks.push_back(chunkGroup);
                }
-               else if (chunk_is_str(next, "readwrite", 9))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  readwrite_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "assign", 6))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  ref_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "retain", 6))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  ref_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "copy", 4))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  ref_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "strong", 6))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  ref_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "weak", 4))
-               {
-                  ChunkGroup chunkGroup;
-                  chunkGroup.push_back(next);
-                  ref_chunks.push_back(chunkGroup);
-               }
-               else if (chunk_is_str(next, "unsafe_unretained", 17))
+               else if (chunk_is_str(next, "assign", 6) ||
+                        chunk_is_str(next, "retain", 6) ||
+                        chunk_is_str(next, "copy", 4) ||
+                        chunk_is_str(next, "strong", 6) ||
+                        chunk_is_str(next, "weak", 4) ||
+                        chunk_is_str(next, "unsafe_unretained", 17))
                {
                   ChunkGroup chunkGroup;
                   chunkGroup.push_back(next);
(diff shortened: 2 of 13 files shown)


CDanU:
Thanks for the fixes, the build currently fails because your changes have not been formatted according to this repos style format. Please use Uncrustify on the changed files with the [forUncrustifySources.cfg](https://github.com/uncrustify/uncrustify/blob/master/forUncrustifySources.cfg) config.

Krizai:
@CDanU Thank you. Now it's complaining about the output of Test_more_Options.sh script. I can't find information about the purpose of this script and the best way to make it happy. Can you help me with this ?

CDanU:
This script tests if the command line options work.

Here is the script itself: [Test_more_Options.sh](https://github.com/uncrustify/uncrustify/blob/master/scripts/Test_more_Options.sh)

It operates on the files in `scripts/Config`, `scripts/Output`, `scripts/Input` and `scripts/More_Options_to_Test/`

You have changed/added some option descriptions so you need to edit `scripts/More_Options_to_Test/show_config.txt` and ...

CDanU:
Ohh, I forgot: You need to remove the version strings in the mentioned files 
> # Uncrustify-0.64-592-b7fc860-dirty

src/tokenize_cleanup.cpp
```diff
@@ -36,6 +36,10 @@ static void check_template(chunk_t *start);
  */
 static chunk_t *handle_double_angle_close(chunk_t *pc);
 
+void cleanup_objc_property_with_parens(chunk_t *start);
+void mark_selectors_in_property(chunk_t *start);
+void mark_attributes_in_property(chunk_t *start);
```
CDanU: Some documentation here would be nice.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
            (end != NULL) ? get_token_name(end->type) : "<null>");
    set_chunk_type(start, CT_COMPARE);
 } // check_template
+
+
+void cleanup_objc_property_with_parens(chunk_t *start)
+{
+   chunk_t *next = chunk_get_next_ncnl(start);
+
+   set_chunk_parent(next, start->type);
```
CDanU: I am not entirely sure, but this looks like it could need a nullptr check for next.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+   }
+   mark_selectors_in_property(start);
+   mark_attributes_in_property(start);
+}
+
+
+void mark_selectors_in_property(chunk_t *start)
+{
+   chunk_t *tmp = chunk_get_next_type(start, CT_PAREN_OPEN, start->level);
+
```
CDanU: I am not entirely sure, but this looks like it could need a nullptr check for tmp.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+   while (tmp->type != CT_PAREN_CLOSE)
+   {
+      if (tmp->type == CT_WORD &&
+          (memcmp(tmp->str.c_str(), "setter", 6) == 0 ||
+           memcmp(tmp->str.c_str(), "getter", 6) == 0)
+          && tmp->next->type == CT_ASSIGN)
```
CDanU: I am not entirely sure, but this looks like it could need a nullptr check for next.
Does obj-C allow comments/newlines in between the setter and the assign ? Something like `setter /* blub */ =`?

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+      {
+         tmp = tmp->next->next;
```
CDanU: One of both next can be nullptr.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+         tmp = tmp->next;
+      }
+   }
+}
+
+
+void mark_attributes_in_property(chunk_t *start)
+{
+   chunk_t *tmp = chunk_get_next_type(start, CT_PAREN_OPEN, start->level);
+
```
CDanU: nullptr check for tmp

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+   while (tmp->type != CT_PAREN_CLOSE)
+   {
+      if ((tmp->type == CT_COMMA || tmp->type == CT_PAREN_OPEN) &&
+          tmp->next->type == CT_WORD)
```
CDanU: Also nullptr check and the in between comment/ newline problem

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1055,73 @@ static void check_template(chunk_t *start)
...
+      {
+         tmp->next->type = CT_OC_PROPERTY_ATTR;
```
CDanU: nullptr check for tmp

src/tokenize_cleanup.cpp
Krizai: next should exist otherwise this method makes no sense, I'll put assert here

src/tokenize_cleanup.cpp
Krizai: paren open chunk should exist for calling this method. Will put assert

src/tokenize_cleanup.cpp
Krizai: Yes, will add.
Comments - never tried, but theoretically yes.

src/tokenize_cleanup.cpp
Krizai: paren open chunk should exist for calling this method. Will put assert

src/tokenize_cleanup.cpp
CDanU: Ok, see if `chunk_get_next_ncnl()` is a better fit for such cases

Krizai: Good one, thank you, will think about

Krizai:
@CDanU  Updated

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1069,86 @@ static void check_template(chunk_t *start)
            (end != NULL) ? get_token_name(end->type) : "<null>");
    set_chunk_type(start, CT_COMPARE);
 } // check_template
+
+
+void cleanup_objc_property(chunk_t *start)
+{
+   chunk_t *open_paren = chunk_get_next_type(start, CT_PAREN_OPEN, start->level);
```
CDanU: This will need a nullptr check for start.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1069,86 @@ static void check_template(chunk_t *start)
...
+   mark_selectors_in_property_with_open_paren(open_paren);
+   mark_attributes_in_property_with_open_paren(open_paren);
+}
+
+
+void mark_selectors_in_property_with_open_paren(chunk_t *open_paren)
+{
+   chunk_t *tmp = open_paren;
+
+   assert(open_paren->type == CT_PAREN_OPEN);
```
CDanU: You made sure that `open_paren` is not a `nullptr` in `cleanup_objc_property()`, but this function could be used somewhere else in the future where that is not going to be guaranteed anymore. A `nullptr` check for `open_paren` is needed here.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1069,86 @@ static void check_template(chunk_t *start)
...
+      }
+   }
+}
+
+
+void mark_attributes_in_property_with_open_paren(chunk_t *open_paren)
+{
+   chunk_t *tmp = open_paren;
+
+   assert(open_paren->type == CT_PAREN_OPEN);
```
CDanU: You made sure that `open_paren` is not a `nullptr` in `cleanup_objc_property()`, but this function could be used somewhere else in the future where that is not guaranteed anymore. A `nullptr` check for `open_paren` is needed here.

CDanU:
Ok so far the code looks ok. Could you squash your commits into one commit?
I need to check if those asserts are not causing too much mess in there. This could potentially lead to problems if `--no-backup` is used.

@mihaipopescu could you also have a look on the changes?

Krizai:
Asserts are only to detect wrong usage of the methods. They will never be executed with the current code.

scripts/More_Options_to_Test/help.txt
```diff
@@ -70,6 +70,6 @@ Note: Use comments containing ' *INDENT-OFF*' and ' *INDENT-ON*' to disable
       processing of parts of the source file (these can be overridden with 
       enable_processing_cmt and disable_processing_cmt).
 
-There are currently 581 options and minimal documentation.
+There are currently 582 options and minimal documentation.
```
mihaipopescu: Not related to this PR but generated files should never be under revision control.

src/combine.cpp
```diff
@@ -5847,63 +5848,28 @@ static void handle_oc_property_decl(chunk_t *os)
          // now.
          while (next != nullptr && next->type != CT_PAREN_CLOSE)
          {
-            if (next->type == CT_WORD)
+            if (next->type == CT_OC_PROPERTY_ATTR)
```
mihaipopescu: Is guarded for OC only ? Can't really tell on my phone.

src/combine.cpp
```diff
@@ -5847,63 +5848,28 @@ static void handle_oc_property_decl(chunk_t *os)
...
             {
-               if (chunk_is_str(next, "atomic", 6))
+               if (chunk_is_str(next, "atomic", 6) ||
+                   chunk_is_str(next, "nonatomic", 9))
```
mihaipopescu: Not related to this PR but why do we have to pass strlen all the time instead of having a template function ?

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1069,88 @@ static void check_template(chunk_t *start)
...
+void mark_selectors_in_property_with_open_paren(chunk_t *open_paren)
+{
+   assert(open_paren && open_paren->type == CT_PAREN_OPEN);
+
+   chunk_t *tmp = open_paren;
+
+   while (tmp && tmp->type != CT_PAREN_CLOSE)
+   {
+      if (tmp->type == CT_WORD &&
+          (memcmp(tmp->str.c_str(), "setter", 6) == 0 ||
```
mihaipopescu: I think we have other string comparison functions.

src/tokenize_cleanup.cpp
```diff
@@ -1068,3 +1069,88 @@ static void check_template(chunk_t *start)
...
+   assert(open_paren && open_paren->type == CT_PAREN_OPEN);
+
+   chunk_t *tmp = open_paren;
+
+   while (tmp && tmp->type != CT_PAREN_CLOSE)
+   {
+      if ((tmp->type == CT_COMMA || tmp->type == CT_PAREN_OPEN) &&
+          tmp->next && tmp->next->type == CT_WORD)
+      {
+         tmp->next->type = CT_OC_PROPERTY_ATTR;
```
mihaipopescu: You don't want to break the loop here ?

src/combine.cpp
Krizai: Yes. It's part of the method handle_oc_property_decl, which will be called only for CT_OC_PROPERTY

src/tokenize_cleanup.cpp
Krizai: No, I want to do it for every attribute ( there can be several)

src/tokenize_cleanup.cpp
Krizai: true. Will change

Krizai:
Incorporated all the comments, except for the once unrelated to this PR.

src/combine.cpp
```diff
@@ -5940,22 +5906,27 @@ static void handle_oc_property_decl(chunk_t *os)
                   }
                   setter_chunks.push_back(chunkGroup);
                }
-               else if (chunk_is_str(next, "nullable", 8))
+               else if (chunk_is_str(next, "nullable", 8) ||
+                        chunk_is_str(next, "nonnull", 7) ||
+                        chunk_is_str(next, "null_resettable", 15) ||
+                        chunk_is_str(next, "null_unspecified", 16)
+                        )
```
CDanU: @gmaurel do we need a new rule in forUncrustifySources.cfg? I think that this paren should not be on a new line.

gmaurel: Yes, we need!
I found some more:
```
grep -n "   )$" *
combine.cpp:2995:                        )
indent.cpp:1526:               )
newlines.cpp:2168:                )
```

Issue #1137, opened 2017-05-05
make_version.py doesn't work properly


guy-maurel:
I want to make a new release.
```
git archive -o uncrustify-0.nn.tar.gz e1181d3b3a78ef9fe0f52af53678fdf3340233c5
ls -l uncrustify-0.nn.tar.gz 
-rw-r--r-- 1 guy users 1373927 Mai  5 18:27 uncrustify-0.nn.tar.gz
mkdir 0.65-test
cd 0.65-test
cp ../uncrustify/uncrustify-0.nn.tar.gz .
gunzip uncrustify-0.nn.tar.gz
tar xvf uncrustify-0.nn.tar
mkdir build
cd build
cmake ..
< snip>
-- Found PythonInterp: /usr/bin/python (found version "2.7.6") 
-- 
CMake Warning at CMakeLists.txt:137 (message):
  64: Unknown version control system.

  , scripts/make_version.py doesn't work properly

```
**Can anybody help?**

CDanU:
This is supp_r a_sed to happen, the `make_version.py` is checking if a .git dir is existing to use git to extract the current version info. It bails with this warning because with the posted steps the `.git` is missing. CMake will use the fallback version string in this case.

CDanU:
Could you postpone the new release until #945 is fixed?

guy-maurel:
Well. The message is not complete: something is missing before the comma.
If it is "... supposed to happen...", I prefer a better message, which explains what happens.
I have not understood I could make the job.
I can wait a little, but I wish to push 0.65 in May.

mihaipopescu:
It's been a while but I can check.
By the way, how are you running this script? I thought make_version is part of CI as well.

guy-maurel:
As shown above, the script was run from the "make" command.

mihaipopescu:
Made a PR to improve messaging but yes, as @CDanU said, you that was a warning and can be ignored.

mihaipopescu:
And as said in #767, you can create a new milestone for 0.65 so we can select what issues we would like have fixed for this milestone. What do you think ?

guy-maurel:
#1138 shows a better (warning) message.
I think this is not a warning but an info. Let us use:
```message(STATUS, ... ```

guy-maurel:
0.64 is written in the CMakeLists.txt file.
Is it possible to generate this with xyz ?

CDanU:
Do you mean automatic generation? 
[...]

PR #1138, opened 2017-05-06
Fix failure of retrieving git hash from a hg-git enabled repository

mihaipopescu:
It also improves a bit the messaging when there is a failure retrieving the hash from a VCS which can be at times confusing #1137.

This re-integrates 3ef03cd6ac3fe218e77f14336e7c17daf75ba551 back.

### DIFF
--- a/scripts/make_version.py
+++ b/scripts/make_version.py
@@ -32,7 +32,7 @@ def main(args):
 
     if exists(git_path):
         try:
-            proc = Popen(['git', 'describe', '--always', '--dirty'], stdout=PIPE, stderr=PIPE)
+            proc = Popen(['git', 'describe', '--always', '--dirty'], stdout=PIPE, stderr=PIPE, cwd=root)
             txt_b, error_txt_b = proc.communicate()
             txt = txt_b.decode("UTF-8").strip().lower()
             error_txt = "%d: %s" % (proc.returncode, error_txt_b.decode("UTF-8").strip().lower())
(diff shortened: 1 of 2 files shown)


CDanU:
Uff, I just wanted to started reviewing this.

scripts/make_version.py
```diff
@@ -42,20 +42,20 @@ def main(args):
...
         try:
             check_call(['hg', 'gexport'])
-            proc0 = Popen(['hg', '--config', 'defaults.log=', 'log', '-r', '.', '--template', '{gitnode}'], stdout=PIPE, stderr=PIPE)
+            proc0 = Popen(['hg', '--config', 'defaults.log=', 'log', '-r', '.', '--template', '{gitnode}'], stdout=PIPE, stderr=PIPE, cwd=root)
             node_b, error_txt_b = proc0.communicate()
             node = node_b.decode("UTF-8")
             error_txt = "%d: %s" % (proc0.returncode, error_txt_b.decode("UTF-8").strip().lower())
 
-            proc1 = Popen(['git', '--git-dir=.hg/git', 'describe', '--tags', '--always','--dirty', node], stdout=PIPE, stderr=PIPE)
+            proc1 = Popen(['git', '--git-dir=.hg/git', 'describe', '--long', '--tags','--always', node], stdout=PIPE, stderr=PIPE, cwd=root)
```
CDanU: Why are those changed? If I remember correctly I removed the long to prevent a hash string on a tagged commit (the actual release). The -dirty flag should not be removed as it is useful information when we receive bug reports. It could be that the uncommitted _dirty_ changes are causing bugs.

mihaipopescu: It's a special setup as I'm not under git but it's an hg repository with a git remote using hggit extension.

I'll try without --long but -dirty doesn't work in my case. It fails with a strange error (iirc something about comittish, was funny).

mihaipopescu: I think this is the closest I ended up matching the git version but indeed, maybe some things have changed. Sorry not taking this into account.

Issue #1002, opened 2017-02-07
Lose space before && in template


jlee975:
But not in the definition of value

```
template< class B1 = void, class B2 = void >
struct conjunction : foo< B1::value && B2::value >
{
	static const bool value = B1::value && B2::value;
};
```

See conditional.cpp, conditional.cpp.uncrustify, conditional.debug in https://github.com/jlee975/uncrustify-tests

Using uncrustify v0.64_d 1f538888ad9b3adc8a46828ebe2ad6578629df29

CDanU:
This is caused by `sp_before_byref = remove`. 

I am not entirely sure what the `&&` in
```cpp
struct conjunction : foo< B1::value && B2::value >
```
are supposed to mean.

jlee975:
It's boolean AND, just as if it were in (true && false).

guy-maurel:
Could please push a complete example. I would like to compile and run it.

jlee975:
I"ve updated conditional.cpp, etc in https://github.com/jlee975/uncrustify-tests

It is a simple implementation of std::conjunction from C++17

guy-maurel:
We don't support C++17 now.
But, if you want to join us, we will be very happy.

jlee975:
Sorry, that's not what I meant. The code will compile under C++03.

I maintain a project (cppbackport) that implements C++11/14/17 stuff in C++03, so that's ultimately where it came from. I was just mentioning what the code does, if you were going to run it.

guy-maurel:
http://en.cppreference.com/w/cpp/types/conjunction shows:
 _template<class... B>
struct conjunction;
	(1) 	(since C++17)_
We take it in the TODO list

jlee975:
Sorry, I think the use of conjunction has confused things. The issue doesn't really have anything to do with std::conjunction except that I chose to use it for the example.

The issue is that for

```
struct conjunction : foo< B1::value && B2::value >
```

the && is formatted with sp_before_byref/sp_after_byref when it should be sp_arith

guy-maurel:
A proposal is PR#1141

gmaurel:
is fixed

jlee975:
Thanks for your help with this

PR #1141, opened 2017-05-08
A bugfix for #1002

guy-maurel:
Test if not in template

### DIFF
--- a/tests/config/bug_1002.cfg
+++ b/tests/config/bug_1002.cfg
@@ -1,6 +1,1 @@
-input_tab_size                  = 4
-output_tab_size                 = 4
-indent_columns                  = 4
-sp_arith                        = force
-sp_assign                       = force
 sp_bool                         = force
--- a/tests/config/bug_1002.cfg
+++ b/tests/config/bug_1002.cfg
@@ -1,1 +1,3 @@
-sp_bool                         = force
+sp_bool         = force
+sp_before_byref = remove
+sp_after_byref  = remove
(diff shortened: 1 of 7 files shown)


tests/config/bug_1002.cfg
```diff
@@ -0,0 +1,6 @@
+input_tab_size                  = 4
+output_tab_size                 = 4
+indent_columns                  = 4
+sp_arith                        = force
+sp_assign                       = force
+sp_bool                         = force
```
CDanU: Are all of those needed? 
As far as I understand the problem only `sp_before_byref`/`sp_after_byref` and `sp_arith` should be involved. 
The option `sp_arith` should be activated while both of the other options should not.

tests/input/cpp/bug_1002.cpp
```diff
@@ -0,0 +1,4 @@
+template< class B1 = void, class B2 = void >
+struct conjunction : bool_constant<B1::value1 && B2::value2>
```
CDanU: I just noticed this, you are not testing if  `sp_bool = force` is actually doing anything here.
You could do this:

```cpp
struct conjunction : bool_constant<B1::value1  &&  B2::value2>
struct conjunction : bool_constant<B1::value1&&B2::value2>
```
To be sure there should also be a test with remove. The other two mentioned options should also be included to make sure that they are not causing misbehavior.

guy-maurel: Thanks

PR #1148, opened 2017-05-11
Extract the fallback version string from the package.json file

CDanU:


CDanU:
ref: #1137

gmaurel:
But we need to fill up the package.json file
Is it possible to prevent this?

CDanU:
The only thing I could do is to write a script that reads the version info and than bumps that up by one.
That could be applied to the package.json file and to the current feadback string inside CMakeLists.txt.
Such a script cannot be included in the normal cmake build process. I could however introduce an option for that ... somehing like `cmake __-Dbump-version__ -DCMAKE_BUILD_TYPE=Release ../` which would internaly call the script than.

Generating version number strings out of thin air is not a thing that I can do, so either `package.json` or `CMakeLists.txt` need to have info that will be used as reference for the version bump.

CMakeLists.txt
```diff
@@ -122,7 +122,22 @@ endif()
...
-set( FALLBACK_VERSION "Uncrustify-0.64_f" )
+
+# get a fallback version string from the package.json file to guard against
+# failure of the make_version.py script
+set(FALLBACK_VERSION "")
+file(READ "package.json" content)
+string(REPLACE "\n" ";" lines "${content}")
+foreach(line ${lines})
+    string(STRIP "${line}" line)
+    if("${line}" MATCHES "^\"version\" *: *\"([0-9.]+)\"")
```
mihaipopescu: How can this work as the line starts with two spaces?
I would also replace whitespaces with \s character just to be sure (if supported).
And why not take any (.*) version string and enforce a specific format ?

CMakeLists.txt
```diff
@@ -122,7 +122,22 @@ endif()
 if( NOT CMAKE_BUILD_TYPE )
   set( CMAKE_BUILD_TYPE Release )
 endif()
-set( FALLBACK_VERSION "Uncrustify-0.64_f" )
+
+# get a fallback version string from the package.json file to guard against
+# failure of the make_version.py script
+set(FALLBACK_VERSION "")
+file(READ "package.json" content)
+string(REPLACE "\n" ";" lines "${content}")
```
mihaipopescu: Why splitting and not searching the entire thing ?

CMakeLists.txt
CDanU: The line is stripped at 132 so there will be no white spaces. 
Unfortunately CMake can only deal with simple regex, so no character classes.

The last sentence I don't understand completely. I am enforcing (fairly loosely) that only digits and dots can be used for the version. Inside `make_version.py` I am more strictly but as this here is used as fallback it should be alright.

CMakeLists.txt
CDanU: I don`t want to let the regex run on the whole file.

mihaipopescu: I can see that, but I think that's premature optimisation.

CMakeLists.txt
mihaipopescu: Right, I see the strip now, cool. 👍

Issue #1139, opened 2017-05-08
closing parenthese and NL


guy-maurel:
As discuss at PR#1137 we need a new option to cover the newline after the closing parenthesis for if.
Look at line 9 here:
```
void a()
{
int a;
               if ((tmp == nullptr) ||
                        ((tmp->type != CT_NUMBER) &&
                         (tmp->type != CT_SIZEOF) &&
                         !(tmp->flags & (PCF_IN_STRUCT | PCF_IN_CLASS))) ||
                        (tmp->type == CT_NEWLINE)
                        ) // closing parenthesis
               {
                  set_chunk_type(next, CT_LABEL_COLON);
               }
}
```

guy-maurel:
Fix is coming soon.

guy-maurel:
Is fixed

PR #1149, opened 2017-05-11
a bugfix for #1139

guy-maurel:


src/newlines.cpp
```diff
@@ -2612,6 +2612,16 @@ void newlines_cleanup_braces(bool first)
...
+         tmp = chunk_get_next_type(pc, CT_SPAREN_CLOSE, pc->level);
+         if (tmp != nullptr)
+         {
+            prev = chunk_get_prev(tmp);
+            if (prev != nullptr)
+            {
+               // Issue #1139
+               newline_iarf_pair(prev, tmp, cpd.settings[UO_nl_before_closing_paren].a);
+            }
+         }
```
CDanU: As far as I see the `nl_before_closing_paren` option is only working on CT_IF chunks? Unless you plan to extend this I'd rename that option to `nl_before_if_closing_paren`.

guy-maurel: no because I use it also with `else if`

CDanU: Yes but `nl_before_closing_paren` would imply that all closing parenthesis, including for example `for(;;)`, would be affected.

guy-maurel: yes... next day

tests/input/cpp/bug_1139.cpp
```diff
@@ -0,0 +1,31 @@
...
+                        )
+               {
+                  set_chunk_type(next, CT_LABEL_COLON);
+               }
+else if ((tmp == nullptr) ||
+                        ((tmp->type != CT_NUMBER) &&
+                         (tmp->type != CT_SIZEOF) &&
+                         !(tmp->flags & (PCF_IN_STRUCT | PCF_IN_CLASS))) ||
+                        (tmp->type == CT_NEWLINE)
+                        )
```
CDanU: You are not testing if force works correctly. Pleas make one branch with 
```cpp
... (tmp->type == CT_NEWLINE))
```
and the other with
```cpp
... (tmp->type == CT_NEWLINE)

)
```

tests/config/bug_1139-f.cfg
```diff
@@ -0,0 +1,6 @@
+output_tab_size            = 3
+indent_columns             = 3
+indent_with_tabs           = 0
+sp_before_sparen           = force
```
CDanU: All of those are not needed to test `nl_before_if_closing_paren`

PR #962, opened 2017-01-26
Less warning for indent

gmaurel:
Part 2

PR #1151, opened 2017-05-12
Fix indentation after colon (C++ base class list or constructor initialization list)

yirkha:
Hi, this PR resolves two backward incompatibilities I noticed while trying out the latest version of uncrustify:

1. Seems like `indent_ctor_init` and `indent_ctor_init_leading` have been affecting not only indentation of member initialization list in constructors, but also indentation of base class list in class declarations - yet only in some cases. Another modification of this misbehavior in #654 (Indent constructor members) broke my config.
I believe it is all a mistake and these settings should affect only constructors because of their name, so I changed the logic to reflect that. If there is a need to have the same fine-grained control also for base class list, there should be new options for that.

2. `indent_ctor_init` used to be able to take negative values since #346 (allow indent_ctor_init to be negative). This feature has been accidentally removed during warning fixes in #962 (Less warning for indent). I have restored this ability in a more safe manner, checking for underflow.

I have also done some small cleanup for clarity and added/modified test cases to make sure these things do not get broken again.

Thanks for your consideration.


tests/input/cpp/indent_ctor_init.cpp
```diff
@@ -1,7 +1,12 @@
-struct MyClass {
+struct MyClass : public Foo,
+                 private Bar {
   MyClass(int a, int b, int c)
 	 : m_a(a), m_b(b),
    m_c(c) {}
 private :
   int m_a, m_b, m_c;
 };
```
gmaurel: It is not pretty to declare more than one variable in a line.
Please split in many lines.

src/indent.cpp
```diff
@@ -1560,56 +1560,51 @@ void indent_text(void)
          frm.pse[frm.pse_tos].indent_tmp = frm.pse[frm.pse_tos].indent;
          frm.pse[frm.pse_tos].indent_tab = frm.pse[frm.pse_tos].indent;
          log_indent_tmp();
-
          indent_column_set(frm.pse[frm.pse_tos].indent_tmp);
 
-         if ((cpd.settings[UO_indent_class_colon].b && (pc->type == CT_CLASS_COLON)) ||
-             (cpd.settings[UO_indent_constr_colon].b && (pc->type == CT_CONSTR_COLON)))
+         prev = chunk_get_prev(pc);
+         next = chunk_get_next(pc);
```
gmaurel: please check of != nullptr

tests/input/cpp/indent_ctor_init.cpp
yirkha: You mean change the configuration for this test to split them or change the test input file itself?
I thought one should not change existing test cases, especially if unrelated to the rest of the commit.

src/indent.cpp
yirkha: Could you please expand on that?

`prev` is used only at line 1583 in `chunk_is_newline(prev)` which is safe, because it checks for null pointer inside. I could add an explicit check here also if you'd prefer so, but seems pointless.

`next` is used two times below, but both times after `if ((next != nullptr) && ...)` already.

tests/input/cpp/indent_ctor_init.cpp
gmaurel: OK.

src/indent.cpp
gmaurel: As prev is only used at line 1583 put the line 1565 just before.
Same case for next. Put it just before you are using it.
One can better see what happens.
Thanks.

yirkha: Oh, right, I see, I could simplify that further.
Adds one level of indentation, but makes the logic even easier to follow.
Thanks for the review!

tests/config/indent_ctor_init.cfg
```diff
@@ -1,1289 +1,18 @@
-# Uncrustify 0.58
```
mihaipopescu: Did you "reduced" the config only to the options with non-default values?

src/indent.cpp
```diff
@@ -1560,56 +1561,57 @@ void indent_text(void)
...
-               }
+            // TODO: Create a dedicated indent_constr_on_colon?
+            if (cpd.settings[UO_indent_class_on_colon].b)
+            {
+               frm.pse[frm.pse_tos].indent = pc->column;
+               log_indent();
+            }
+            else if (cpd.settings[UO_indent_ctor_init].n != 0)
+            {
+               frm.pse[frm.pse_tos].indent = std::max<ptrdiff_t>(frm.pse[frm.pse_tos].indent + cpd.settings[UO_indent_ctor_init].n, 0);
```
mihaipopescu: why `ptrdiff_t` ?

src/options.cpp
```diff
@@ -747,7 +747,7 @@ void register_options(void)
                   "Whether to indent the stuff after a leading class initializer colon");
    unc_add_option("indent_ctor_init_leading", UO_indent_ctor_init_leading, AT_UNUM,
                   "Virtual indent from the ':' for member initializers. Default=2");
-   unc_add_option("indent_ctor_init", UO_indent_ctor_init, AT_UNUM,
+   unc_add_option("indent_ctor_init", UO_indent_ctor_init, AT_NUM,
                   "Additional indenting for constructor initializer list");
```
mihaipopescu: There should be an indication in the description message that now it accepts negative numbers and their meaning.

yirkha: Good point, added.

tests/config/indent_ctor_init.cfg
yirkha: I tried to diff the config with defaults for uncrustify 0.58, but there were way too many changes - it's someone's complete config with all modifications. So I kept only the options relevant to what the test case seemed to check (just one, 30955, actually references this config).

Seems pointless to keep 50 KB configs around. Like there are just minimal code snippets, there should be just minimal configs. It's much easier to see what's going on and also to create derivative configs, like I subsequently did with `indent_ctor_init_leading.cfg` here.

But all that's just IMO, if you guys prefer to keep it in its entirety, I will just remove that commit :)

src/indent.cpp
yirkha: If the `max()` function was specialized with `size_t` (the type of `frm.pse[frm.pse_tos].indent` and of the add expression), it would be effectively a no-op, because `size_t` is unsigned and even a "negative" value would be always >= 0.

In order for the safety comparison to work and yet not go from possibly 64-bit `size_t` to 32-bit `int` and back and possibly trigger some noisy truncation warnings, I have used the "signed integer type of the same size as `size_t`", which is `ptrdiff_t`.

tests/config/indent_ctor_init.cfg
CDanU: No, that is actually quiet nice. I am slowly trying to reduce the config/test bloat as well. I wrote a somewhat faulty script for such things [here](https://gist.github.com/CDanU/54446524c8919b1d490147f6c8165bb6#file-option_reducer-py). I'll run that on the config and see if I get the same results.

yirkha: I see, that's a great systematic way to do it! Yeah, the only differences I see are

- `indent_with_tabs`, which I just add always for good measure, even though it does not make a difference here, and

- my `nl_collapse_empty_body` vs. your `nl_class_leave_one_liners`, and there's probably no harm in leaving my stronger option in.

Thanks for the verification.

src/indent.cpp
mihaipopescu: I understand now. It would be better to have this comment inside the code because it's not really obvious.

yirkha: All right, updated. Thanks!

PR #1156, opened 2017-05-14
Update emscripten interface

CDanU:


mihaipopescu:
Cool. Code gen or manual ?

CDanU:
Semi automatic: [genEmscriptenEnumBindingCode.sh](https://github.com/uncrustify/uncrustify/blob/master/scripts/genEmscriptenEnumBindingCode.sh). It uses clang in a super hackish way that breaks when the `CompileCommands.json` is available in the `src` dir. I have to write a python script for that someday.

Issue #1108, opened 2017-04-10
indent_with_tabs = 0 produce neverending indentation


ksanderon:
**version**: uncrustify 0.64
**tested on**: mac os x, debian linux
**config:** 
```
indent_with_tabs = 0
input_tab_size = 3
output_tab_size = 3
indent_columns = 3
indent_braces	 = TRUE
nl_assign_brace = add
nl_enum_brace = add
nl_union_brace= add
nl_struct_brace= add
nl_do_brace= add
nl_if_brace= add
nl_for_brace= add
nl_else_brace= add
nl_while_brace= add
nl_switch_brace= add
nl_fcall_brace= add
nl_fdef_brace= add
nl_brace_while= add
nl_brace_else= add
```

**source:**
```
int foo()
   {
   const std::map<std::string, int> bar =
      {
          { "-13", -13 },
          { "0", 0 },
          { "0x13", 0x13 },
          { "1", 1 },
          { "4294967297", 2147483648 * 2 + 1 }
      };
   return 5;
   }
```

**result first time:**
```
int foo()
   {
   const std::map<std::string, int> bar =
      {
              { "-13", -13 },
              { "0", 0 },
              { "0x13", 0x13 },
              { "1", 1 },
              { "4294967297", 2147483648 * 2 + 1 }
      };
   return 5;
   }

```

**result N time:**
```
int foo()
   {
   const std::map<std::string, int> bar =
      {
                                                                                                                                                              { "-13", -13 },
                                                                                                                                                              { "0", 0 },
                                                                                                                                                              { "0x13", 0x13 },
                                                                                                                                                              { "1", 1 },
                                                                                                                                                              { "4294967297", 2147483648 * 2 + 1 }
      };
   return 5;
   }
```

So as you see in case when `indent_with_tabs = 0`, uncrustify behaviour is totally incorrect, do you have idea why? Maybe needed to set some additional parameter to disable this?

CDanU:
I can confirm this on the current master a6cd6afc7fb82f07d91fb2d55069aa02e0423454.
Basically this looks like that one problem where you have I1 ->O1 = I2 -> O2 = I3 -> ....
We try to mitigate those issues but its not that easy to find and fix such things.

Testing this for a short amount of time it seems that `indent_braces	 = TRUE` in combination with some of the other options is causing this.

guy-maurel:
I am working on it

guy-maurel:
A proposal to fix the bug is #1163

[...]

PR #1163, opened 2017-05-16
a bugfix for #1108

guy-maurel:


tests/config/bug_1108.cfg
```diff
@@ -0,0 +1,6 @@
+indent_with_tabs = 0
+input_tab_size = 3
+output_tab_size = 3
+indent_columns = 3
+#indent_braces	 = false
```
CDanU: Unneeded options, only indent_braces matters here.

use:
```cpp
int foo()
{
	const std::map<std::string, int> bar =
	{
		{ "abcXYZ", -13 },
	};
	return 5;
}
``` 
as input/output.

CDanU:
It is not working because you forgot to change the input/output files.

PR #1175, opened 2017-05-19
Don't force CMAKE_BUILD_TYPE

AMDmi3:
If build type is not set, it should not be forced. For example, it should remain unset for FreeBSD port which is required to respect systemwide C/C++ compiler flags passed via CMAKE_C/CXX_FLAGS, and adding extra flags (which setting build type does) is allowed.


CDanU:
Superseded by #1179

Issue #1177, opened 2017-05-19
Version generation is broken


AMDmi3:
JFYI, for FreeBSD port I've had to remove whole version generation stuff, it would be nice to have it removed upstream as well (or at least a way to disable it), because it's terminally broken.

- it adds unneeded dependency on python
- it adds unneeded dependency on git
- it's unreliable as it may pick a tag from completely unrelated repository (if uncrustify source directory doesn't have .git, but some parent directory, say ~, does)
- it hinders reproducible builds

CDanU:
While I agree that forcing to have CMAKE_BUILD_TYPE defined can be problematic and that the script currently is is not checking for the right .git directory is a problem I would not say that the generation is downright broken. 
The version string generation was intended to be used if python and git are available, if not the fallback string should be used. I forgot to remove the `find_package(PythonInterp REQUIRED)` stuff.

Would you still need an extra option to remove the version string generation if the problems I mentioned are fixed? I personally don't yet see how it would than hinder reproducible builds.

AMDmi3:
Unfortunately, yes. I absolutely don't want to add python and git dependencies for no reason. If they are not added, build behavior will be inconsistent as it'll depend on whether git and python are actually available.

CDanU:
I don't really see why you want to be so stringent on something that is only altering the version string, but I will make an option so that the fallback string will be used.

AMDmi3:
> I don't really see why you want to be so stringent

I do not want anything beyond the most basic packaging requirements. I've already explained: package builds are required to be predictable, consistent and reproducible. Besides other things, they are only allowed to use depends which are explicitly defined in the package and should build the same way on a clean system where only these depends are available and on a real system where lots of third party packages are installed. Any optional dependency makes a build impossible to test, and each optional dependency makes a build an order of magnitude more fragile.

See also https://reproducible-builds.org/

> on something that is only altering the version string

Don't let yourself be deluded. For "only altering the version string" you're bringing in python (which can be of two incompatible major branches), 100+ lines of non-obvious python code using 6 modules, huge regexps, subprocess calling, complex system interaction and two third party utilities with their own massive complexity, list of depends, compatibility issues and lots of possible versions.

> but I will make an option so that the fallback string will be used even if python and git are around

Fine with me, thank you!

> The main reason for why we use info that is provided by git

[...]

PR #1179, opened 2017-05-21
Make git version string generation optional

CDanU:
use `cmake -DNoGitVersion=True ...` to disable it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,29 +117,36 @@ endif()
 # Generate uncrustify_version.h
 #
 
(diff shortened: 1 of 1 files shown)


CDanU:
ref: #1177

guy-maurel:
Could it be wise to use `CMAKE_SYSTEM_NAME` ?

CDanU:
I could include that, but personally I don't see the point as we don't use OS related functions much and those that we do use, worked without a flaw up to this point.

guy-maurel:
Which one is the one we need: #1175 or #1179 ?

CDanU:
This, the other one can be closed.

PR #1190, opened 2017-05-29
Extend README.md

CDanU:


CDanU:
@mihaipopescu could you check that I linked the correct unity objectiv-c branch?

mihaipopescu:
Yes and it's a very good point. @aet made those changes in a separate branch we have failed to find the time to integrate them.

Issue #1111, opened 2017-04-11
AT_NUM size_t underflow


CDanU:
After the somewhat recent transition from int to size_t some AT_NUM options are buggy because they underflow the size_t type if they are set to negative values.

~- [x] `reindent_line()`~
~- [x] `UO_pp_indent_region`~
~- [x] `UO_pp_indent_if`~
~- [x] `UO_indent_var_def_blk`~
~- [x] `UO_indent_case_brace`~
~- [x] `UO_indent_label`~
~- [x] `UO_indent_access_spec`~
~- [x] `UO_align_right_cmt_gap`~
- [ ] `UO_cmt_sp_after_star_cont` ... did that one ever work? We have no tests for it, it depends on `UO_cmt_width` or (`UO_cmt_indent_multi` and `UO_cmt_sp_before_star_cont`). That is a strange combination.

The progress on this will be here: https://github.com/CDanU/uncrustify/commits/fix-underflow

----------

While fixing some of the overflows I discovered that some options still behave strangely with too negative options, take for example `indent_var_def_blk = -21`:

Input:
```cpp
struct s
{
    void f(int i)
    {
        int y = 2;
    }
}
```
Output:
```cpp
struct s
{
voidf(int i)
	{
inty = 2;
	}
}
```
<details><summary>click to expand Debug</summary><pre>
# Uncrustify-0.64-581-6fe6fc65
indent_var_def_blk              = -21
# option(s) with 'not default' value: 1
#
# -=====-
# Line              Tag           Parent          Columns Br/Lvl/pp     Flag   Nl  Text
#   1>           STRUCT[            NONE][  1/  1/  7/  0][0/0/0][     70002][0-0] struct
#   1>             TYPE[          STRUCT][  8/  8/  9/  1][0/0/0][         2][0-0]        s
#   1>          NEWLINE[            NONE][  9/  9/  1/  0][0/0/0][         2][1-0]
#   2>       BRACE_OPEN[          STRUCT][  1/  1/  2/  0][0/0/0][ 100000002][0-0] {
#   2>          NEWLINE[            NONE][  2/  2/  5/  0][1/1/0][         2][1-0]
#   3>             TYPE[        FUNC_DEF][  0/  5/  9/  0][1/1/0][    470402][0-0]void
#   3>         FUNC_DEF[            NONE][  5/ 10/ 11/  1][1/1/0][   1800402][0-0]     f
[...]

PR #1191, opened 2017-05-29
Fix underflow

CDanU:
ref: #1111

### DIFF
--- a/scripts/More_Options_to_Test/show_config.txt
+++ b/scripts/More_Options_to_Test/show_config.txt
@@ -1727,7 +1727,7 @@ align_oc_decl_colon             { False, True }
 cmt_width                       Unsigned Number
   Try to wrap comments at cmt_width columns
 
-cmt_reflow_mode                 Number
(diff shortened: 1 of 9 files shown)


stefan-nu:
Hi @CDanU,

in File indent line 312 you have changed
`size_t col_delta = column - pc->column;`
to
`auto col_delta = static_cast<int>(column) - static_cast<int>(pc->column);`

Shouldn't the same code on line 233 be changed as well?
`size_t col_delta = column - pc->column;`

CDanU:
Yes, good catch, I will make a PR.

stefan-nu:
You have changed the option "align_right_cmt_gap" From `AT_NUM` to `AT_UNUM`.

Shouldn't the configuration files be changed accordingly?
In test/config/ger.cfg line 192: 
`align_right_cmt_gap                      = -5`
Should probably become 
`align_right_cmt_gap                      = 0`

guy-maurel:
The option is no more present.

PR #1194, opened 2017-06-02
Some debug, better debug lines

guy-maurel:


tests/cpp.test
```diff
@@ -436,3 +436,5 @@
 
 34170  empty.cfg                       cpp/i1082.cpp
 34171  i1181.cfg                       cpp/i1181.cpp
+
+34190 bug_1003.cfg                     cpp/bug_1003.cpp
```
CDanU: You have this and a fparen check included from the other PR. The actual tests are missing so the build fails.

PR #1198, opened 2017-06-05
fix warning of printing uint64_t

codable:
This PR fixes the warning for printing uint64_t on 32bit system.

### DIFF
--- a/src/unc_tools.cpp
+++ b/src/unc_tools.cpp
@@ -160,7 +160,7 @@ void dump_out(unsigned int type)
          fprintf(D_file, "  orig_col %zu\n", pc->orig_col);
          fprintf(D_file, "  orig_col_end %zu\n", pc->orig_col_end);
          fprintf(D_file, (pc->orig_prev_sp != 0) ? "  orig_prev_sp %u\n" : "", pc->orig_prev_sp);
-         fprintf(D_file, (pc->flags != 0) ? "  flags %zu\n" : "", pc->flags);
+         fprintf(D_file, (pc->flags != 0) ? "  flags %" PRIu64 "\n" : "", pc->flags);
          fprintf(D_file, (pc->column != 0) ? "  column %zu\n" : "", pc->column);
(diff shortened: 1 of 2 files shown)


src/unc_tools.cpp
```diff
@@ -160,7 +160,7 @@ void dump_out(unsigned int type)
          fprintf(D_file, "  orig_col %zu\n", pc->orig_col);
          fprintf(D_file, "  orig_col_end %zu\n", pc->orig_col_end);
          fprintf(D_file, (pc->orig_prev_sp != 0) ? "  orig_prev_sp %u\n" : "", pc->orig_prev_sp);
-         fprintf(D_file, (pc->flags != 0) ? "  flags %zu\n" : "", pc->flags);
+         fprintf(D_file, (pc->flags != 0) ? "  flags %llu\n" : "", pc->flags);
```
CDanU: The `llu` is for unsigned `long long ints`, I do not where this could cause a problem but its not guaranteed that it will be equally sized as `uint64_t`. I would recommend the use of the more fitting C99 `PRIu64`. 
The problem with `PRIu64` on the other hand would be that Microsoft added support for that only with [VS2013}(https://blogs.msdn.microsoft.com/vcblog/2013/07/19/c99-library-support-in-visual-studio-2013/). 

@gmaurel what minimum VS version do we support?

CDanU: Btw. we already handle `PRIx64` in [src/windows_compat.h](https://github.com/uncrustify/uncrustify/blob/5bf42800b0117a1f0451d56fffb2e5defb0a1d02/src/windows_compat.h#L29-L31) so I guess we could do the same for `PRIu64` / `llu`.
@codable could you update your PR?

codable: I tried PRIu64 before, it fails on windows, then I changed to `llu` directly. It's awesome we have `windows_compat.h`. Updated.

codable:
any more comments?

PR #1202, opened 2017-06-10
Prevent underflows in align_to_column()

CDanU:


gmaurel:
@CDanU a conflikt to solve. Thanks

PR #1206, opened 2017-06-14
Hang on multi-line comments fixed.

seeraven:
Uncrustify hangs when indenting bad formatted multi-line comments like in
```.cpp
void function()
{
    printf( "Hello World\n" );
        /*
 //test
        /// Another comment
//end test
*/
}
```

This was caused by a subtraction without boundary check, so that the unsigned variable `ccol` actually got huge values. By ensuring the subtraction will not result in negative values this problem was fixed.

A test was added which shows the original problem.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -1492,7 +1492,12 @@ static void output_comment_multi(chunk_t *pc)
              * This is not the first line, so we need to indent to the
              * correct column. Each line is indented 0 or more spaces.
              */
-            ccol -= col_diff;
+            // Ensure ccol is not negative
+            if (static_cast<int>(ccol) >= col_diff)
+            {
+               ccol -= col_diff;
+            }
+
             if (ccol < (cmt_col + 3))
             {
                ccol = cmt_col + 3;
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -438,3 +438,5 @@
 34171  i1181.cfg                       cpp/i1181.cpp
 
 34190 bug_1003.cfg                     cpp/bug_1003.cpp
+
+34191 empty.cfg                        cpp/comment-align-multiline.cpp
--- a/tests/output/cpp/34191-comment-align-multiline.cpp
+++ b/tests/output/cpp/34191-comment-align-multiline.cpp
@@ -1,0 +1,11 @@
+#include <stdio.h>
+
+void function()
+{
+	printf( "Hello World\n" );
+	/*
+	   //test
+	   /// Another comment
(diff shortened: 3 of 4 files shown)


src/output.cpp
```diff
@@ -1492,7 +1492,12 @@ static void output_comment_multi(chunk_t *pc)
              * This is not the first line, so we need to indent to the
              * correct column. Each line is indented 0 or more spaces.
              */
-            ccol -= col_diff;
+            // Ensure ccol is not negative
+            if (int(ccol) >= col_diff)
```
CDanU: A static _cast would be better here.

tests/config/comment-align-multiline.cfg
```diff
@@ -0,0 +1,159 @@
...
+sp_func_def_paren=remove
+sp_inside_fparens=remove
+sp_inside_fparen=add
+sp_func_call_paren=remove
+sp_func_class_paren=remove
+sp_return_paren=add
+nl_start_of_file=remove
+nl_end_of_file=remove
+nl_struct_brace=add
+pos_arith=trail
```
CDanU: A config file is not needed for this testcase. Please use the `empty.cfg`

tests/cpp.test
```diff
@@ -438,3 +438,5 @@
 34171  i1181.cfg                       cpp/i1181.cpp
 
 34190 bug_1003.cfg                     cpp/bug_1003.cpp
+
+34191 comment-align-multiline.cfg      cpp/comment-align-multiline.cpp
```
CDanU:
```
34191 empty.cfg      cpp/comment-align-multiline.cpp
```

tests/output/cpp/34191-comment-align-multiline.cpp
```diff
@@ -0,0 +1,11 @@
...
+
+void function()
+{
+    printf( "Hello World\n" );
+    /*
+       //test
+       /// Another comment
+       //end test
+     */
+}
```
CDanU: Without the unnecessary changes (made by the config file) the output should be this:
```cpp
#include <stdio.h>

void function()
{
	printf( "Hello World\n" );
	/*
	   //test
	   /// Another comment
	   //end test
	 */
}

```
(Note that uncrustify defaults to tab indents)

guy-maurel:
You can prevent the failure by using ./scripts/Run_uncrustify_for_sources.sh

seeraven:
Good points! I've corrected the commit according to your suggestions.

PR #1209, opened 2017-06-16
Add options for preprocessor block indentations

brmqk3:
Adds options that control indenting case statements, extern
blocks, function definitions, and braces inside of
prepreprocessor blocks. indent_switch_pp controls indenting
preprocessor blocks inside switch statements. Also adds
tests for these options.

Change-Id: I730bac79371f7f3ae225aad049d2da0d194f2616


CDanU:
Don't be discouraged yet. If you are sure that you code is right check if the tests are possibly faulty.

While you do that I already would like to give you some notes.
- The option `indent_switch_pp` here is the most general one, it should be overwritten by the newly introduced options. The problem is that the new options default to `false` which makes `ndent_switch_pp`  useless. To fix this you either invert the logic of the new options (example: `pp_indent_brace` -> `pp_not_indent_brace`) or you set the defaults for that options to true in `options.cpp` `set_option_defaults()`.
- Please  add a test case for each new option. This makes it much more easier to see which options are tested and also allows people that are new to Uncrustify to see what a single option is doing.
- Also keep in mind that you are adding new options, this will change the output of some CLI options so you also have to adopt the changes for the CLI testcases (read [README.md:a-note-about-pull-requests](https://github.com/uncrustify/uncrustify#a-note-about-pull-requests))

brmqk3:
No problem! I'm currently making these changes, will update tomorrow.

Issue #1207, opened 2017-06-15
Unwanted deletion of semicolon after something like   return vector<int>{1};


deepseer:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastbin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

Uncrustify 0.6.5

Original file:
```
typedef vector<int> VI;
VI func() {
return vector<int>{1};
}
```
It is compiled in g++ 4.8.4 with -std=c++11

Uncrustified file:
```
typedef vector<int> VI;

VI func()
{
  return vector<int>{ 1 }
}
```

```
error: expected ‘;’ before ‘}’ token
 }
 ^
```

guy-maurel:
It works pretty fine for me with an empty configuration file.
1. What is "vector"?
2. Please push your configuration file.

deepseer:
It's a C++ source file. I only posted the relevant part.
`include`, `using namespace std` and `main()` were omitted.
Source file and configuration file have been uploaded to my repository: https://github.com/deepseer/deepseer_repository

PR #1210, opened 2017-06-17
Semicolon removal fix

CDanU:
`return vector<int>{1};` <- Semicolon was removed.

The check_unknown_brace_close() function did not check for _template type argument-list angle brackets_ which lead to the faulty removal.

---------

ref: #1207

---------

<s>Will be closed temporarily until an other issue is fixed.</s>

### DIFF
--- a/src/semicolons.cpp
+++ b/src/semicolons.cpp
@@ -119,12 +119,13 @@ static void check_unknown_brace_close(chunk_t *semi, chunk_t *brace_close)
    LOG_FUNC_ENTRY();
    chunk_t *pc = chunk_get_prev_type(brace_close, CT_BRACE_OPEN, brace_close->level);
    pc = chunk_get_prev_ncnl(pc);
-   if (  (pc != nullptr)
-      && (pc->type != CT_RETURN)
-      && (pc->type != CT_WORD)
-      && (pc->type != CT_TYPE)
-      && (pc->type != CT_SQUARE_CLOSE)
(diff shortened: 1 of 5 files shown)


mihaipopescu:
We have a similar example of how `mod_remove_extra_semicolon` removes semicolon after class level variable initialisers (C#,OC+).

```mm
IIterable<int>^ GenericIEnumerableProjectionTestTarget::GetIIterableOfTwoIntegers(int first, int second)
{
	auto array = ref new Platform::Array<int>
	{
		first,
		second
	};
	return ref new Iterable<int>(array);
}
```


```cs
public static readonly Vector2Quantization Uncompressed = new Vector2Quantization() {};
```

```cs
public static readonly Vector2Quantization Uncompressed = new Vector2Quantization() {}; // doesn't work because ; gets removed but 
public static readonly Vector2Quantization Uncompressed = new Vector2Quantization {}; // works and ; doesn't get removed
```

I will be creating another issue with these  but I was wondering if this fix will cover them as as well as the bugs are very similar.

CDanU:
The first example might be fixed by this, I will test it out and report.
The second example however will not.

CDanU:
I actually tested your examples now:
- the first example the semicolon is __not__ removed with the PR __and__ also with the master
- the second example still has the problem (note that if the language is set to _C++_ the semicolon stays, but is also removed if set to _C_)

PR #1216, opened 2017-06-21
Pp indent

brmqk3:


brmqk3:
Sorry for all the pushing... Never forked a repo and tried to merge it back. More complicated than I thought.

CDanU:
So the tests are running through now and the files are properly formatted now. You still have to change the CLI Tests. You could look up how t is doe based on already closed PRs.

brmqk3:
Does anyone know why my patch is failing verification? I'm not sure what it wants me to do with that line, they look the same. I attempted converting them to tabs, but that didn't work.

CDanU:
Yep, that a weird one, I will test your changes locally.

brmqk3:
Yeah also how do you run that script locally? When I run ./scripts/Test_more_Options.sh it gives me tons of console output, like entire config files instead of just the diff lines like I see in the Travis build.

CDanU:
Just run it from the repo root directory. Before that you should have a Uncrustify Release build in the build dir.
I also get way more output than shown on the Travis webpage.

CDanU:
In the help.txt file there is one space missing at the end of the line:
```diff
index 61f9e6cc..394b5082 100644
--- i/scripts/More_Options_to_Test/help.txt
+++ w/scripts/More_Options_to_Test/help.txt
@@ -67,7 +67,7 @@ uncrustify -c my.cfg --no-backup foo.d
 uncrustify -c my.cfg --prefix=out -F files.txt
 
 Note: Use comments containing ' *INDENT-OFF*' and ' *INDENT-ON*' to disable
-      processing of parts of the source file (these can be overridden with
+      processing of parts of the source file (these can be overridden with 
       enable_processing_cmt and disable_processing_cmt).
 
 There are currently 595 options and minimal documentation.

```

CDanU:
Better yet just remove that useless space from uncrustify.cpp:280.

CDanU:
Currently this PR is lacking descriptions in some places. 

- Usually the input test cpp files should be formatted with an empty config. This does not work in this case as you need options to indent the pp code blocks.
  Instead of an empty config I _think_ the __input__ test cpp files should be formatted with this options:
  ```sql
  pp_indent_count = 4 
  pp_if_indent_code = true
  ```
  As this deviates from the norm it would be nice to have a comment about this two option in the input cpp file.
  Formating the input files this way also removes the unnecessary space alterations that are currently done.

- Are the test config files minimized to the least amount of options that show what the, to be tested new option is doing? It seems like maybe some option in there could be removed.

- The option descriptions are not precise enough. Example:
  > Whether to indent case statements between #if, #else, and #endif
  This is only true for the first (nested) pp block, the code is indented in aspect to a 2nd (parent) pp block.

brmqk3:
Okay I think I have addressed all your comments i.e.

- Removed all unnecessary options from config files for tests
- Added comments to input cpp files explaining why I use two options
- Added more precise description to new options

CDanU:
I am confused what is it that you actually want to test with your test cases?

CDanU:
So you are saying that if usually there would be for example 3 indents but because of the new option only two indents are there that means that the options does something. Yes, technically that works. 
The problem is that I would need to know how much indents I am usually would expect without your option.
At this point in time I am aware of this, but what about in 10 years? 
I would prefer tests that make it blatantly clear what is being done.

Btw. I noticed that the option `pp_if_indent_code` won't work for your examples as some special handling is done for header files where the pp block _almost_ spans the whole file.
To solve this just add some variable declaration before the #if statement.

brmqk3:
Okay that's pretty easy, I'll push up a new patch soon with the inputs indented.

CDanU:
@brmqk3 I'll have to kindly ask you to review your code one more time, especially the tests. 
After the test clean up (#1446, #1453) all of your options are left without tests. This happened most likely because your tests did not actually test the new options (or my script was buggy). 

ref: #1463

Issue #1218, opened 2017-06-23
Unexpected new_line after function call with ternary condition


Shohoo:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastbin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"
[](https://pastebin.com/2ywmS4B0)
file:
```
if (1)
    test_attribute(j, (aaaaaaaaaaaaaaaaaaa + i),
                   aaaaaaaaaaaaaaaaaa[i].bbbbbb ? AAAAAAAA : BBBBBBBBBBBBBB);
else
    test_attribute(j, (aaaaaaaaaaaaaaaaaaaa + i),
                   DDDDDDDDDDDDDDD);
if (1)
    test_attribute(j, (aaaaaaaaaaaaaaaaaaa + i),
                   aaaaaaaaaaaaaaaaaa[i].bbbbbb);
else
    test_attribute(j, (aaaaaaaaaaaaaaaaaaaa + i),
                   DDDDDDDDDDDDDDD);
```

result:
```
if (1)
	test_attribute(j, (aaaaaaaaaaaaaaaaaaa + i),
[...]

PR #1219, opened 2017-06-23
Prevent addition of newline because of vbrace

CDanU:
When the option code_width is set to M, a virtual close brace (vbrace) can
be placed at column N, N>M, this caused an unexpected insertion of a
newline.

To prevent this, the vbrace that is outside of the column range is checked
if it is the last chunk on the code line. If so no break (which causes
the newline) is made.

----------------

ref: #1218


gmaurel:
@CDanU Could you have a look on it and solve?

CDanU:
Done

PR #1225, opened 2017-07-07
Improve indent_namespace_single_indent

markstegeman:
indent_namespace_single_indent worked correctly only for consecutively nested namespaces at top level. With this PR, nested namespaces should be indented more reasonably when not at top level or when not nested consecutively (e.g. when sub-namespaces are contained inside an extern "C" {}).
The included tests should explain more clearly.

The idea is to count consecutive levels of namespace nesting. Every _paren_stack_entry_t_ has an _ns_cnt_ of 0 initially, which for namespaces is then set to the _ns_cnt_ of the previous _paren_stack_entry_t_ increased by 1.
Then, having an _extern "C" {}_ or _#define_ level between nested namespaces, the _ns_cnt_ would be reset to 0.


gmaurel:
Thanks for it. We need such work(s)!

PR #1226, opened 2017-07-07
Adds options for empty parens after functions

brmqk3:
Adds options that override the normal behavior of newlines
between function names and opening parentheses when there
are no parameters in the function


brmqk3:
This last issue is the same thing where there needs to be a space at the end of the line in help.txt, and should be resolved when my other patch gets merged.

guy-maurel:
Run the script ./scripts/Test_more_Options.sh localy
Use `diff ./results/help.txt ./scripts/More_Options_to_Test/help.txt`
or `meld ./results/help.txt ./scripts/More_Options_to_Test/help.txt`
to check the problem. It might be a space which cannot displayed directly.

src/newlines.cpp
```diff
@@ -3263,6 +3275,10 @@ void newlines_cleanup_braces(bool first)
       {
          newline_iarf(pc, AV_REMOVE);
       }
+      else if (pc->type == CT_BYREF)
+      {
+         newline_iarf(pc, AV_REMOVE);
+      }
```
dirk-thomas: This inserted block seems to be responsible for the regression described in #1360.

PR #1228, opened 2017-07-08
Fix #1196 unmatched extern "C" brace

### LINKED ISSUE #1196: extern "C" { } not matched
found in latest release 0.65.

debug file: test.h
```cpp
#ifdef __cplusplus
extern "C" {
#endif


void test(void);

#ifdef __cplusplus
} // extern "C"
#endif
```
run uncrustify,
```sh
$ uncrustify --version
Uncrustify-0.65
$ uncrustify -c forUncrustifySources.cfg test.h
Output suffix: .uncrustify
Parsing: test.h as language CPP
Unmatched BRACE_CLOSE
at line=9, column=1
```
current version: uncrustify 0.65
working version: uncrustify 0.64

pkolbus:
A common pattern for headers that are used in both C and C++ sources
is to conditionally wrap declarations that need C linkage in an
extern C block, but guarded by preprocessor. Since uncrustify's
parser doesn't yet match braces across preprocessor blocks, the brace
close for this "extern C" was incorrectly seen as unmatched, resulting
in fatal error.

This patch suppresses this fatal error when inside a PP block, until
the parser can be improved.

### DIFF
--- a/src/brace_cleanup.cpp
+++ b/src/brace_cleanup.cpp
@@ -22,6 +22,7 @@
 #include "keywords.h"
 #include "logger.h"
 #include "helper_for_print.h"
+#include "indent.h"
 
 
 /*
@@ -691,11 +692,12 @@ static void parse_cleanup(parse_frame_t *frm, chunk_t *pc)
       LOG_FMT(LSTMT, "%s(%d): %zu> reset expr on %s\n",
               __func__, __LINE__, pc->orig_line, pc->text());
    }
-   else if (pc->type == CT_BRACE_CLOSE && pc->pp_level == 0)
+   else if (pc->type == CT_BRACE_CLOSE)
    {
       if (!cpd.consumed)
       {
-         if (!cpd.unc_off_used)
+         size_t file_pp_level = ifdef_over_whole_file() ? 1 : 0;
+         if (!cpd.unc_off_used && pc->pp_level == file_pp_level)
          {
             // fatal error
             char *outputMessage;
--- a/src/brace_cleanup.cpp
+++ b/src/brace_cleanup.cpp
@@ -691,7 +691,7 @@ static void parse_cleanup(parse_frame_t *frm, chunk_t *pc)
       LOG_FMT(LSTMT, "%s(%d): %zu> reset expr on %s\n",
               __func__, __LINE__, pc->orig_line, pc->text());
    }
-   else if (pc->type == CT_BRACE_CLOSE)
+   else if (pc->type == CT_BRACE_CLOSE && pc->pp_level == 0)
    {
       if (!cpd.consumed)
       {
--- a/scripts/Input/unmatched_close_pp.cpp
+++ b/scripts/Input/unmatched_close_pp.cpp
@@ -1,0 +1,12 @@
+#ifndef _Include_H
+#define _Include_H
+namespace Namespace
+{
+	// class Class {
+		public void foo()
+		{
+
+		}
+	}
+}
+#endif
--- a/scripts/Output/unmatched_close_pp.txt
+++ b/scripts/Output/unmatched_close_pp.txt
@@ -1,0 +1,2 @@
+Unmatched BRACE_CLOSE
+at line=11, column=1
--- a/scripts/Test_more_Options.sh
+++ b/scripts/Test_more_Options.sh
@@ -173,7 +173,7 @@ do
   fi
 done
 
-Liste_of_Error_Tests="I-842"
+Liste_of_Error_Tests="I-842 unmatched_close_pp"
 for Error_T in ${Liste_of_Error_Tests}
 do
   ConfigFile="${CONFIG}/${Error_T}.cfg"
--- a/tests/c.test
+++ b/tests/c.test
@@ -236,6 +236,8 @@
 01070  label_colon_nl_1.cfg    c/various_colons.c
 01071  label_colon_nl_2.cfg    c/various_colons.c
 
+01080  empty.cfg               c/bug_1196.c
+
 # big general tests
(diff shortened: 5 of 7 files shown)


src/brace_cleanup.cpp
```diff
@@ -691,7 +691,7 @@ static void parse_cleanup(parse_frame_t *frm, chunk_t *pc)
       LOG_FMT(LSTMT, "%s(%d): %zu> reset expr on %s\n",
               __func__, __LINE__, pc->orig_line, pc->text());
    }
-   else if (pc->type == CT_BRACE_CLOSE)
+   else if (pc->type == CT_BRACE_CLOSE && pc->pp_level == 0)
```
mihaipopescu: Does this mean that this test would not fail as expected ?
```cpp
#ifndef _INCLUDE_H_
#define _INCLUDE_H_

class Class
{
;
#endif
```

gmaurel:
We need some more tests.

src/brace_cleanup.cpp
pkolbus: Yes, that header no longer fails. What's the best way through this-- check and adapt for ifdef_over_whole_file() ?

On the other hand, that code will not compile... maybe it's easier and better to let the compiler catch such a problem and not leave as defined behavior how Uncrustify will handle such an input? If so then the whole error message could be removed.

pkolbus:
@guy-maurel, happy to add more tests but none come to mind at the moment. Do you have specific cases that you see as missing?

Also-- as far as I can tell, tests/run_tests.py doesn't have provision for tests that are expected to fail... is there another test runner that does, or should I also update run_tests.py?

src/brace_cleanup.cpp
mihaipopescu: There is a test in `scripts` folder that checks if an unmatched brace error is given. You can extend that. If that would have contained the test above this patch would introduce a regression which is not ok.

Your argument about that it's ok that the code will not compile doesn't really hold because you're fixing another case while breaking another.

CDanU:
I see it this way: Issue #1196 is quiet the showstopper for most C/C++ projects. They don't even need to have a bug in their code to run into this.
To fix that issue we will, most likely, have to completely change how the algo that handles the frames is working. Meaning that we won't have a optimal fix soon.

We can handle which bug we want to have: One that makes Uncrustify useless for most projects, or one that makes Uncrustify useless for buggy projects.

Now there is the matter of urgency:
If we want to not break the release cycle we can try to solve this properly, i.e. we won't merge this PR. Most people simply won't care to compile Uncrustify and will sit on the previous version.
The other alternative will be to merge this PR (including the less severe bug) and make a new Release.
Here I would decide to rather take option 1. In this cause we should maybe notify distros, that ship Uncrustify, to skip this release.

mihaipopescu:
I don't understand. In both cases we're dealing with code that doesn't compile, whether is guarded by some ifdefs or not. And I think the usecase I presented is quite common practice. What if you have:

```cpp
#ifndef _HEADER_INCLUDED_
#define _HEADER_INCLUDED_

#ifdef __cplusplus
extern "C" {
#endif

void foo(void);

#ifdef __cplusplus
}
#endif
#endif
```

CDanU:
The fix is for the issue #1196 where Uncrustify fails because it gets a open brace in pp-block N but forgets about it in pp-block M > N so that it appears that the closing brace has no matching opening brace. 

Your first example, on its own, indeed won't compile but the second should.
Am I missing something?

mihaipopescu:
Maybe i dont get it. What error where we showing before and how? Why are we counting brace level in pp?

CDanU:
Inside #1196 you can see the current error message that we are throwing. We don't count the brace level in pp, but every chunk has a pp_level to keep track in how much pp blocks it is nested. Out of the blue I don't know why this is done.

In Uncrustify 0.64 checking if closing braces had opening braces was different. I think we had it in two places.  In 0.65 the checks where moved to brace_cleanup.cpp, that made the checks more strict that before, which is now causing the problems.

mihaipopescu:
I agree we should do something to stop this error being erroneously reported on good code (if I understand correctly) but it looks like that the brace level is incremented in the pp blocks but not decrememted or something. And the fix looks like it's dealing with the effects and not really the cause.

mihaipopescu:
This is really bad. I'm seeing the regression on our code-base as well.

src/brace_cleanup.cpp
pkolbus: Correction: looking closer at the example you provide, that's an unmatched brace open. But I've added a test for unmatched brace close -- basically I-842 but wrapped in a whole file PP. And fixed my breakage.

pkolbus:
Thanks for the pointers.
FWIW, I'm in agreement that this isn't attacking the root cause, just a symptom... but it does unbreak uncrustify for a common use (mixed C/C++ projects) and enables those who want the latest improvements to build and use master again. I'm happy to keep poking at the parser code and replace this code change with something better in a future PR... unless someone else gets there first.

mihaipopescu:
No it's fine. Great work on last fix.

src/brace_cleanup.cpp
```diff
@@ -695,7 +696,8 @@ static void parse_cleanup(parse_frame_t *frm, chunk_t *pc)
    {
       if (!cpd.consumed)
       {
-         if (!cpd.unc_off_used)
+         size_t file_pp_level = ifdef_over_whole_file() ? 1 : 0;
+         if (!cpd.unc_off_used && pc->pp_level == file_pp_level)
```
mihaipopescu: While this is not ideal is a lot better than before as it covers two most used patterns.

Issue #1229, opened 2017-07-09
Debian #867376


CDanU:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=867376

gerasiov:
I did some research, results are in #1231 

I do not think, that one should try to fix v0.65, for Debian I'll just exclude it for now from s390x, ppc64 and sparc64. But there are still problems exists on master with my patches: 268 tests fail. I hope this could be fixed for next release.

I'll right more information here about tests failure later.

mihaipopescu:
Maybe we could also add at least `ppc64` and `sparc64` to travis test matrix. We have not supported Solaris systems before but it would be interesting to start supporting some RISC architectures as well.

Tangentially, we've got an `arm` build working on android (which is quite nice if you want to work on an android tablet).

gmaurel:
Yes! How to do it?

mihaipopescu:
I don't know. Maybe follow https://github.com/rust-lang/rust/blob/master/.travis.yml or using docker https://github.com/rust-lang/rust/pull/39647.

gerasiov:
@mihaipopescu all arm tests passed in Debian without problem. You can check matrix here: https://buildd.debian.org/status/package.php?p=uncrustify

mihaipopescu:
That is so cool. I didn't even knew uncrustify was compiled on so many architectures. Thanks for all the effort !

guy-maurel:
@mihaipopescu : Is it possible to add a command to the file `run_test.cmake` to get more informations within `if(files_are_different)` such as `more ${TEST_RESULT}`
and (maybe) stop the process ?

gmaurel:
Thanks for the PR

mihaipopescu:
> Is it possible to add a command to the file run_test.cmake to get more informations within if(files_are_different) such as more ${TEST_RESULT}
and (maybe) stop the process ?

Can you elaborate? I don't understand the relation with this issue.

guy-maurel:
Yes.
The bug report: `https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=867376` shows that we had a
problem. The output of the test script was very, very big. Because the same error which produces
many lines of output, was call many times.
It would be fine to detect that, make a little more output for debugging _if(files_are_different) such as more ${TEST_RESULT}_ and stop the test script, to only get the error one time. This is the relation.
But: How to add this?

mihaipopescu:
[...]

PR #1231, opened 2017-07-10
Unum fixup

gerasiov:
Error on arch s390x (and some others): wrong assumption that sizeof(int)==sizeof(size_t)

I'm packaging uncrustify for Debian and found the following serious issue:
uncrustify 0.64 builds and passes all tests without problem
uncrustify 0.65 builds and passes tests on all architectures but ppc64, sparc64, s390x.

See also #1229 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=867376

The problem came from commits 3f1565c312e95b0ac05f0265dead6d38e0a184cf and 62958075351d74f9e9d2ea47107905d50d277493.

You made wrong assumption that sizeof(int)==sizeof(size_t), but unfortunately there are platforms where this is wrong. Thus you could not handle AT_UNUM and AT_NUM the same way.

What the problem looks like:
Current master, build type Debug, every test fails with

> 36:   Uncrustify error: option
>36:   'warn_level_tabs_found_in_verbatim_string_literals' is not correctly set:
>36: 
>36:   The default value '8589934592' is more than the max value '3'.

After debugging a little bit, I found 3 errors in code:
1. Default initialization of AT_UNUM should be done with .u field, not .n (fixed in 71ca57b5)
2. In convert_value() AT_NUM and AT_UNUM should be handled differently (fixed in 5469dd1c)
3. In convert_value() assignment should be processed a little bit more complex (also fixed in 5469dd1c)

After I fixed this errors, there are only 287 failed tests left.

I looked through code a little bit more and found some more wrong usages of the union. (fixed in 5587cb3 and c6888e8)
This reduces number of failed tests to 268.

So this branch fixes some, but not all of the errors, more research is needed. But these changes are atomic, looks correct and could already be merged into master.

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -553,9 +553,9 @@ void align_right_comments(void)
 
             if (pc->orig_col < prev->orig_col_end + cpd.settings[UO_align_right_cmt_gap].u)
             {
-               LOG_FMT(LALTC, "NOT changing END comment on line %zu (%zu <= %zu + %d)\n",
+               LOG_FMT(LALTC, "NOT changing END comment on line %zu (%zu <= %zu + %zu)\n",
                        pc->orig_line, pc->orig_col, prev->orig_col_end,
-                       cpd.settings[UO_align_right_cmt_gap].n);
+                       cpd.settings[UO_align_right_cmt_gap].u);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


gmaurel:
You should use
```
if ( ...
{
}
else
{
}
```
and run yourself `./scripts/Run_uncrustify_for_sources.sh;`

gerasiov:
Wow. github and travis correctly handle branch rewrite. Great.

PR #1273, opened 2017-07-12
Add update_emscripten_bindings.py

CDanU:
Replaces `genEmscriptenEnumBindingCode.sh`, with `update_emscripten_bindings.py` which applies changes automatically to the needed files.


CDanU:
I kinda don;t understand the diff that is generated for the uncrustify_emscripten.cpp for commit 9f7a110756eb3f17888825c0f2c29cdb539d049c. Why is it including lines that are identical?

mihaipopescu:
I didn't checked but I was wondering if uncrustify has line ending normalization feature or we could use git's.

CDanU:
I had the same idea, I will open op a pr for this [branch](https://github.com/CDanU/uncrustify/commits/formatting-newlines) once this PR is merged. Uncrustify does have line end options but those won't work on the python scripts and plain text files, so I let git handle that via the `.gitattributes` file.

mihaipopescu:
On our side we actually have a generic formatter that deals with this kinds of things:
- tabs to spaces
- line endings normalization
- BOM removal
- UTF-8 encoding

PR #1274, opened 2017-07-12
AT_TFI is an extension for option(s) of type AT_BOOL.

guy-maurel:
The values could "false", "true" and "ignore".
The last one should help to realize a "null" configuration,
where uncrustify does "nothing".
This would give the possibility to use only one option.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -226,7 +226,7 @@ static void unc_add_option(const char *name, uncrustify_options id, argtype_e ty
       break;
 
    default:
-      fprintf(stderr, "FATAL: unc_add_option: Illegal option type %d for '%s'\n", type, name);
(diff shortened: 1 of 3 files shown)


src/options.cpp
```diff
@@ -2556,6 +2600,27 @@ string bool_to_string(bool val)
...
+      return("false");
+
+   case TFI_TRUE:
+      return("true");
+
+   case TFI_IGNORE:
+      return("ignore");
+
+   default:
+      fprintf(stderr, "tfi_to_string: Unknown argval '%d'\n", val);
```
CDanU: Shouldn't `__func__` be better here (and in every other log message) instead of the raw string?

guy-maurel: windows_compat.h:41
`MSVC compilers before VC7 don't have __func__ at all;
`
It is almost the same work to write `__func__` or _raw string function name_

CDanU: Do we want to support VC7? We already use __func__ plenty of times.

PR #1283, opened 2017-07-17
Add option for paren after func def

brmqk3:
The new option adds an indent to the open paren after a
function definition, if the paren is on a different line than 
the definition.


tests/input/cpp/indent_paren_after_func_def.cpp
```diff
@@ -0,0 +1,5 @@
+void SomeFunction
+(
+int const aTest,
+int const aResult
+);
```
CDanU: That is an example for a function declaration  ....

scripts/More_Options_to_Test/show_config.txt
```diff
@@ -829,6 +829,9 @@ indent_paren_close              Unsigned Number
   1: Align under the open paren
   2: Indent to the brace level
 
+indent_paren_after_func_def     { False, True }
+  Controls the indent of the open paren after a function definition, if on it's own line.If True, indents the open paren
```
CDanU: > __after__ a function definition

The parenthesis that are indented in your example are part of a function definition/declaration.

tests/input/cpp/indent_paren_after_func_def.cpp
brmqk3: It actually works the same for definition and declaration (due to how the code is already structured), but I understand what you're saying.

CDanU:
You did not understand me correctly. The problem is not your test case but the faulty description of what it is your option is going to do. Do you want to support both function declarations and definitions? Or on of them? 

While you are at it could do the following: implement an (generic) option that works for both cases (I assume that is what currently your PR does), and add specialized options for each case that overwrite the generic option. 
See how some of the options that we already have, work only on func decls or func defs.

brmqk3:
Alright, I added separate options for function calls, definitions, and declarations.

CDanU:
Ok thanks, you kinda did one of the (optional) improvement suggestions. 
The main critique about this PR ,for now, still applies.

brmqk3:
So the main critique is that the inputs aren't default formatted?

CDanU:
And the faulty descriptions. Parenthesis are part of a function call/decl/def.

brmqk3:
Ah, okay gotcha. Will fix that.

CDanU:
Descriptions are OK now.
> Your tests are a bit short, what about function decl/def that are inside a a body(struct, class, namespace, ...)? The input file is not default formatted with an empty config.

__Still__ missing ... this kinda gets too tedious.

brmqk3:
Sorry about that, forgot about the class. Alright, all of them are inside classes now.

PR #1295, opened 2017-07-20
correct 'format string is empty [-Wformat-zero-length]' warnings

nivekkagicom:
This cleans up the warnings from the Clang compiler and also eliminates the minor overhead of calling fprintf when there is really nothing to print.


mihaipopescu:
LGTM! Jus make sure `scripts/Run_uncrustify_for_sources.sh` passes on your side.

guy-maurel:
Use `./scripts/Run_uncrustify_for_sources.sh` to see:
We don't use "one liner", at src/unc_tools.cpp

nivekkagicom:
FWIW Run_uncrustify_for_sources.sh has its own set of issues:
# ./scripts/Run_uncrustify_for_sources.sh 
rmdir: illegal option -- -
usage: rmdir [-p] directory ...
some problem(s) are still present

PR #1304, opened 2017-08-04
Fix a couple of small bugs pointed out by gcc 7.1

clalancette:
The first bug has to do with the truncation of the md5_str
in backup.cpp.  snprintf was given too small of a buffer,
so it truncated the last character.  Luckily this was
the newline, so it doesn't make too much of a difference,
but this will make it more correct.

The second bug has to do with testing whether a character
is of type BRACE or PAREN.  I believe I got it right and
that we want to check if it is either a BRACE or a PAREN,
and not some logical-OR of the two.  All of the tests
pass, at least.

Signed-off-by: Chris Lalancette <clalancette@gmail.com>

### DIFF
--- a/src/backup.cpp
+++ b/src/backup.cpp
@@ -37,7 +37,7 @@ int backup_copy_file(const char *filename, const vector<UINT8> &data)
 {
    char  newpath[1024];
    char  md5_str_in[33];
-   char  md5_str[33];
+   char  md5_str[34];
    UINT8 dig[16];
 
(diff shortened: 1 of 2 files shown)


gmaurel:
Thanks for the PR.

mihaipopescu:
@clalancette I would have separated the two logical changes in two commits.
Also, comments in code are better than comments on PRs.

It would have been nice if the change would have been accompanied by a test, especially for the second fix.

PR #1328, opened 2017-08-22
Use trusty at travis

guy-maurel:


guy-maurel:
see https://docs.travis-ci.com/user/reference/trusty/

PR #1332, opened 2017-08-31
options.cpp: Copy from defaults to settins whole union, not first field.

gerasiov:
Otherwise size_t was not copied correctly on s390x arch.

This is followup to #1229 and #1231.

Before this patch on s390x we have
    63% tests passed, 283 tests failed out of 763
after
    100% tests passed, 0 tests failed out of 763

Signed-off-by: Alexander GQ Gerasiov <gq@cs.msu.su>

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -2523,7 +2523,7 @@ void set_option_defaults(void)
    // copy all the default values to settings array
    for (unsigned int count = 0; count < UO_option_count; count++)
    {
-      cpd.settings[count].a = cpd.defaults[count].a;
+      cpd.settings[count] = cpd.defaults[count];
    }
 } // set_option_defaults
 


guy-maurel:
Please make a comment to document which OS, which compiler are used.
Is there any way to open a bug for that? and hope to get a solution!

gerasiov:
>Please make a comment to document which OS, which compiler are used.
It was Debian Linux sid arch s390x. (I think gcc 7.1.0). The same problem was on ppc64 and sparc64.

>Is there any way to open a bug for that? and hope to get a solution!
Eh? Open bug? Isn't opening PR enough?

guy-maurel:
I think about opening a bug at gcc

gerasiov:
>I think about opening a bug at gcc

Why so? There was the error in uncrustify code, I believe.

PR #1351, opened 2017-09-20
Add options empty function calls

brmqk3:
This adds the option to override nl_func_call_paren for
empty function calls, and an option that does the same thing
as nl_func_def_empty, but for function calls. Also fixes some
code in newlines.cpp, and the tests associated with that code.


guy-maurel:
We are using about three words for almost the same:
 `src/options.cpp:724` with `indent_continue`
and some more,
`nl_func_call_paren_empty` (yours) `Overrides nl_func_call_paren for ..`.
`src/options.cpp:478` with` sp_after_operator_sym_empty` as:
`Have precedence of sp_after_operator_sym`
`src/options.cpp:361` with `sp_before_unnamed_ptr_star` as:
`If set to 'ignore', sp_before_ptr_star is used instead`
`src/options.cpp:966` with `nl_typedef_blk_start` as:
`the option 'nl_after_access_spec' takes preference over...`

Should we not try to:
1. define what we wanted to say,
2. modify/delete some expressions,
3. get a more relevant text for the options?

brmqk3:
Well I'm using it exactly the same way for nl_func_paren and nl_func_def_paren, so I'm trying to keep the same wording as those options. I wouldn't be opposed to changing the other places that we are using similar wording to the same wording i.e. "Overrides xxxx option for yyyy condition". Should I do that?

CDanU:
I would not recommend to change the names of other options yet, over the years there surely where created things that now are messy and can e made easier. Changing the naming would mean that we break old config files. We should try to minimize te amount of such breaks, so we have to do this in an more organized way.

Making a table of names and their planned replacements would be nice.
Chancing descriptions is always possible though.
 
About your commit: it is fairly long and it changes existing tests, could you try to break it up into smaller commits?

guy-maurel:
I spoke ***only*** about the descriptions of the options.
Should we try to find some vocabulary which are the same for all options?

brmqk3:
Okay I can split this into 3 commits:
1. Will do the fixes for the newline stuff and the associated tests
2. Add newline options for function calls and associated tests, will rename the function to show it deals with function calls.
3. When we decide on the wording, create commit that modifies the descriptions to stick with the same vocabulary for these similar options.

In reference to the scope of the variables, I think they need to be at that scope. Otherwise I'm going to be declaring them 3 separate times in this function. And they shouldn't affect anything, as we assign them a value right before we use them.

CDanU:
>> Otherwise I'm going to be declaring them 3 separate times in this function

Which is fine, I know this is just nitpicking but I don't like those big blocks of variable definitions at the beginning.

>> Okay I can split this into 3 commits:
Sound good, but make that four, one additional only for the function renaming.

gmaurel:
Let us informed when the split is ready.

gmaurel:
@brmqk3 : which is the status now?

brmqk3:
Hey @guy-maurel, this has been split up into multiple branches and merged already. I'll go ahead and abandon this.

CDanU:
Hmm I havn't even noticed that they where indeed merged, please add references to this PR too keep everything in check.

brmqk3:
Sure, should be these:
#1362 
#1364 
#1356

PR #1356, opened 2017-09-26
Fix newline between function name and open paren

brmqk3:
The current test cases don't accurately test what these
options are supposed to do, and the current implementation
doesn't work correctly either. This fixes both the
implementation and test cases so that they correctly format
the newline between the function name and open paren. Also 
adds the options that these override, to prove they do in 
fact override them.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -2253,7 +2253,7 @@ static void newline_func_def(chunk_t *start)
       atmp = cpd.settings[is_def ? UO_nl_func_def_paren_empty : UO_nl_func_paren_empty].a;
       if (atmp != AV_IGNORE)
       {
-         prev = chunk_get_prev_ncnl(pc);
+         prev = chunk_get_prev_ncnl(start);
          if (prev != NULL)
          {
             newline_iarf(prev, atmp);
--- a/tests/config/nl_func_def_paren_empty.cfg
+++ b/tests/config/nl_func_def_paren_empty.cfg
@@ -1,1 +1,4 @@
+# The option that is being overridden by nl_func_def_paren_empty
+nl_func_def_paren = force
+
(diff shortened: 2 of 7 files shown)


CDanU:
Good catch I did not notice this change in the first PR.

Could you change the tests so that but variants are available: 
```
function( 
)
```
and
```
function
()
```
It would also be nice if you could add a comment to the options that are going to be overwritten. 

The reason for this is that at some point I will continue to clean up the tests so that unnecessary options are going to be removed. See [option_reducer.py](https://gist.github.com/CDanU/54446524c8919b1d490147f6c8165bb6#file-option_reducer-py), that script unfortunately bluntly removes unneeded options, including the ones that are going to be overridden. A comment in there would save up a lot of time while validating the results.

brmqk3:
So as input the tests would look like:
```
function
    (
    )
```
and
```
function
    ()
```
And the ouptuts would respectively look like:
```
function(
    )
```
and
```
function()
```
That's what we are looking for right?

And should I add those comments in the config files for the tests?

CDanU:
> That's what we are looking for right?

almost, the in the first variant the parenthesis should be on the line with `function`
I am currently not entirely sure how the output would look like because of the closing parenthesis.

>And should I add those comments in the config files for the tests?

into the config files

brmqk3:
Okay for this input:
```
function(
    )
```
The output would be unchanged, since our config is to remove the newline between the function name and the open paren.

CDanU:
In that case:
```
function
    (
    )
```
```
function(
    )
```
```
function
    ()
```

should create:
```
function(
    )
```
```
function(
    )
```
```
function()
```

brmqk3:
Correct. I will include all those permutations in the tests.

PR #1358, opened 2017-09-27
Add options for align_func_params (span, thresh, gap)

KVic:
This adds extra options for aligning parameter definitions in function on parameter name:
- align_func_params_span
- align_func_params_thresh
- align_func_params_gap


CDanU:
>The Command line interface (CLI) output is tested by the scripts/Test_more_Options.sh script. It operates on the files inside this directories:
>
>    scripts/Config/
>    scripts/Output/
>    scripts/Input/
>    scripts/More_Options_to_Test/
>
> __If a PR is altering the CLI output__, files inside those directories might need to be manually updated. __This often happens when options are added__, removed or altered. Keep in mind that the version string line (example: # Uncrustify-0.65_f) of outputs from commands like --show-config should be replaced with a blank line.

guy-maurel:
Very fine.
Thanks.

PR #1359, opened 2017-09-27
Fix alignment with extra spaces after the * or &.

KVic:
Aligning does not work correctly in variable definitions with extra spaces between '*'/'&' and variable name:
```
void test()
{
float & c;
double &     d;
}
```
produce:
```
void test()
{
	float & c;
	double & d;
}
```
expected:
```
void test()
{
	float &  c;
	double & d;
}
```

Config:
```
align_var_def_span   = 1
align_var_def_thresh = 4
```

This PR is an option how to fix this issue.

### DIFF
--- a/src/align_stack.cpp
+++ b/src/align_stack.cpp
@@ -65,140 +65,143 @@ void AlignStack::Add(chunk_t *start, size_t seqnum)
 
    m_last_added = 0;
 
+   // Threshold check should begin after
+   // tighten down the spacing between ref and start
+
+   /*
+    * SS_IGNORE: no special handling of '*' or '&', only 'foo' is aligned
+    *     void     foo;  // gap=5, 'foo' is aligned
+    *     char *   foo;  // gap=3, 'foo' is aligned
+    *     foomatic foo;  // gap=1, 'foo' is aligned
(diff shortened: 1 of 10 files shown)


CDanU:
Neat, could you check if 
```sql
# Controls the positioning of the '*' in typedefs. Just try it.
# 0: Align on typedef type, ignore '*'
# 1: The '*' is part of type name: typedef int  *pint;
# 2: The '*' is part of the type, but dangling: typedef int *pint;
align_typedef_star_style        = 1        # unsigned number

# Controls the positioning of the '&' in typedefs. Just try it.
# 0: Align on typedef type, ignore '&'
# 1: The '&' is part of type name: typedef int  &pint;
# 2: The '&' is part of the type, but dangling: typedef int &pint;
align_typedef_amp_style         = 1        # unsigned number
```

is working with this? I am not sure if there is already a test for that.

KVic:
I added more tests for the option align_var_def_thresh.

This fix does not affect the align_typedef_xxx so much. Also, there is no align_typedef_thresh option. Therefore, it should be in another fix.

gmaurel:
Thanks for the fix.

PR #1362, opened 2017-09-29
Standardize option descriptions to "override"

brmqk3:
These options have similar descriptions:
*sp_after_operator_sym_empty - Have precedence of...
*nl_typedef_blk_start - takes preference over...
*nl_func_paren_empty - Overrides...

This change modifies these descriptions to use the
"override" wording. This doesn't make any changes to
the options like sp_before_unnamed_ptr_star, where if it's
set to ignore, another options is used, and isn't really
overridden.


gmaurel:
You have to modify the file scripts/Output/mini_nd_ucwd.txt to reflect your changes.

brmqk3:
Modified.

PR #1364, opened 2017-09-29
Add options for empty function calls and rename nl_func_def method

brmqk3:
This pull request brings in two commits:
1. Adds options for empty function calls
2. Renames nl_func_def to nl_func_def_or_call since it now also deals with function calls.


gmaurel:
It is not clear if you mean "the option" or the "name of the function in the code".
Please change the description to:
_Rename the function name nl_func_def to show it deals with calls _

PR #1367, opened 2017-10-03
Fixed #1366 - Wrong indentation of the break in switch

### LINKED ISSUE #1366: Wrong indentation of the break in switch
The issue is mentioned in the #663, so extracting it to the separate one.

**Break** positioned outside of the brackets in the switch case, indented aligning to the end of the case line, not to the beginning of it.
In:
```
@implementation UCTestClass

- (void) test{

    switch (test) {
      case "longlonglonglong":{
          i = 1;
      }
      break;
    }
}

@end
```

Actual out:

```
@implementation UCTestClass

- (void)test
{
    switch (test) {
        case "longlonglonglong": {
            i = 1;
        }
                                 break;
    }
}

@end
```

Krizai:
Place the **break** based on the first item in the previous line

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1547,10 +1547,17 @@ void indent_text(void)
             && prev->type == CT_BRACE_CLOSE
             && prev->parent_type == CT_CASE)
          {
-            // issue #663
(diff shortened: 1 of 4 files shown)


gmaurel:
Thanks

Issue #1287, opened 2017-07-18
rule-3 of comment alignment should not work for #endif's


mihaipopescu:
Expected versus output diff:
---
```diff
diff --git "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\60030-UNI-21727.cpp" "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\60030-UNI-21727.cpp"
index c2b6711..645ea21 100644
--- "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\60030-UNI-21727.cpp"
+++ "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\60030-UNI-21727.cpp"
@@ -1,10 +1,10 @@
 void foo()
 {
 #if defined(SUPPORT_FEATURE)
     bar();
 #endif // SUPPORT_FEATURE
-    // Handle error
+       // Handle error
     if (error != 0)
     {
     }
 }
```

Debug support files:
---
[60030-UNI-21727.cpp](https://gist.githubusercontent.com/mihaipopescu/2bab90a2932dfbf879ce21a44aa38fa8/raw/c2b6711439c0ce65e480bb0a11d3ce734074efb6/60030-UNI-21727.cpp)
[60030-UNI-21727.cpp.log](https://gist.githubusercontent.com/mihaipopescu/2bab90a2932dfbf879ce21a44aa38fa8/raw/726cb5b92a0d37ce52d54d34e8070a420ab8de06/60030-UNI-21727.cpp.log)
[60030-UNI-21727.cpp.unc](https://gist.githubusercontent.com/mihaipopescu/2bab90a2932dfbf879ce21a44aa38fa8/raw/1d7488da376a0c4eed260b6b7f57233643dfa5d0/60030-UNI-21727.cpp.unc)
[Uncrustify.Cpp.cfg](https://gist.githubusercontent.com/mihaipopescu/2bab90a2932dfbf879ce21a44aa38fa8/raw/23d62d9dba5caf352fdc1793792e98eed4b81929/Uncrustify.Cpp.cfg)
[UNI-21727.cpp](https://gist.githubusercontent.com/mihaipopescu/2bab90a2932dfbf879ce21a44aa38fa8/raw/c2b6711439c0ce65e480bb0a11d3ce734074efb6/UNI-21727.cpp)

Disclaimer:
---
Note: This is a generated issue from our internal issue tracker.
[...]

PR #1372, opened 2017-10-09
 Fix indent comment alignment rule 3

KVic:
Also this fixes issue #1287.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -2622,6 +2622,33 @@ static bool single_line_comment_indent_rule_applies(chunk_t *start)
 } // single_line_comment_indent_rule_applies
 
 
+static size_t calc_comment_next_col_diff(chunk_t *pc)
+{
+   chunk_t *next = pc;
+
+   do
+   {
+      chunk_t *next_nl = chunk_get_next(next);
+
+      if (next_nl == NULL || next_nl->nl_count > 1)
+      {
+         // FIXME: Max thresh magic number 5000
+         return(5000);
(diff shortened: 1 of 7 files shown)


src/indent.cpp
```diff
@@ -2622,6 +2622,33 @@ static bool single_line_comment_indent_rule_applies(chunk_t *start)
...
+   chunk_t *next = pc;
+
+   do
+   {
+      chunk_t *next_nl = chunk_get_next(next);
+
+      if (next_nl == NULL || next_nl->nl_count > 1)
+      {
+         // FIXME: Max thresh magic number 5000
+         return(5000);
```
mihaipopescu: I would use `int` and return -1 here to express an invalid value. Also, the name of the function could contain `abs` to clarify that.

There is no need to use `size_t` here.

mihaipopescu:
Thanks for looking into this one.

src/indent.cpp
KVic: The int(-1) is also a magic number :). A std::optional should be used here, but this is a С++17. The magic number size_t(5000) is widely used in the options.cpp. It must be replaced everywhere with some constant, such as std::numeric_limits<size_t>::infinity().

mihaipopescu: I agree with `-1`. I'm was just arguing against using `size_t`, and any other unsigned values, for anything other than sizes or bit-field operations. They introduce a lot of problems when used along with `int` due to implicit integer promotions.

PR #1373, opened 2017-10-11
Fix keywords for Objective-C and C

gentoo-root:
Add missing Objective-C nullability keywords. Remove LANG_C flag from
C++-only keywords. Add Objective-C BOOL type.


guy-maurel:
Have a look to:
_https://travis-ci.org/uncrustify/uncrustify/jobs/286963341_
2 tests have not passed:
```
c_00453_ (Failed)
objective-c_50800 (Failed)
```
You can test it yourself for commiting by using the script:
`./scripts/Test_more_Options.sh`

gentoo-root:
Yes, thank you, I've seen it, and I'm working on it. I'll let you know when I get that fixed.

In master, there are some invalid tests that have incorrect output, but are passing, because uncrustify fails to work correctly. After my changes these tests began to fail, and I'm going to fix the tests. The objective-c_50800 test indeed caught a bug caused by my changes, and I'm going to fix my code.

gentoo-root:
Hmm, this time I don't understand the errors of the CI. I can't reproduce it on my machine, but CI fails to test debug builds showing the following message:
```
1: CMake Error at C:/projects/uncrustify/tests/run_test.cmake:44 (message):
1:   Uncrustify error: keywords_are_sorted: bad sort order at idx 24, words
1:   '__nullable' and '_Bool'
1: 
1:   Assertion failed: keywords_are_sorted(), file
1:   C:\projects\uncrustify\src\uncrustify.cpp, line 324
1: 
1: 
```
I double-checked with ASCII table, and I'm sure that _Bool should be before __nullable, so I don't understand why it complains. I've run scripts/Run_uncrustify_for_sources.sh, and it doesn't complain. Also I don't understand why it happens during testing and why it doesn't happen on my machine.

What should I do with that? :)

CDanU:
I get the same error on my machine, the release builds are running through because they most likely ignore the assert.

`_` has ASCII 95, while `B` has 66

gentoo-root:
My bad, I have them in the correct order locally, but I forgot to push that change :(
Sorry for the extra noise.

CDanU:
Ahh ok, I just noticed that my last comment wasn't all that helpful. 

The commits are a bit messy. I think that you could combine commit 1 and 3. If it doesn't cause you to much trouble I would prefer to have the new keywords added and than all the formatting applied in an extra commit.

gentoo-root:
>If it doesn't cause you to much trouble I would prefer to have the new keywords added and than all the formatting applied in an extra commit.

OK, no problem, I'll reorganize my commits.

gentoo-root:
I've reorganized the commits, and all checks still pass.

PR #1374, opened 2017-10-11
Fix alignment in 'operator' declarations.

KVic:
It fixes some incorrect alignments in the `operator` declarations.

Config:
```
indent_columns = 4

sp_after_operator = remove

align_func_proto_span = 2
align_var_def_span    = 2
align_mix_var_proto   = true
align_on_operator     = true
```

Source:
```
// 'sp_after_operator = remove' eats the space
Foo& Foo::operator =(const Foo& a);
void *Foo::operator new(size_t x);

// 'align_on_operator = true' invalid alignment
bool operator==(const FooBar& a);
FooBar& operator =(const FooBar& a);
```

Invalid result:
```
// 'sp_after_operator = remove' eats the space
Foo& Foo::operator =(const Foo& a);
void *Foo::operatornew(size_t x);

// 'align_on_operator = true' invalid alignment
bool operator   ==(const FooBar& a);
FooBar&         operator=(const FooBar& a);
```

Valid result:
```
// 'sp_after_operator = remove' eats the space
Foo&  Foo::operator=(const Foo& a);
void *Foo::operator new(size_t x);

// 'align_on_operator = true' invalid alignment
bool    operator==(const FooBar& a);
FooBar& operator=(const FooBar& a);
```

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -1310,6 +1310,7 @@ static chunk_t *align_var_def_brace(chunk_t *start, size_t span, size_t *p_nl_co
 
       if (fp_active && !(pc->flags & PCF_IN_CLASS_BASE))
       {
+         // WARNING: Duplicate from the align_func_proto()
          if (  pc->type == CT_FUNC_PROTO
             || (  pc->type == CT_FUNC_DEF
                && cpd.settings[UO_align_single_line_func].b))
@@ -1317,7 +1318,17 @@ static chunk_t *align_var_def_brace(chunk_t *start, size_t span, size_t *p_nl_co
             LOG_FMT(LAVDB, "%s(%d): add=[%s], orig_line is %zu, orig_col is %zu, level is %zu\n",
                     __func__, __LINE__, pc->text(), pc->orig_line, pc->orig_col, pc->level);
 
-            as.Add(pc);
+            chunk_t *toadd;
+            if (  pc->parent_type == CT_OPERATOR
+               && cpd.settings[UO_align_on_operator].b)
+            {
+               toadd = chunk_get_prev_ncnl(pc);
+            }
+            else
+            {
+               toadd = pc;
+            }
+            as.Add(step_back_over_member(toadd));
             fp_look_bro = (pc->type == CT_FUNC_DEF)
                           && cpd.settings[UO_align_single_line_brace].b;
          }
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -1628,8 +1628,17 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
                   || tmp->type == CT_QUALIFIER)
                {
                   set_chunk_type(prev, CT_TYPE);
-                  set_chunk_type(pc, CT_ADDR);
-                  chunk_flags_set(next, PCF_VAR_1ST);
+                  if (  next->type == CT_OPERATOR
+                     || next->type == CT_TYPE
+                     || next->type == CT_DC_MEMBER)
+                  {
+                     set_chunk_type(pc, CT_BYREF);
+                  }
+                  else
+                  {
+                     set_chunk_type(pc, CT_ADDR);
+                     chunk_flags_set(next, PCF_VAR_1ST);
+                  }
                }
                else if (tmp->type == CT_DC_MEMBER)
                {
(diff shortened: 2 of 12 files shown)


guy-maurel:
Don't forget to use the tests:
`./scripts/Run_uncrustify_for_sources.sh`
and
`./scripts/Test_more_Options.sh`
before commiting. It might be necessary to make some more changes.

KVic:
I'm on Windows. So I use Run_uncrustify_for_sources.bat. Use Test_more_Options.sh is problematic. It would be better if these scripts were python scripts, like run_tests.py.

KVic:
And I don't understand what [this error](https://travis-ci.org/uncrustify/uncrustify/jobs/286450346#L5489) is about :confused:.

CDanU:
The error is about the output generated by the `./scripts/Run_uncrustify_for_sources.sh` script does not match the expected files: The script checks if the generated `./results/66.txt.sed` is matching the existing `scripts/Output/66.txt`. Your commits changes the generated files, so you need to check if the generated output is correct and than alter `scripts/Output/66.txt`.

KVic:
@CDanU, thanks for the explanation!

I explored the [differences](https://travis-ci.org/uncrustify/uncrustify/jobs/286450346#L5490). They are caused by [this change](https://github.com/uncrustify/uncrustify/pull/1374/commits/2e1820e4dcdce2baceb98604cffe86fffe5585b8#diff-51544c86dac45f8436f233abfc7f73feR2294), which forces the space after the token (in particular `operator`) with the `PCF_FORCE_SPACE` flag. This is the essence of this PR, so the the generated `./results/66.txt` should be correct. But in Windows, I can't generate the correct `66.txt` to replace the `scripts/Output/66.txt`. The output of logs, in general, looks different:

`scripts/Output/66.txt`
```
do_space(): orig_line is , orig_col is , first->text() 'const', type is QUALIFIER
log_rule(): Spacing: first->orig_line is , first->text() is 'const', [QUALIFIER/NONE] <===>
   second->text() 'char', [TYPE/NONE] : rule sp_after_type[line ]
space_col_align(): av is ,    len is    => coldiff is 
space_col_align(): orig_line is , orig_col is , [TYPE/NONE] 'char' <==> orig_line is , orig_col is  [PTR_TYPE/NONE] '*' [CallStack:-DEBUG NOT SET-]
```
my logs:
```
???(143): orig_line is 3, orig_col is 15, first->text() 'const', type is QUALIFIER
???(124): Spacing: first->orig_line is 3, first->text() is 'const', [QUALIFIER/NONE] <===>
   second->text() 'char', [TYPE/NONE] : rule sp_after_type[line 1711]
 <force between 'const' and 'char'>???(2296): av is 3,    len is 5   => coldiff is 6
???(2285): orig_line is 3, orig_col is 21, [TYPE/NONE] 'char' <==> orig_line is 3, orig_col is 25 [PTR_TYPE/NONE] '*' [CallStack:-DEBUG NOT SET-]
```

CDanU:
The outputs a bit different because internally `Test_more_Options.sh` uses sed to apply some replacements. `scripts/Output/66.txt` is adjusted to that.

I have finished the [python version](https://github.com/CDanU/uncrustify/blob/aa6952d40bd552992679337e9459dd79e794ffc9/tests/cli/test_cli_options.py) of the `Test_more_Options.sh` script, it is set up to run the tests in a new location, so you have to wait until it is merged.

gmaurel:
#1379 is now merged.
You can make your next changes and tests

CDanU:
The position of scripts/Output/66.txt (and all the other cli related tests) has been changed to tests/cli/*.

KVic:
#1377 should also be merged.

And, I think, symbolic links (in tests\cli\Input) under version control is not a good idea.

CDanU:
Symbolic links? That is indeed not a good Idea, I can't remember putting in some.

KVic:
It looks like 9.cpp - 92.cpp (except 28.cpp) files are symbolic links to 28.cpp. Their content is simple:
`28.cpp`

CDanU:
Right, those where set up before my commits this way ... I will changed that soon. Is this blocking you in any way?

gmaurel:
#1377 is now merged.

KVic:
It looks like I done this PR  :sweat:.

gmaurel:
may I merge?

KVic:
Yes, thanks.

PR #1377, opened 2017-10-14
Fix log output in Windows with GCC

KVic:
Checking `_MSC_VER` sets `__func__` to `???`, which is incorrect for GCC.


gmaurel:
Thanks

Issue #638, opened 2016-08-21
Alignment of second line of a ternary conditional operator


jayvdb:
For ternary conditional operators, especially when the line length is exceeded, I would like the line split on the `:`, and the second line to be aligned to after the `?`.

i.e. the following:

``` c
    y = x ? foo :
            bar;

    if (x ? foo :
            bar) {
        baz;
    }
```

I cant find an existing incantation to achieve this.  My apologies if that is already possible.

jayvdb:
They are currently being reformatted to be:

```c
    y = x ? foo :
        bar;

    if (x ? foo :
        bar) {
        baz;
    }
```

I suspect that I can improve the way it is reformatted, but I would prefer it isnt reformatted.

gmaurel:
Please push your actual config

jayvdb:
@gmaurel , how is my config related?

Are you implying that there is support for aligning ternary conditional operators spanning multiple physical lines as I have indicated in the first post?

gmaurel:
Are you using an empty config?

jayvdb:
I have tried empty config, and 50+ other config combinations. None appear
to add support for the special indentation alignment I need for ternary
conditional operators.
I've also dug around in the source and not found the functionality I need.

bengardner:
I don't think there is an option for that style.   A new option would be needed.

Soukyuu:
If such an option will be added, I'd like to also be able to align the ternary operator like this, if possible:
```
int a = flag ? b
             : b + c;
```
That's about the only thing missing for me to be able to uncrustify my code, I'm amazed how far I got considering how early some of the other tools I tried failed...

jayvdb:
I would be happy to switch code to the same style as @Soukyuu offers directly above:
```
int a = flag ? b
             : b + c;
```

And likewise, this is the last hurdle for me; I can swallow most other minor changes needed to deploy uncrustify, but ... the currently imposed layout for ternary conditional operators is a deal-breaker.

mihaipopescu:
@Soukyuu @jayvdb This option will be nice indeed ! Patches are most welcomed !

PR #1381, opened 2017-10-17
Don't force-align with tabs when indenting ternary ':'

ojohnny:
When using the following settings, a ternary operator ':' should be
aligned to '?', yet also adhering to the 'indent with tabs, align with
spaces' setting.

```
indent_with_tabs        = 1
indent_ternary_operator = 2
```

The result however is that the ':' is aligned to '?' as if
`indent_with_tabs = 2`.
(using tabs as much as it can, then aligning the last bit using spaces)

I tracked this down to a special handling of ':' in `src/output.cpp`.

Here is where I have a caveat: I don't understand what special case this
line was supposed to handle. I tried tracking it down in the git history,
but I ended up at commit fe25b6d0 ("Import r1644 from subversion"),
and then I couldn't find the old subversion repository.

No tests broke when removing this line (except for the test for
aligning of the ternary operator ':'), so it would appear to only exist
to make sure that the ternary ':' is aligned using tabs even though
`indent_with_tabs = 1` which I would argue is wrong.

I guess my only hope to understand what this line really is supposed to
do is to remove it and wait for someone to yell at me that something
would/did break when applying this commit.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -604,10 +604,9 @@ void output_text(FILE *pfile)
                size_t lvlcol;
                /*
                 * FIXME: it would be better to properly set column_indent in
-                * indent_text(), but this hack for '}' and ':' seems to work.
+                * indent_text(), but this hack for '}' and '#' seems to work.
                 */
                if (  pc->type == CT_BRACE_CLOSE
-                  || chunk_is_str(pc, ":", 1)
                   || pc->type == CT_PREPROC)
                {
                   lvlcol = pc->column;
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -607,6 +607,7 @@ void output_text(FILE *pfile)
                 * indent_text(), but this hack for '}' and '#' seems to work.
                 */
                if (  pc->type == CT_BRACE_CLOSE
+                  || pc->type == CT_CASE_COLON
                   || pc->type == CT_PREPROC)
                {
                   lvlcol = pc->column;
--- a/tests/output/c/09604-indent_ternary-2.c
+++ b/tests/output/c/09604-indent_ternary-2.c
@@ -1,13 +1,13 @@
 void foo(void)
 {
 	int a = x ? y
-		  : z,
+	          : z,
 	    b = x ? (y)
-		  : (z),
+	          : (z),
 	    c = x ? *y
-		  : *z,
+	          : *z,
 	    d = x ? &y
-		  : &z;
+	          : &z;
 
 
 	if (x ? y


CDanU:
Thanks for the PR, I will review it later. The old SVN repo was transitioned recently, see: [Full-history](https://github.com/uncrustify/uncrustify/wiki/Full-history)

ojohnny:
Thank you for the pointer to the old history, it hadn't crossed my mind to look at the wiki for it.

The line I removed here seems to have been added in e5779e3de ("fix mixed tab indenting for case stuff"), but it doesn't seem right in that context either.

Also, after reading #638 there is no indication that the current behaviour (using tabs for alignment) is the intended one.

CDanU:
I also don't see what the line you removed was intended to do. 
What would your expected output look like? 

----------------------

I certainly am unsure what `indent_with_tabs` is actually supposed to do in the fist place when value 1 or 2 is used. 
```sql
# How to use tabs when indenting code
# 0=spaces only
# 1=indent with tabs to brace level, align with spaces (default)
# 2=indent and align with tabs, using spaces when not on a tabstop
indent_with_tabs                = 0        # unsigned number
```

What does `brace level` mean here? Indenting until the brace level of the block the processed line is?
That would mean the this wouldn't be applied on functions (unless they are in a class/struct or a namespace)


Here is a small test (with 37c30cf5):
C:
```sql
input_tab_size                  = 3
output_tab_size                 = 3
indent_columns                  = 3
indent_ternary_operator         = 2
```



I:
```cpp
class P
{
public:
void f()
{
    int a = 7;
	int b = (a < 0) ? 1
			 : 0;

	switch(b)
	{
           case 1
:
              a = 0;
break;
case 2
:
{
 a = 1;
break;
}
	}
}
};
```
O:
```cpp
class p
{
public:
	void f()
	{
		bool b = (a < 0) ? 1
							  : 0;

		switch(b)
		{
		case 1
			:
			a = 0;
			break;
		case 2
			:
		{
			a = 1;
			break;
		}
		}
	}
}
```
There is not a single line that is aligned with tabs until the brake level.

CDanU:
Looking at the debug I see that case colons have their own types, so to be safe you could change the unspecific char comparison of the ':' with a comparison of the token type with `CT_CASE_COLON`

ojohnny:
I agree that 'brace level' is misleading, and I thought (and still hope) it meant 'indent level'.

The thing I'm complaining about is this line in your output:
```c
		bool b = (a < 0) ? 1
							  : 0;
```
It should be clear that the `: 0` isn't aligned to `? 1` at all if the tabstop isn't kept the same. (which is why I would like it to indent with tabs and align with spaces)

Here is the output of your sample using 0fc6091f (this merge request).

```c
class P
{
public:
	void f()
	{
		int a = 7;
		int b = (a < 0) ? 1
		                : 0;

		switch(b)
		{
		case 1
			:
			a = 0;
			break;
		case 2
			:
		{
			a = 1;
			break;
		}
		}
	}
};
```
The only real difference I can see is the alignment of the ternary operator, which is now correct. The other colons have the same indentation as in your output.

But if you'd like, I can add the `CT_CASE_COLON` to `src/output.cpp` in a new commit, but I'm still not sure it makes any difference.

CDanU:
>> But if you'd like, I can add the CT_CASE_COLON to src/output.cpp in a new commit, but I'm still not sure it makes any difference.

It most likely does not, but a new release is planned soon so I'd rather be safe. I will remove the line myself once the new release is out.

gmaurel:
The information about `https://joskar.se` is wrong.
Could you correct it? Thanks

ojohnny:
Assuming you are talking about the SSL error on my domain, then I just noticed and (hopefully) corrected it. Apache weirded out on me. Sorry for the inconvenience.

PR #1383, opened 2017-10-18
Fix test_cli_options.py for Windows newlines issues.

KVic:


CDanU:
Thanks for the PR, this time I actually tested the script on windows and for some reason it still, even with your changes does not work. 

According to the documentation of the [io package](https://docs.python.org/2/library/io.html#index-0):
>newline controls how universal newlines works (it only applies to text mode). It can be None, '', '\n', '\r', and '\r\n'. It works as follows:
>
>    On input, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.
>    On output, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string.

the line separators should remain untouched with `''`, on windows I still get `\r\n` written with python 3.6.

I edited your commit so that it now always replaces line endings, before the manipulations, comparing and writing. I will submit that soon. One thing that I can't wrap my Head around are the changes from:
```
 gen_result_manip=reg_replace(r'\# Uncrustify.+', '')):
```
to
```
 gen_result_manip=reg_replace(r'\# Uncrustify.+\n', linesep)):
```

My regex shouldn't catch the line ending, and therefore it should not be changed.

KVic:
The line ending may depend on the `core.autocrlf` git config option. I have it `core.autocrlf = true` in the systemwide git config. I can set `core.autocrlf = input` for the Uncrustify repository, but I'm not sure that this decision is good. Someone else with `core.autocrlf = true` in Windows can face the same problem.

With
`gen_result_manip=reg_replace(r'\# Uncrustify.+', '')):`
i have the first line with `LF`, and the rest with `CRLF`, but in the sample files in `tests\cli\Output` everywhere `CRLF` and they do not match.

gmaurel:
May I close?

KVic:
Yes, #1384 is much better.

gmaurel:
Fine

PR #1385, opened 2017-10-20
Dc member

guy-maurel:
We need a call to a new function `chunk_get_prev_ncnlnpnd`
The function look back over DC_MEMBER also.


gmaurel:
@CDanU @KVic : why do I get a big difference between the test with AppVeyor and the one with Travis?

src/combine.cpp
```diff
@@ -4123,7 +4100,8 @@ static void mark_function(chunk_t *pc)
 #ifdef DEBUG
       LOG_FMT(LFCN, "%s(%d):", __func__, __LINE__);
 #endif
-      LOG_FMT(LFCN, "  Checking func call: prev=%s", (prev == NULL) ? "<null>" : get_token_name(prev->type));
+      LOG_FMT(LFCN, "  Checking func call: prev->text() '%s', prev->type is %s",
+              prev->text(), (prev == NULL) ? "<null>" : get_token_name(prev->type));
```
KVic: Here `prev` can be `NULL`.

KVic: `prev->text()` leads to crash.

guy-maurel:
@KVic : Many thanks to you for your help.
Still a question:
Why don't I get an error message/crash with Linux?
Do anybody understand the difference(s)?

KVic:
The macro `NO_MACRO_VARARG` is defined in Windows (windows_compat.h(16)). With this macro, the `log_fmt` (logger.h(117)) is called and leads to crash. Without this macro, the `log_sev_on` (logger.h(121)) returns `true` and the `log_fmt` is not called and we are safe :).

src/combine.cpp
```diff
@@ -4196,8 +4186,24 @@ static void mark_function(chunk_t *pc)
...
-            prev = chunk_get_prev_ncnlnp(prev);
+#ifdef DEBUG
+            LOG_FMT(LFCN, "%s(%d):", __func__, __LINE__);
+#endif
+            LOG_FMT(LFCN, " <skip '%s'>", prev->text());
+#ifdef DEBUG
+            LOG_FMT(LFCN, "\n");
+#endif
+            // Issue #1112
+            prev = chunk_get_prev_ncnlnpnd(prev);
```
aneesh1127: i feel something wrong here, we are trying to skip words before DC_MEMBER, so why is DC_MEMBER being skiped?
i ran the new code with "uncrustify::name::func();" this line becomes a FUNC_PROTO instead of FUNC_CALL.

guy-maurel: I cannot reproduce.
Please push your configuration file and the output with
uncrustify ... -p p.txt

PR #1387, opened 2017-10-22
Fix test_cli_options.sh

CDanU:
__E:__
Setting the working directory to the script directory caused that wrong diff
paths where printed if the script was not run inside the script directory.

To fix this a perl command is used to generate a relative path from the
(initial) working directory to the script directory. This relative path is
attached to paths that are relative to the script directory.

As end result now pahts are printed that are relative to the (initial) working
directory, instead of the script directory.


mihaipopescu:
What's wrong with it? Can you add something in description? Thanks!

gmaurel:
Thanks

PR #1390, opened 2017-10-24
add option_reducer.py

CDanU:


mihaipopescu:
Glad to see this being added. Did you saw my changes https://gist.github.com/mihaipopescu/64c667fab8b8c36a462ac3d71d7abe08/revisions or https://github.com/Unity-Technologies/uncrustify/blob/master/scripts/option_reducer.py ?

CDanU:
Yes, I took some ideas from the first link and noticed that you added the script into the unity fork.

The script in this PR now additionally adds options back if they where needed but got removed. There are cases where options can be removed one by one without problem, but problems occur when two or more specific options are removed together.
A similar problem prevents the removal of options in on run, so I added a loop that applies the removal multiple times.
The shortened configs run through, but in some cases new instabilities are generated because of the changes.

mihaipopescu:
Nice you sorted that out and I'm glad to see this in. Nice work!

I've added another script in Unity repository to easily add new tests https://github.com/Unity-Technologies/uncrustify/blob/master/scripts/new_test.py. It assumes you want no change after formatting but it can be addapted.

For me, I find this to be the pipeline of submitting new bug reports.

PR #1394, opened 2017-10-26
Fix out-of-bounds memory write

gaoethan:
The following code merely checks length greater than 200:

   ` if (length > BUFFERLENGTH)`

that means the `length `may equal to 200. so it will overrun
array "`buf`" of 200 bytes at byte offset 200 for the following
code when `length `is 200 and results in unexpected issue:

```
    memcpy(buf, fmt, length);
    buf[length] = 0;

```
Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>


gmaurel:
Thanks

PR #1396, opened 2017-10-26
suppress invalid compilation warning in gcc < v5 (fixes #1098)

### LINKED ISSUE #1098: #1085 introduces a warning
Using the compiler option -Wmissing-field-initializers, I get a warning:
```
/home/guy/Software/uncrustify/src/uncrustify.cpp:424:34: warning: missing initializer for member ‘stat::st_dev’ [-Wmissing-field-initializers]
          struct stat tmp_stat = {};
```
@CDanU could you a look to it?

waldyrious:
The warning is a bug in GCC itself, per https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55805,
so hiding it prevents people using older versions of GCC from assuming there's an issue with uncrustify,
as happened in https://github.com/uncrustify/uncrustify/issues/1098.

I confirmed this change to be working as expected with the following versions of GCC and CMake:
- GCC v4.8.5 / CMake v3.5.1
- GCC v4.8.4 / CMake v2.8.12.2
- GCC v4.7.3 / CMake v3.9.4
- GCC v4.7.3 / CMake v2.8.7


gmaurel:
Thanks

PR #1399, opened 2017-10-27
distinguish "indent" vs. "ident" vs. "identifier"

waldyrious:
There were a few typos where "ident" was used where "indent" was intended; additionally, there was a use of "ident" as an actual token that differed from other instances of the term, which were abbreviations of "identifier". These changes hopefully clarify the distinctions among these three cases.

I ran the tests locally (especially to confirm that the files `tests/output/d/*-Lexer.d` remained valid), and they succeeded (`100% tests passed, 0 tests failed out of 794`). I don't know D, though, so I'd appreciate if someone versed in the language could confirm the validity of the changes in the `*.d` test files.

### DIFF
--- a/documentation/htdocs/index.html
+++ b/documentation/htdocs/index.html
@@ -44,7 +44,7 @@ Create a highly configurable, easily modifiable source code beautifier.</p>
 
 <h2>Features</h2>
 <ul>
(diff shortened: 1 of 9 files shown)


gmaurel:
Thanks

PR #1406, opened 2017-10-31
Support multiple MSVC versions on AppVeyor

jibsen:
Instead of using the build matrix for debug/release, this uses it to set the generator CMake uses. This could also allow building with mingw-w64 and clang on AppVeyor if we want.

I am opening a PR to get some feedback from the CI.


jibsen:
Somewhat surprisingly that appeared to work.

gmaurel:
Thanks

PR #1407, opened 2017-10-31
Fix 'override' as CT_QUALIFIER for LANG_CPP

KVic:


gmaurel:
Could you please provide a test for it?

KVic:
I added test files, but I didn't find relevant test cases for `override`. There was a bug when detecting `braced-init-list` in the statement `void bar () override {}`, but it was fixed.

gmaurel:
Thanks

PR #1408, opened 2017-10-31
Fix the detection of lambda-expression without lambda-declarator

KVic:
Lambda-declarator '( params )' in the lambda-expression is optional.
The minimum lambda-expression can have the form: [captures] {body}.


gmaurel:
Thanks

Issue #1236, opened 2017-07-11
__asm__ blocks indentation


mihaipopescu:
Asm blocks have a few ways of being opened and closed depending on the compiler. Detect and ignore the contents, including alignment.

Workaround: can always fall back on disable/enable_processing_cmt.

Expected versus output diff:
---
```diff
diff --git "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\10054-UNI-1344.cpp" "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\10054-UNI-1344.cpp"
index bda6487..f8af7aa 100644
--- "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\10054-UNI-1344.cpp"
+++ "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\10054-UNI-1344.cpp"
@@ -1,20 +1,20 @@
 // Asm blocks have their own special indentation where lables must remain at indent 0 relative to __asm__ block.
 // They few ways of being opened and closed depending on the compiler.
 // For now, we can at least detect and ignore the contents, including alignment.
 
 // Workaround: can always fall back on disable/enable_processing_cmt.
 
 void foo()
 {
     int head, bar;
     __asm__ __volatile__
     (
         "movq %0,%%xmm0\n\t"    /* asm template */
-    "0:\n\t"
-        "bar    %0, [%4]\n\t"    // in template
-    "1:\n\t"
+        "0:\n\t"
+        "bar\t%0, [%4]\n\t" // in template
+        "1:\n\t"
         : "=a", (bar)
         : "=&b", (&head), "+m", (bar)
         : "cc"
     );
 }
```

Debug support files:
---
[10054-UNI-1344.cpp](https://gist.githubusercontent.com/mihaipopescu/6a04c29dc9725a8465afb3ea59f76e4f/raw/bda64870a3d6e1cec07e2fa4606d8bd5094a0b94/10054-UNI-1344.cpp)
[10054-UNI-1344.cpp.log](https://gist.githubusercontent.com/mihaipopescu/6a04c29dc9725a8465afb3ea59f76e4f/raw/f7cf2084742709e5ee2484f19fb8a6b450d88e02/10054-UNI-1344.cpp.log)
[...]

PR #1419, opened 2017-11-05
Adopt the method from aneesh1127 for #1236

guy-maurel:
Thanks for the proposal.


CDanU:
It would be nice if you could use gits _cherry picking_ functionality to keep authorship information. 

On the other hand it would be far better if @aneesh1127, @puppalasatish, @ KalyanaChakravarthi, @satishp87 and everyone else that creates commits for the unity fork would also create PRs here.

Also give the [option_reducer.py](https://github.com/uncrustify/uncrustify/blob/master/scripts/option_reducer.py) script a try to remove useless options from the configs.

guy-maurel:
Right!
My python is under /usr/bin. Could you modify script option_reducer.py ?

PR #1421, opened 2017-11-07
option_reducer.py fix python path

CDanU:


scripts/option_reducer.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/python
+#!/usr/bin/python
```
mihaipopescu: less flexible than `#!/usr/bin/env python`

PR #1441, opened 2017-11-13
Fix indent_func_const.

kraiskil:
Without this fix, the 'indent_func_const' configuration is used
as an absolute value, not as a relative to the current indentation
level.

The C++ test 30270 has in its configuration
'indent_func_const = 69',
yet the outputs 'const' started on column 69, i.e. 68 spaces indent.
The error is even more noticable when the 'const' is part of a class
member function, already indented. This appends the 30270 test case
with such an example.


gmaurel:
Thanks

PR #1444, opened 2017-11-15
More control pse tos

guy-maurel:
The frame_count variable must be controlled.

### DIFF
--- a/src/controlPSECount.cpp
+++ b/src/controlPSECount.cpp
@@ -1,0 +1,20 @@
+/**
+ * @file controlPSECount.cpp
+ *
+ * @author  Guy Maurel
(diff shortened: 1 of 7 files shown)


src/controlPSECount.cpp
```diff
@@ -0,0 +1,20 @@
...
+ * @file controlPSECount.cpp
+ *
+ * @author  Guy Maurel
+ * @license GPL v2+
+ */
+
+#include "controlPSECount.h"
+
+
+void controlPSECount(size_t value)
```
CDanU: Do you intend to use that function in other files than indent.cpp? If not this could be just a static function.

PR #1451, opened 2017-11-23
Test a better error handling.

guy-maurel:
It should produce a larger error output for the first cs-test


guy-maurel:
How to solve the problem?

Issue #1405, opened 2017-10-30
C++ list initialization from a braced-init-list


KVic:
[_List-initialization_](http://en.cppreference.com/w/cpp/language/list_initialization) is initialization of an object or reference from a _braced-init-list_ (C++11 "8.5 Initializers", C++17 "11.6 Initializers") . This is the [preferred syntax](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax) for initialization. Therefore, the Uncrustify should pay more attention to a _braced-init-list_.

I think that the _braced_init_list_ should have its own parentheses chunk type: `CT_IL_BRACE_OPEN` and `CT_IL_BRACE_CLOSE` (IL - initialization list), for example. `braced_init_list` is more like a function call with arguments, than the curly braces that determine the structure of the source code. `braced_init_list` should have its own options (now it's `xx_type_brace_init_lst_yyy`) or refer to parentheses of the function call.

I added a few examples to the [tests/input/cpp/braced_init_list.cpp](https://github.com/KVic/uncrustify/blob/81ce3ff37dca0ef7d4d637ace444a3782a9673bc/tests/input/cpp/braced_init_list.cpp) to get started.

gmaurel:
It would be fine

kintel:
@KVic I tested your latest additions in #1404, and found this snipped of code to sneak past your checks for braced init lists:
```
class ClassA : ClassB {
public:
ClassA() : ClassB(ClassC{1.0}) {
}
};
```

-> It looks like `ClassC{1.0}` doesn't get parsed correctly and gets an extra space: `ClassC {1.0}`.

kintel:
FYI: With the latest patches applied, this runs correctly for my entire codebase.

KVic:
@kintel, thank you for the information and testing.

PR #1468, opened 2017-11-30
Fix the nested braced-init-list in the constructor

KVic:
Fixes the initialization of the class member with the nested braced-init-list in the constructor:
```
class Base {
Base() : a{int{1}}{}
int a;
};
```
Issue #1405


gmaurel:
Thanks

PR #1469, opened 2017-12-01
 Fix newline before a closing brace in a braced-init-list

KVic:
braced-init-list is more like a function call with arguments, than curly braces that determine a structure of a source code, so, don't add a newline before a closing brace.

Issue #1405.


gmaurel:
Thanks

PR #1470, opened 2017-12-01
 Fix 'if constexpr' C++17 feature

KVic:
'if constexpr' and 'else if constexpr' are marked as valid statements.


gmaurel:
Thanks

PR #1491, opened 2017-12-10
Adopt UT 3484

guy-maurel:


CDanU:
It would be good to have references to the issues/pull requests form our and/or  the unity repository.

tests/config/U10-Cs.cfg
```diff
@@ -0,0 +1,1942 @@
+### This file holds rules specific to C#
```
CDanU: Seems like you forget to run this through the option_reducer script

PR #1495, opened 2017-12-13
Adopt UT test UNI-2650.

guy-maurel:
Thanks to Mihai

### DIFF
--- a/src/controlPSECount.h
+++ b/src/controlPSECount.h
@@ -11,6 +11,8 @@
 #include "uncrustify_types.h"
 
 
+void controlPSECountMinus(size_t value);
+
 void controlPSECount(size_t value);
 
 #endif /* CONTROLPSECOUNT_H_INCLUDED */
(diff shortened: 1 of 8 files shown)


src/controlPSECount.h
```diff
@@ -11,6 +11,8 @@
 #include "uncrustify_types.h"
 
 
+void controlPSECountMinus(size_t value);
+
 void controlPSECount(size_t value);
```
CDanU: Why do we need both of those functions scattered throughout the `indent.cpp` file? The `frm.pse_tos` var is only changed in the pop and push functions. The checks should be done there. If we need more control than that we should make `frm` a class with a private `pse_tos`.

guy-maurel: No. The variable frm.pse_tos might get the value zero. And zero - 1 is "too big"

src/indent.cpp
```diff
@@ -1806,6 +1824,7 @@ void indent_text(void)
          {
             // Skip any continuation indents
             controlPSECount(frm.pse_tos - 1);
+            controlPSECountMinus(frm.pse_tos);
             idx = frm.pse_tos - 1;
```
CDanU: That one for example.

PR #1499, opened 2017-12-15
Test

CDanU:


src/indent.cpp
```diff
@@ -618,35 +618,25 @@ static void _log_indent_tmp(const char *func, const uint32_t line, parse_frame_t
...
 
    // dummy top-level entry
+   parse_frame_t frm{};
    frm.pse[0].indent     = 1;
    frm.pse[0].indent_tmp = 1;
    frm.pse[0].indent_tab = 1;
    frm.pse[0].type       = CT_EOF;
 
-   pc = chunk_get_head();
+   auto pc = chunk_get_head();
```
KVic: I think there is no need to force auto against simple types. In such cases, "auto" has no advantages, but it makes it difficult to read and refactor the code.

PR #1500, opened 2017-12-15
Fix 'mod_full_brace_if_chain' with braced statements

KVic:
The `mod_full_brace_if_chain = true` option modifies the code:
```
if ( a ) {
    {a = 5;}
    {b = 5;}
}
```
to:
```
if ( a )
    {a = 5;}
    {b = 5;}
```
which is incorrect. This PR fixes it.


gmaurel:
Thanks.

PR #1532, opened 2018-01-06
Adopt UT test 60040

guy-maurel:


tests/config/UNI-30498_2.cfg
```diff
@@ -0,0 +1,1939 @@
+### This file holds rules common to all c-style languages (currently C# and C++)
```
CDanU: Needs to be reduced.

PR #1535, opened 2018-01-08
Adopt ut crash handling

guy-maurel:


src/uncrustify.cpp
```diff
@@ -2052,6 +2104,7 @@ static lang_name_t language_names[] =
    { "PAWN", LANG_PAWN          },
    { "OC",   LANG_OC            },
    { "OC+",  LANG_OC | LANG_CPP },
+   { "CS+",  LANG_CS | LANG_CPP },
```
CDanU: Is this intended to be in this PR?

mihaipopescu: You are right. Should not be part of this PR. It is a special c++/c# language used at Unity.

PR #1545, opened 2018-01-12
The logger has not enough memory to store a long message,

guy-maurel:
which will be unproperly cut.
An fatal error message will be produced.


mihaipopescu:
I would love to see a cli test for this.

PR #1546, opened 2018-01-12
Adopt UT test 10051

guy-maurel:
Thanks to Kalyana Chakravarthi

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -1229,8 +1229,9 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
       {
          chunk_t *tmp = chunk_get_prev_type(first, first->parent_type, first->level);
          tmp = chunk_get_prev_ncnl(tmp);
-         if (tmp != NULL && tmp->type == CT_NEW)
+         if (chunk_is_token(tmp, CT_NEW))
          {
+            log_rule("sp_fparen_brace_initializer");
             return(cpd.settings[UO_sp_fparen_brace_initializer].a);
          }
       }
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -1223,6 +1223,17 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
          log_rule("sp_fparen_dbrace");
          return(cpd.settings[UO_sp_fparen_dbrace].a);
       }
+      // To fix issue #1234
+      // check for initializers and add space or ignore based on the option.
+      if (first->parent_type == CT_FUNC_CALL)
+      {
+         chunk_t *tmp = chunk_get_prev_type(first, first->parent_type, first->level);
+         tmp = chunk_get_prev_ncnl(tmp);
+         if (tmp != NULL && tmp->type == CT_NEW)
+         {
+            return(cpd.settings[UO_sp_fparen_brace_initializer].a);
+         }
+      }
       log_rule("sp_fparen_brace");
       return(cpd.settings[UO_sp_fparen_brace].a);
    }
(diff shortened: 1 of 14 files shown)


src/space.cpp
```diff
@@ -1223,6 +1223,17 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
          log_rule("sp_fparen_dbrace");
          return(cpd.settings[UO_sp_fparen_dbrace].a);
       }
+      // To fix issue #1234
+      // check for initializers and add space or ignore based on the option.
+      if (first->parent_type == CT_FUNC_CALL)
+      {
+         chunk_t *tmp = chunk_get_prev_type(first, first->parent_type, first->level);
+         tmp = chunk_get_prev_ncnl(tmp);
+         if (tmp != NULL && tmp->type == CT_NEW)
```
CDanU: use chunk_is_token() for the CT_NEW check

tests/config/UNI-1338.cfg
```diff
@@ -0,0 +1,8 @@
+sp_fparen_brace                 = add
+indent_columns                  = 4
+indent_with_tabs                = 0
+indent_class                    = true
+nl_class_leave_one_liners       = true
+nl_end_of_file                  = force
+nl_end_of_file_min              = 1
+sp_inside_braces                = add
```
CDanU: This test (config, input, output) does not seem to belong to this PR. The new option is missing here.

tests/c-sharp.test
```diff
@@ -83,8 +83,9 @@
 10036  U14-Cpp.cfg                          cs/inttypes.h.mm
 10039  615_nested_usings.cfg                cs/615_nested-usings.cs
 10044  U04-Cs.cfg                           cs/ifcomment.cs
-10045  U05-Cs.cfg                           cs/UNI-1288.cs
+10045  UNI-1288.cfg                         cs/UNI-1288.cs
 10046  U15-Cpp.cfg                          cs/UNI-1333.mm
+10051  UNI-1338.cfg                         cs/UNI-1338.cs
```
CDanU: ↑

src/space.cpp
```diff
@@ -1223,6 +1223,17 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
...
+         {
+            return(cpd.settings[UO_sp_fparen_brace_initializer].a);
```
CDanU: `log_rule("sp_fparen_brace_initializer") ;` missing

tests/config/UNI-1288.cfg
```diff
@@ -0,0 +1,9 @@
+sp_assign                       = add
+sp_fparen_brace_initializer     = add
```
CDanU: This config does have too many options in order to only check `sp_fparen_brace_initializer`

guy-maurel: It use the default value of the new option.

CDanU: I am not entirely sure what you mean by that. 
I wanted to point out that most of the  options are not sql to test the behavior of the new option, esp. options like: 

```sql
```

are a big annoyance that is spreading throughout almost all tests. We should maybe consider to set those to default values in Uncrustify itself.  
I have never seen anyone that uses the current 8 space default indent (and spaces are '__*obviously*__' better than tabs).

mihaipopescu: As much as I would like those to be the new defaults changing them it's going to break backwards compat.

gmaurel: I think we may not change the defaults, even if they have a "wrong" value.

CDanU: OK, we will discuss the default values issue somewhere else. 
You can however change the tests to that the unneeded options aren't used anymore.

PR #1547, opened 2018-01-12
add sp_cpp_before_struct_binding to support C++17 structured binding declarations

nivekkagicom:
C++17 structured binding declarations use square brackets. The option allows

`auto [x, y, z] = f();`

to be formatted with a space following the variable and before the space. This also works for

`auto& [x, y, z] = f();`
`auto const [x, y, z] = f();`
`auto const& [x, y, z] = f();`


src/space.cpp
```diff
@@ -809,6 +809,17 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp, bool comp
       return(cpd.settings[UO_sp_after_oc_msg_receiver].a);
    }
 
+   // c++17 structured bindings e.g., "auto [x, y, z]" vs a[x, y, z]" or "auto const [x, y, z]" vs "auto const[x, y, z]"
+   if (cpd.lang_flags & LANG_CPP
+	  && (first->type == CT_BYREF || first->type == CT_QUALIFIER || first->type == CT_TYPE)
```
CDanU: Thanks for the PR, the build fails because of the source code formatting rules, but it's not you first time so I guess you know what to do. (You will also have to change the CLI files)

Pleas take a look at 3ebb43db21c40bfbb5c4d28541daedd0897897b9, we currently don't really have that information somewhere where people could find it, or any enforcement on that but as you will have do fix the formatting anyways I would like you to also break up those `||` on new lines.

tests/input/cpp/cpp17.cpp
```diff
@@ -0,0 +1,8 @@
+bool CompareGenomeByFeatureResults::clickOnLink(std::string const& inLink) {
+   auto const [sequence, type, firstPosition, lastPosition] = parseLink(inLink);
```
CDanU: Your Test here isn't really doing anything as there is already a space before the '['.

You could add more spaces and use `sp_cpp_before_struct_binding    = force` to remove them. What does `remove` do on this file, are there any cases where it would be useful?

CDanU:
Ok, squash everything down into the first commit and this one is good to go.

CDanU:
I cleaned up your PR a bit and will submit it soon, so hang in tight. That last commit should be in a different PR.

nivekkagicom:
FWIW I was in the process of splitting the pull request (#1553 was 1/2 but now closed); however, I will be patient

gmaurel:
As the dicussion goes on, I reopen

CDanU:
I included everything from this PR in the one that I have opened, the last commit also now has its own PR. This therefore can be closed and the referenced PR merged.

nivekkagicom:
I personally believe this has been replaced by CDanU's reworking of this (my original) PR into his #1552 and my #1650.

PR #1557, opened 2018-01-15
allow test_cli_options.sh to work on more platforms

nivekkagicom:
Add a newline to end of DOC_TEXT_END such that output always ends with a line end and such that the "option(s) with 'not default' value: 0" is not preceded by double hash (##) comment. Also, replace readlink and rmdir on macOS so that the script will run.


gmaurel:
Thanks for it
Nice to have a macOS guy.

PR #1567, opened 2018-01-19
Adopt UT test 60015.

guy-maurel:
Little typo at brace_cleanup.cpp

### DIFF
--- a/tests/output/cs/60015-UNI-14131.cs
+++ b/tests/output/cs/60015-UNI-14131.cs
@@ -1,0 +1,16 @@
+using System.Collections.Generic;
+
+class C
+{
+    public void S()
+    {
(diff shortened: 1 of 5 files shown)


tests/output/cs/60015-UNI-14131.cs
```diff
@@ -0,0 +1,16 @@
...
+    public void S()
+    {
+        Action localMethod = () => {
+                SomeClass.OtherMethod(new Dictionary<string, string>
+            {
+                {"a", "one"},
+                {"b", "two"},
+                {"c", "three"}
+            });
+            };
```
CDanU: @mihaipopescu is this really the output that should be generated? Looks strange to me ...

PR #1568, opened 2018-01-21
apply fix_enum_struct_union even if there is a typdef before an enum

CDanU:
The `fix_enum_struct_union` function does set the `PCF_IN_ENUM` flag which is needed
for some config options related to enums.

The `sp_enum_assign = force` option had to be added to the rdan.cfg because
before this commit the enum flags where missing so that `sp_assign` was applied.
With the flags in place now `sp_assign`is not used any more and
`sp_enum_assign` is needed to recreate the expected formatting.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -1058,10 +1058,7 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
       || pc->type == CT_STRUCT
       || pc->type == CT_UNION)
    {
-      if (prev->type != CT_TYPEDEF)
-      {
-         fix_enum_struct_union(pc);
-      }
+      fix_enum_struct_union(pc);
    }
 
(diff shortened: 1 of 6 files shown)


src/combine.cpp
```diff
@@ -1058,10 +1058,7 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
       || pc->type == CT_STRUCT
       || pc->type == CT_UNION)
    {
-      if (prev->type != CT_TYPEDEF)
```
mihaipopescu: Do you know why this was introduced in the first place?

CDanU: No, this goes as far back as cb24cf1487844dbd63a010571dbb7266c7417b9a, but there are no clues about why this was introduced. According to the commit message this commit fixes a bug, but the change set is rather big so I am not sure what exactly got fixed.

Issue #1562, opened 2018-01-17
Migrate html documentation to wiki


albfan:
uncrustify 0.63

http://uncrustify.sourceforge.net/options.html is not linked to any anchor in http://uncrustify.sourceforge.net/index.html and http://uncrustify.sourceforge.net/options_NewLines.html#nl_after_access_spec do not link to documentation.

Given that `--show-config-with-doc` documents every property, is it possible to migrate all info into a wiki? I find it really useful

CDanU:
Thanks for the info about the missing links. 
I am nor entirely sure if we actually still need Sourceforge. GitHub pretty much can do everything Sourceforge does. 

About the `--show-config-with-doc`: As far as I know the wiki is open to everybody(?) I wouldn't however advice to manually add the output in there. It is bound to get out of date that way and will cause more confusion that having it currently only inside the application itself.

albfan:
I think too sourceforge is not the way anymore, but is not bad to fix there too. Is there a GitHub Page for uncrustify?

Yes, wiki is for everyone, but I find people respectful, and little scared to edit there (but don't known for a project with this traction). Maybe https://hotdoc.github.io/installing.html is an option. I think a good example is meson.  It starts with GitHub wiki and finally migrated all to hotdoc

https://readthedocs.org which uses sphinx is a even better option

I can help to migrate if you want to do so

gmaurel:
Sourceforge is historical.
github is pretty fine to hold the code and start some tests.
Sourceforge is used to hold binaries.
Any help to complete the documentation is welcome.

CDanU:
Github can also hold binaries. If we decide to switch we would need to do that gradually and we would need to notify all of the distro package owners. 

------------------

No we don't have an github.io page (or for that matter a fancy logo) yet. 

I don't particularly like that the hotdoc generated documentations seem to rely on JS, I do however like readthedocs. How good is sphinx at supporting C++ nowadays?

We started to discuss this a bit in #875. I am not sure if we need a full blown static page, but as mentioned we need automation. 
We could write some scripts that are triggered by Travis to update the wiki but that seems be be too much work. Maybe readthedocs could reduce that.

albfan:
All sounds D's great. Let's start small.

- Make all the docs reachable from index
- Expose on github wiki
- Script something to generate docs from existing options like `--show-config`

Futures:
- Automate with Travis
- Link to github releases

albfan:
A first quick migration to markdown using pandoc:

[...]

PR #1575, opened 2018-01-23
Add options to htdocs

albfan:
relates with #1562

### DIFF
--- a/documentation/htdocs/index.html
+++ b/documentation/htdocs/index.html
@@ -57,11 +57,17 @@ Create a highly configurable, easily modifiable source code beautifier.</p>
    <li>Highly configurable - 608 configurable options as of version 0.66</li>
 </ul>
 
+<a href="uncrustify.html">What it does?</a>
+
 <p>
    See some example <a href="examples/example.c">output</a>.
 </p>
 
 <br />
+
+<h2>Configuration</h2>
+
+See avaliable <a href="options.html">Options</a>
 
 <h2>Where to get Uncrustify</h2>
 


documentation/htdocs/index.html
```diff
@@ -57,12 +57,18 @@ <h2>Features</h2>
    <li>Highly configurable - 608 configurable options as of version 0.66</li>
 </ul>
 
+<a href="uncrustify.html">What it does?</a>
+
 <p>
```
guy-maurel: I wish that the related file to be as current as possible, in the same PR

albfan: This was more to show unlinked files (maybe just to delete them) 

https://github.com/uncrustify/uncrustify/blob/master/documentation/htdocs/uncrustify.html
> I'm not sure if I can describe all things uncrustify does at this time. Let me give it a try!

Then I will try to link other txt files there too

gmaurel: I wait until you are ready

albfan: I really try, but don't know what uncrustify support or not at this time:

This is the content of that file (esentially):

What it does:

        - Indent code
	- Align code on '=' and variable definitions
	- Align #define stuff
	- Align backslash-newline stuff
	- Reformat comments (a little bit)
	- Fix inter-character spacing

What it doesn't do (yet):
	- Split long lines
	- Handle format changing #if/#else stuff

Please provide feedback on this

CDanU: Thanks for linking the two files, I looked on both of the and they are really outdated (uncrustify.html) or not exactly something that I would want to link right now (options.html) 

I am not entirely sure what you are doing now but please don't try to describe all options on you own. That is a way too big target that should be done gradually.

I think we should discuss that again in #1562 and close this PR.

albfan:
Yes, exactlyI was pointing this exactly. Let's close this and work o. That issue

PR #1598, opened 2018-02-07
Show actual difference of files rather than asking user to find differences

JavDevGames:
Instead of reporting that there are differences in files and asking the users to look find the files, the script shows the actual differences in the files now


mihaipopescu:
@CDanU I have two questions:
1. Why do we still have the bash script around? Can't we just kill it and update `.travis.yml`?
2. The python script could take an argument to apply the results to the output folder. What do you think? You were saying you would like something similar once but I guess I'm not up-to-date.

gmaurel:
The information might become very large if a little format errors where done.
It will pollute the output. The same information is available with the command `diff` to the 
developer.

CDanU:
I do agree with gmaurel here, printing a diff for every file will be messy. I guess you could add an optional script arg if you want to have that feature.

--------

@mihaipopescu 
1. yes that should be possible, I just have to update the script again, gmaurel made some changes to the .sh script.
2. Yes I mentioned that, I still haven't found an easy, free, and secure (not reveling ssh keys) service that could trigger a python script and push the created changes back into a repository.
I guess I have to search for a free vhost of some kind an put a the infrastructure on that.

mihaipopescu:
I agree with the new option to print the diff.
1. Nice. Keeping both scripts in sync is going to be a pain anyway.
2. There is a concept of a secret variable in Gitlab. Unsure about github. Maybe Travis has this but I'm pretty sure it's a solved problem.

I guess we do hire grads. Why do you ask?

JavDevGames:
@gmaurel I agree, defaulting to always show differences was not the best approach. I've added a command line option allow for displaying diffs if the developer decides to. 

With regard to the auto-push of the changes, I'll investigate this and report back with what I can find.

JavDevGames:
@CDanU I've added the -apply flag, to allow for overwriting the files in the output folder if specified.

tests/cli/test_cli_options.py
```diff
@@ -136,6 +148,8 @@ def check_generated_output(gen_expected_path, gen_result_path, result_manip=None
     True
     """
 
+    no_additional_options = bool(program_args.diff is False and program_args.apply is False)
```
mihaipopescu: you can remove this check and replace `if no_additional_options:` with `if not program_args:`

tests/cli/test_cli_options.py
```diff
@@ -70,6 +72,13 @@ def proc(bin_path, args_arr=()):
 
     return out_txt, err_txt
 
+def write_to_output_path(output_path, result_str):
+    """
+    writes the contents of result_str to the output path
+    """
+    print("Auto appending differences to: " + output_path)
+    with open(output_path, 'w', encoding="utf-8", newline="") as f:
```
mihaipopescu: why `newlin""` ?

tests/cli/test_cli_options.py
```diff
@@ -441,6 +492,12 @@ def main(args):
     # set working dir to script dir
     sc_dir = dirname(relpath(__file__))
 
+    parser = argparse.ArgumentParser(description='Test CLI Options')
+    parser.add_argument('-diff', help='show diffs when there is a test mismatch', action='store_true')
+    parser.add_argument('-apply', help='auto apply the changes from the results folder to the output folder', action='store_true')
```
mihaipopescu: Usually, arguments are using their short notation `-d` or a complete word but with two dashes `--diff`. I haven't seen this mixture. You can also use this reference http://www.catb.org/esr/writings/taoup/html/ch10s05.html if you want to add short options that are commonly used and preserve expectations.

tests/cli/test_cli_options.py
```diff
@@ -441,6 +492,12 @@ def main(args):
...
+
+    parsed_args = parser.parse_args();
```
mihaipopescu: remove `;`

gmaurel:
@JavDevGames Please take care of the last comments (@mihaipopescu )

JavDevGames:
Updated the code to reflect the changes requested by @mihaipopescu 

When looking specifically at the newlines="" we discovered that there was some unintended behavior happening when it came to writing files out.

To explain a bit further, here are the scenarios we found:
newline = None: this outputs  \r\n
newline = "\r": this outputs  \r
newline = "\n": this outputs  \n
newline = ""  : this outputs  \n    
For the sake of consistency, all newlines are now being written out as \n
However, if the result_str itself contains \r\n, then \r\n will be output
as this code doesn't post process the data being written out

This means that the need for the post-processing of the text (the manipulator function that removed the \r) was no longer necessary and was thus removed.

However, in order to ensure the tests were accurate I edited the .gitattributes file to not change the line endings via the -crlf setting.

There was also an improvement done to the regex that was removing the #Uncrustify-xxx.xxx tag at the start of the file, which was inadvertently removing the very first \r from the p.txt Output file.

All these changes can be seen in the latest commit.

Please let me know if this raises any concerns and I will address them.

JavDevGames:
I'm looking into this.
It appears that there is an issue with p.txt in the sense that specifying:
```newlines = crlf" inside of mini_nd.cfg is still returning a string with only LF

**Comment by CDanU** [comment]
@JavDevGames despite me saying that I will check your code I still haven't found the time to do that. 
The PR got a bit messy in the mean time, do you know how to clean that up (interactive rebase)?
If not I will do that myself (sometime, hopefully soon) so I would ask @gmaurel to **NOT** merge this PR in the current state, even if the changes might be OK.

**Comment by JavDevGames** [comment]
Hey @CDanU. Thanks for getting back!
I'll clean up the PR. A lot of the debugging stuff was done because I couldn't get travis setup locally so that added a lot of noise. What I feel that we did achieve was more consistency in the files (removed all \r post processing of text, etc.).
It did grow beyond what I originally intended...but that's what happens when one starts messing with CRLF 
I'll fix it in the next few days and let you know once its ready!

**Comment by JavDevGames** [comment]
@CDanU does this look alright now?
I've dropped and squashed a lot of the commits to hopefully make each step of the way a little bit clearer.

**Comment by CDanU** [comment]
Yes this has gotten quiet a bit cleaner, but there are still comments that could be cleaned up. 

I kinda don't know how persistent I want to be, but personally I'd always make sure that nobody sees my mistakes in order to get the history as readable as possible. 

Currently there are changes that get removed some commits later and there is a duplicate of the test script that should be removed.

File: .travis.yml [hunk 5]
```diff
@@ -88,5 +88,5 @@ script:
     # test if all the sources are conform to the forUncrustifySources.cfg config file
   - if [ "x${BUILD_TYPE}" == "xrelease" ]; then cd ../; ./scripts/Run_uncrustify_for_sources.sh; fi
     # test if uncrustify runs for (some) command line options
-  - if [ "x${BUILD_TYPE}" == "xrelease" ]; then ./tests/cli/test_cli_options.sh; fi
+  - if [ "x${BUILD_TYPE}" == "xrelease" ]; then cd tests/cli; ./test_options_cli.py; fi
```
CDanU: Have you tested that? It looks like this should fail as you are in the `uncrustify/builld` dir (lines 57 & 89).
If you actually need that cd, I'd expect that you jump back (`cd -`) in the end. The same applies for line 89.

JavDevGames:
Thanks for looking at it. I'll fix it up and update you when I've addressed the issues you mentioned!

CDanU:
I decided that I am taking this over, I submit a cleaned up PR based on this tonight/tomorrow.

PR #1621, opened 2018-02-15
Fix invalid alignment of the right comments

KVic:
The align_right_cmt_span option adds extra spaces before the right comments at a large level of nested braces.

Config:
`align_right_cmt_span = 3`

Source:
```
namespace A
{
namespace B
{
namespace C
{


struct D
{
	int a; // a.
	int b;
	int c;
}; // struct D


} // namespace C


struct E {};


} // namespace B


struct F {};


} // namespace C
```
Result:
```
namespace A
{
namespace B
{
namespace C
{


struct D
{
	int a;                  // a.
	int b;
	int c;
};                      // struct D


}               // namespace C


struct E {};


}       // namespace B


struct F {};


} // namespace C
```
Expected:
```
namespace A
{
namespace B
{
namespace C
{


struct D
{
	int a; // a.
	int b;
	int c;
}; // struct D


} // namespace C


struct E {};


} // namespace B


struct F {};


} // namespace C
```

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -1538,17 +1538,12 @@ chunk_t *align_trailing_comments(chunk_t *start)
 
          if (cmt_type_cur == cmt_type_start)
          {
-            col = 1 + (pc->brace_level * cpd.settings[UO_indent_columns].u);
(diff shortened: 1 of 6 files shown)


gmaurel:
Thanks

PR #1627, opened 2018-02-17
The use of CT_IGNORE_CONTENT is not necessary.

guy-maurel:
Ref. #1236 and https://github.com/Unity-Technologies/uncrustify/issues/102

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1684,26 +1684,6 @@ void indent_text(void)
             }
          }
       }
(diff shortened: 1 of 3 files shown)


aneesh1127:
please refer Config file, We set the Token in such a way, by default nothing is tokenized as CT_IGNORE_CONTENT, unless set in config files.
PR: Unity-Technologies#24, please check the tests added, and files changed

PR #1639, opened 2018-02-23
Add basic processing of @available() syntax in objc

Krizai:
Implemented basic processing of the @available() syntax in objc. Just don't modify anything inside the brackets, not to break the things.


gmaurel:
Thanks

PR #1656, opened 2018-03-05
sun.cfg: nl_func_var_def_blk = 1

coeo:
Fix error: "The option 'nl_func_var_def_blk' is too big against the option 'nl_max'"

Per: http://www.oracle.com/technetwork/java/codeconventions-150003.pdf page 14
"One blank line should always be used in the following circumstances:
Between the local variables in a method and its first statement"

### DIFF
--- a/etc/sun.cfg
+++ b/etc/sun.cfg
@@ -1046,7 +1046,7 @@ nl_after_square_assign                    = remove   # ignore/add/remove/force
 
 # The number of blank lines after a block of variable definitions at the top of a function body
 # 0 = No change (default)
-nl_func_var_def_blk                       = 2        # number
+nl_func_var_def_blk                       = 1        # number
 
 # The number of newlines before a block of typedefs
 # 0 = No change (default)


guy-maurel:
The value 2 means "one empty line", exactly what you wish.
The warning _... is too big against ..._ means that the two options do not work together.
The user should change one of the two values.

coeo:
> The value 2 means "one empty line", exactly what you wish.

**No**, it doesn't. The value 1 means "one empty line".

`$ grep -B 2 nl_func_var_def_blk cfg/sun.cfg
# The number of blank lines after a block of variable definitions at the top of a function body
# 0 = No change (default)
nl_func_var_def_blk                       = 1        # number`

`$ cat NewLines.java
public class NewLines {
        public static void main(String[] args) {
                int x = 4;
                System.out.println(x);
        }
}`

`$ ./uncrustify.exe -c cfg/sun.cfg -f NewLines.java -o NewLines.java.new
Parsing: NewLines.java as language JAVA`

`$ cat NewLines.java.new
public class NewLines {
    public static void main(String[] args) {
        int x = 4;

        System.out.println(x);
    }
}`

**1 blank line!**

>The warning ... is too big against ... means that the two options do not work together.

Yes, which is why I am submitting the fix.

It was broken in [commit 46b366a](https://github.com/uncrustify/uncrustify/commit/46b366ae18d144e7ec259ac14e876892faf12d32#diff-7fc989c6ae46452648aef6b78a04fbb9)

> The user should change one of the two values.

You're calming that one of your example configurations is intentionally broken?

guy-maurel:
Sorry, my error:

1. I was confused with _nl_count_ at the _newline_-token and the description of the option _nl_func_var_def_blk_ .
2. the example _sun.cfg_ is historical (not intentionally broken)! Thank for the message.

Issue #1400, opened 2017-10-27
segmentation fault with malformed input file


waldyrious:
Any file ending with a `#`, without newline at the end, crashes Uncrustify with a segmentation fault (found by experimenting with [afl-fuzz](http://lcamtuf.coredump.cx/afl/)).

Reproduction steps:

    $ uncrustify -v
    Uncrustify-0.65-270-bc41097
    $ echo -n '#' > crash.c
    $ uncrustify -c etc/defaults.cfg crash.c
    Output suffix: .uncrustify
    Parsing: crash.c as language C
    Segmentation fault

(calling with `-p debug.txt` produces no debug.txt file)

Edit: this also happens if the `#` is followed by whitespace (` `, `\v`, `\t`), but not `\n` or `\r`. If the line contains any non-whitespace character, there's no crash.

waldyrious:
Two other simple patterns found by afl-fuzz:

- Any file starting echo `&&` followed by anything other than whitespace (`echo '&&a' > crash.c`)
- Any file starting with `<<`, regardless of what follows (`echo '<<' > crash.c`)

CDanU:
Unfortunately I believe that should be a significant amount of similar issues. Some errors can be easily found with clang tools, fuzzying should find even more. 

It would be nice if you could provide some execution scripts or a short writeup on how to use afl-fuzz.

waldyrious:
> It would be nice if you could provide some execution scripts or a short writeup on how to use afl-fuzz.

That was my plan, yeah :) I have some assorted notes that, once cleaned up, I'll submit as a [tldr-pages](https://github.com/tldr-pages/tldr) entry. It may take a few days or weeks, depending on available time.

jibsen:
Just for reference:
  - `#` crashes at indent.cpp:391 with pc=0
  - `&&a` crashes at tokenize_cleanup.cpp:844 with prev=0
  - `<<` crashes at align_stack.cpp:180 with ref=0

waldyrious:
For convenience:

> `#` crashes at indent.cpp:391 with pc=0

https://github.com/uncrustify/uncrustify/blob/5194333f3c8fbfd949375f3ac1d7eebd3abe3949/src/indent.cpp#L391-L392

> `&&a` crashes at tokenize_cleanup.cpp:844 with prev=0

[...]

PR #1659, opened 2018-03-06
Less crashes

guy-maurel:
Ref. #1400 
Thanks for the message

### DIFF
--- a/src/tokenize_cleanup.cpp
+++ b/src/tokenize_cleanup.cpp
@@ -68,8 +68,8 @@ static chunk_t *handle_double_angle_close(chunk_t *pc)
 
    if (next)
    {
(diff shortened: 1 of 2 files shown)


guy-maurel:
#1659 is a bugfix proposal.

src/tokenize_cleanup.cpp
```diff
@@ -859,7 +862,8 @@ void tokenize_cleanup(void)
          && (cpd.lang_flags & LANG_CPP)
          && chunk_is_str(pc, "&&", 2))
       {
-         if (prev->type == CT_TYPE)
+         if (  prev != nullptr
+            && prev->type == CT_TYPE)
```
CDanU: use the `chunk_is_token` function here

Issue #1641, opened 2018-02-26
feature request: Save spaces before newline


Krizai:
Xcode default behaviour is to put spaces according to current indent level in newline-only lines. Uncrustify remove these spaces. It would be nice to add an option to indent such a newlines.

So let's suppose the input is (where _ represents space)
```
- (instancetype)init{
    int i = 1;
____
____
    if (self) {
        i++;
    }

    return self;
}
```

The expected output should have the same amount of spaces, while current actual output will remove all of them

CDanU:
It is unlikely that we will do that. The problem is a fundamental one, apart from newlines we don't save white spaces as chunk/tokens (the amount of spaces/tabs between two chunks is saved within the chunks). They are all gone by the end of the tokenization process. 

To make possible what you want would mean to introduce a new token type and change all the places where we currently only ignore newlines and/or comments (currently we `use chunk_get_next/prev_ncnl`: _non comment, non newline_) to additionally ignore white space tokens, but of course only in the appropriate places. That could either be fairly easy but it is likely hat it is going to be the opposite. 

In the case of leading spaces it might be easier to maybe add the amount of spaces to newline chunks and alter the way they are processed in the output stages (writing to a file) to add them if needed. That also would require some changes in the tokenization so that two newline chunks won't become one newline chunk that acts as both.

You can of course see for yourself and report with what you have found out. If it is feasible we can flash out a proper solution based on your prototype. Do ask if you need any kind of help.

Krizai:
My idea was to go simple way and add some additional "column" property for newline chunks and set it in the same way as they are set for normal chunks during indentation. In this way, we don't need to break current grouping of the newlines.

gmaurel:
FFixed

PR #1660, opened 2018-03-06
Add feature to indent lines which contain only single newline

Krizai:
Solution to https://github.com/uncrustify/uncrustify/issues/1641

Added config parameter indent_single_newlines
Enabling this parameter will allow indenting single newline lines in the same way as normal lines.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1084,6 +1084,12 @@ void indent_text(void)
       indent_column_set(frm.top().indent_tmp);
       log_indent_tmp();
 
+      if (  pc->type == CT_NEWLINE
+         && cpd.settings[UO_indent_single_newlines].b)
+      {
+         pc->nl_column = indent_column;
+      }
+
       if (  !chunk_is_newline(pc)
          && !chunk_is_comment(pc)
          && log_sev_on(LINDPC))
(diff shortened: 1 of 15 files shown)


gmaurel:
Thanks for the PR.
Please change it to use the `argtype_e::AT_TFI` for the new option.
Change **newline** to **empty** line.
The value `AT_TFI` should be supported to get: "not any change" for the empty lines.
Provide an example (or more) to test the value `AT_TFI` on empty lines, lines with not enough spaces, lines with too much spaces.

Krizai:
Didn't get an idea. What will T/F/I mean for this option?

guy-maurel:
Many users wish to have a _NULL_ configuration file, to test one option at the time.
We don't have it just now. But it would be fine to work in this direction.
All new options should support that: false / true / ignore instead of false / true
For the new option you propose:
`false`: remove the spaces on empty lines,
`true`: add spaces on empty lines (according your proposal)
`ignore`: don't modify empty lines (even there are some spaces)

Krizai:
Ignore will be tricky, as now we are collapsing all the newlines to one, so the only "easy" option I see is to save an array of the number of spaces for newline chunks.

gmaurel:
You are right, the number of spaces is not known. Let us forget it and
Stay with the `argtype_e::AT_BOOL`
Please search for  better name for **newline** or **empty line** to make clear that the option works on line without token.

Krizai:
What about "Indent empty lines - lines which contain only spaces before newline character"?

gmaurel:
It would be fine.

guy-maurel:
You need to use the test:` ./tests/cli/test_cli_options.py ` to check if all the tests are OK.

gmaurel:
Sorry, it is not quite exact.
You need to use the test:
```
cd ./tests/cli
./test_cli_options.py
```
to check if all the tests are OK.

PR #1662, opened 2018-03-09
Show actual difference of files rather than asking user to find differences

CDanU:
cleaned and fixed version of #1598


gmaurel:
Fine work. Thanks.

mihaipopescu:
while this was nice, it's now causing case-folding collisions on non-case-sensitive systems

Issue #1637, opened 2018-02-23
Extra space added when used option 'mod_paren_on_return=remove' 


KalyanaChakravarthi:
```diff --git a/UNI-24076.cpp b/test.out.cpp
index ac31eea8..368ae2b8 100644
--- a/UNI-24076.cpp
+++ b/test.out.cpp
@@ -1,2 +1,2 @@
 //Test Case-001
-inline double GetAudioBitrateForQuality(double f) { return (56000 + 200000 * (f)); }
+inline double GetAudioBitrateForQuality(double f) { return 56000 + 200000 * (f);  }
```
[UNI-24076.cpp](https://github.com/uncrustify/uncrustify/files/1750583/UNI-24076.cpp.txt)

KalyanaChakravarthi:
In unity fork this issue is addressed with below PR
Unity-Technologies#108

guy-maurel:
Is the test file CPP or CS?

gmaurel:
Fixed

PR #1678, opened 2018-03-17
Bug 1637

guy-maurel:
fixing columns when mod_paren_on_return set to remove
    
Ref. #1637 
Thanks to Kalyana Chakravarthi


guy-maurel:
Fixed

PR #1731, opened 2018-04-30
Prepare a test using gcov

guy-maurel:
**Please read the TAKE ATTENTION lines**!

_Running the test is long. I need about 20 minutes.
This is about 40 times so much as the ctest.
The disk space necessary is also very big, about 3 Gbytes
This is about 1500 times bigger as the sources._

```txt
The script prepare a new version of uncrustify with the compile options:
 CMAKE_CXX_FLAGS                  -fprofile-arcs -ftest-coverage
 CMAKE_C_FLAGS                    -fprofile-arcs -ftest-coverage
to use the facilities from gcov.
Running uncrustify with all the test data will mark all parts of the sources
which are used.
The "not marked" portions, if any, should give the opportunity to prepare new
test data to complete the whole tests.
The results are stored in the directory ${TOTALS_DIR}
The name of the file is ${source_file}.total
The line(s) of code which are still not used by any of the tests cases are
marked which the token "#####" at the beginning of the line.
As the testing part (unc_tools.cpp, backup.cpp) are only used by a developper,
all the lines are marked.
Also the detect.cpp part of uncrustify is completly marked.
```


mihaipopescu:
Nice! I think we can add this test to our CI test matrix (travis and appveyor) and report the coverage status with a badge on the front page.

Issue #672, opened 2016-09-09
long method and variable declarations not correctly indented


gmaurel:
hi Ben,

The fix for 536 solves the problem for the class declarations. 

Long method and variable declarations though are still not being indented correctly. They work normally, but in case of templatized parameters they break. Adding an additional check should handle this problem.
Around line 1430 in indent.cpp we have this condition check. We need to add this additional check. See comment alongside the new condition.

<code>
     else if ((vardefcol > 0) &&
              (pc->level == pc->brace_level) &&
              ((pc->type == CT_WORD) || (pc->type == CT_FUNC_CTOR_VAR)) &&
              (prev != NULL) &&
              ((prev->type == CT_COMMA) ||
               (prev->type == CT_TYPE) ||
(cpd.lang_flags & LANG_JAVA && prev->parent_type == CT_TEMPLATE) || // New condition added
(prev->type == CT_WORD)) &&
((pc->flags & PCF_VAR_DEF) != 0))
{
LOG_FMT(LINDENT, "%s: %d] Vardefcol => %d\n",
func, pc->orig_line, vardefcol);
reindent_line(pc, vardefcol);
}
</code>

For the long method problem though, I am afraid I don't have a solution.

Reported by: kaiser101

Original Ticket: [uncrustify/bugs/555](https://sourceforge.net/p/uncrustify/bugs/555)

kaiser101:
Can someone help me with this? Everything seems to be working except class definition statements. Long method declarations, statements all work fine with indent_continue.

Checking logs with an earlier version of uncrsutify suggests that there are 2 passes. First is in indent.cpp when the line is split and indent continue comes into play. In this case the next line is indented correctly i.e. using the value set in indent_continue. But then there is a second pass in one of the cleanup.cpp files, brace or tokenize I dont remember. Thats where the indent for the split line is set back to the original indent, i.e. the indent_continue value is disregarded.

In a nutshell the feature is expected to behave as follows for class declarations
indent_columns = 4
indent_continue = 8
code_width = 80

Input
public abstract class KeyValueItemWriter<K, V> implements ItemWriter<V>, InitializingBean {

Expected output
public abstract class KeyValueItemWriter<K, V> implements ItemWriter<V>,
        InitializingBean { // indented by 8 spaces, instead of 4

Actual output // incorrect
public abstract class KeyValueItemWriter<K, V> implements ItemWriter<V>,
    InitializingBean { // indented by 4 spaces

[...]

PR #1765, opened 2018-05-29
A bugfix proposal for #672

guy-maurel:
We introduce a new frame after **class**.
The indentation is set to the value of the option **indent_continue**, if not zero.
We close it before the opening **brace**, if opened.
The indentation is reset to 1.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -2964,7 +2964,7 @@ void indent_text(void)
       // Issue #672
       if (  chunk_is_token(pc, CT_CLASS)
          && language_is_set(LANG_CPP | LANG_JAVA)
-         && cpd.settings[UO_indent_continue].u != 0
+         && cpd.settings[UO_indent_continue_class_head].u != 0
          && !classFound)
       {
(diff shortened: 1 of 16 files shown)


gmaurel:
@mihaipopescu I don't anderstand what is happening (screenshot here
github-may-be-merge )
Why the text in red?

mihaipopescu:
It clearly says: reviewers with write access. Look in repo settings maybe
you can fix that to allow reviewers from maybe a list or something else but
not necessarily that have write access.

mihaipopescu:
Maybe we can try with a CODEOWNERS file https://help.github.com/articles/about-codeowners/ but that will require some more involvement and responsibility from the most active collaborators.

If that doesn't work, you can deactivate it.

gmaurel:
The CODEOWNERS file is installed. I try to review.

gmaurel:
It works with the new file.

mihaipopescu:
I think it worked because you approved it yourself.

PR #1767, opened 2018-05-29
Change

guy-maurel:
Change the code
```.cpp
   if (!end)
```
to
```.cpp
   if (end == nullptr)
```

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -5406,7 +5406,7 @@ static void handle_cpp_lambda(chunk_t *sq_o)
       // REVISIT: really should check the stuff we are skipping
       br_o = chunk_get_next_type(br_o, CT_BRACE_OPEN, br_o->level);
    }
-   if (br_o == nullptr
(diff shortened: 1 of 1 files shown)


mihaipopescu:
How does Uncrustify knows that `end` is a pointer type?

gmaurel:
end is a place holder for a pointer.

mihaipopescu:
Ok, I totally got the description of the PR wrong. I thought you want to add a new option to uncrustify to convert `if (!p)` => `if (p==null)`.

As expressed in our [guidelines](https://github.com/uncrustify/uncrustify/blob/master/CONTRIBUTING.md) an ideal PR description should tell you _why_ a change was made and _why_ it's better than before instead of saying _how_ or _what_ has changed which should be obvious by looking at the changes. That's why we do PRs in the first place.

And as expressed earlier, it would be amazing if we can enforce code reviews in GitHub. What do you say?

Issue #1758, opened 2018-05-25
Uncrustify failed to detect braced blocks


sgarnotel:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastbin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

When I use uncrustify in such a loop, containing a bradec block, it remove the braces of the for loop (it detects it as a one line for loop)

```
for(int f=0; f<(Element::nf)*2; f++)
{
	if (f%2==1) {p = p-1;}
	{
	this->pInterpolation[i]=p;
	this->cInterpolation[i]=0.;
	this->dofInterpolation[i]=e+f;
	this->coefInterpolation[i]=1.;
	i++;
	p++;
	}
}
```

EDIT: this bug does not appear if:
```
# Add or remove braces on single-line 'for' statement.
mod_full_brace_for              = force   # ignore/add/remove/force
```

guy-maurel:
Please push your config and the output you expect to get with.

sgarnotel:
The configuration file can be found [here](https://github.com/FreeFem/FreeFem-sources/blob/079f9e7dcff7c8f82c9d7cab6fbac0575e22367d/tools/.uncrusify).

I expect to get:
```
[...]

PR #1785, opened 2018-06-08
A block was not correctly recognize.

guy-maurel:
Add some more LOG_FMT for a better logging
A bugfix proposal for #1758


tests/config/bug_1758.cfg
```diff
@@ -0,0 +1,2043 @@
+#
```
CDanU: That needs a run through the reduction script.

I remember that you had a reason to postpone the usage of it but I don't quiet remember anymore why and if it is applicable for this case. I believe it was something related to the Unity fork.

tests/config/bug_1758.cfg
```diff
@@ -0,0 +1,11 @@
+output_tab_size                 = 4
+sp_arith                        = force
+sp_assign                       = force
+sp_compare                      = force
+sp_before_sparen                = force
+indent_columns                  = 4
+nl_end_of_file                  = force
+nl_end_of_file_min              = 2
```
CDanU: Still not to happy about the config, some of those options are definitely not needed. This needs a alteration of the test file. Id also like to see a test case where we have something like: `while(){if(){}}`

guy-maurel: Why not "happy"?
This is the output of option_reducer.py

CDanU: Yes the reduction script tries to reduce the option without changing the output formatting. The problem in this case is that the output file itself is to complex, it need to be simplified.

The initial problem is that the for braces are disappearing when they should not.

That means we definitely do not need any of:
```sql
sp_*
nl_end_of_file                  = force
nl_end_of_file_min              = 2
```

we most likely do not need:
```sql
output_tab_size                 = 4
```

and  _maybe_ we do need:
```sql
nl_for_brace                    = remove
nl_after_if                     = force
mod_full_brace_if               = remove
```

CDanU:
So now only the example with the on liner is missing:
> I'd also like to see a test case here where we have something like: for(){if(){}}

And it would be nice if you could reformat the test input with an empty config so that we can get a diff in the end that is only depending on the test options (excluding changes that are made by the default options)

guy-maurel:
I prefer a new test case for your proposal.

src/braces.cpp
```diff
@@ -528,14 +558,18 @@ static void examine_brace(chunk_t *bopen)
                || chunk_is_token(pc, CT_WHILE)
                || chunk_is_token(pc, CT_SWITCH)
                || chunk_is_token(pc, CT_USING_STMT)
-               || (  chunk_is_token(pc, CT_BRACE_OPEN)
-                  && chunk_is_token(prev, CT_FPAREN_CLOSE)))
+               || chunk_is_token(pc, CT_BRACE_OPEN)) // Issue #1758
```
mihaipopescu: Really hard to tell apart changes. Is this the only functional change?

guy-maurel: It is the consequence of the bug #1758 
The condition  `&& chunk_is_token(prev, CT_FPAREN_CLOSE)` was not apropriated to solve the case.

mihaipopescu: The question is "Is this the **only** change that fixes the bug" ?

mihaipopescu: because they are mixed with other changes, that's why I ask.

guy-maurel: Yes. It is the only line.
I made lot of changes by the use of the debug function `LOG_FMT(...)`
to find out what the matter is.
I might be better to make to PR for that.

mihaipopescu: no, it's fine like this but yeah, maybe next time different commits with different things changed would help in review.

guy-maurel: Anybody to approve?

CDanU: Hmm that whole function is a bit strange ... here is one failure:

```cpp
int main()
{

for(int f=0; f < 1; f++)
{
	auto a = int{1};
}

return 0;
}
```

mihaipopescu:
@CDanU what do you think about the recent changes?

PR #1790, opened 2018-06-10
fix make_version.py

CDanU:
removes the upper char limit for git hashes
git describe started to display longer hashes for some reason


mihaipopescu:
because, very rarely they collide.

PR #1800, opened 2018-06-13
Fix potential nullptr dereference issue

testkit:
Check if tmp is nullptr before be dereferenced.


testkit:
I'd like to, but didn't find a proper case yet. It's ok to close this PR currently if it blocks somethings. I'll provide PR later on with proper cases.

PR #1814, opened 2018-06-17
Integrate coveralls

mihaipopescu:
While working on adding OSX on the travis test matrix, I noticed that @jibsen 2 years ago in https://github.com/uncrustify/uncrustify/commit/215605ad60e64dfd1b39486f045798bb4522e71f he might have wanted to add testing coverage reporting to uncrustify but never finished.

After struggling with lack of complete documentation from both cmake and coveralls, I implemented it for our fork and it works, yielding an impressive 85%.

The only thing I need on this PR is a token from @gmaurel which I requested.


cmake/CodeCoverage.cmake
```diff
@@ -15,7 +15,6 @@ if ( NOT DEFINED CODECOV_HTMLOUTPUTDIR )
 endif ( NOT DEFINED CODECOV_HTMLOUTPUTDIR )
 
 if ( CMAKE_COMPILER_IS_GNUCXX )
-    find_program( CODECOV_GCOV gcov )
     find_program( CODECOV_LCOV lcov )
```
CDanU: Neat! 
I think we can also drop lcov as it is used with gcov together.

CDanU: Ohh I haven't looked too closely at the changes that you have made,  I thought that coveralls is doing all of the gcov+lcov runs.

mihaipopescu:
now travis coverage build fails with
```
{'error': True, 'message': "Couldn't find a repository matching this job."}
```

because I need the right app key from @gmaurel.

cmake/CodeCoverage.cmake
mihaipopescu: `coveralls` can do `gcov` runs but it'll be a hassle to configure that path now. I found the `lcov` approach more flexible and `coveralls` can use its output. If at any point we want to have html output we can use the other target added `coverage_html`.

mihaipopescu:
interesting, I'm getting coverage reports from coveralls in my PRs like this:
https://github.com/Unity-Technologies/uncrustify/pull/148#issuecomment-398326480

gmaurel:
Sorry for delay. I don't know how to to it: `right app key`

mihaipopescu:
This is how it looks on our fork when coverage drops from a PR https://coveralls.io/builds/17619279.
@guy-maurel You still didn't received anything in your email? Try to look in spam of the email associated with your github account. Or try to add the project yourself to coveralls. I don't know understand why you don't get the email.

guy-maurel:
No mail.
guy-maurel is my account as a developer.
gmaurel is my account as admin.
I try myself to do the work.

mihaipopescu:
Indeed it's strange that they need that level of access. Apparently other people complained here https://github.com/lemurheavy/coveralls-public/issues/521 and looks like the workaround is to revoke the permission later.

mihaipopescu:
And this is how you revoke it from your Github Profile > Settings > Applications > Authorized OAuth Apps:
screen shot 2018-06-22 at 23 54 05

I think there is nothing to worry about. We are using coveralls at Unity for many other projects on github (public and private) and you can revoke the access later if you don't want the Pro version.

Other alternatives are codecov and lcov itself which can generate html output. But that has to be uploaded somewhere and maybe to appear as a comment here or something. But it means more work and it's not going to be the same as other integrated solutions.

gmaurel:
@mihaipopescu Done.
Please test and report

.travis.yml
```diff
@@ -62,11 +62,23 @@ matrix:
     - os: osx
       env: BADGE=osx BUILD_TYPE=release
 
+    - os: linux
+      dist: trusty
+      env: BUILD_TYPE=coverage COVERALLS_TOKEN="iJXKQxvt6RJMcxkn99GIhakcoCx7XvzbU"
```
mihaipopescu: @gmaurel Can you comment here the token you got from Coveralls? Thanks

PR #1816, opened 2018-06-18
Split travis build status badge for linux and osx

mihaipopescu:
We can actually split the badge when our travis build matrix is targeting multiple operating systems using an environment variable as mentioned here https://github.com/travis-ci/travis-ci/issues/9579 and in the docs here https://github.com/exogen/badge-matrix\#travisuserrepo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,18 +5,18 @@ matrix:
   include:
     - os: linux
       dist: trusty
-      env: BUILD_TYPE=debug
+      env: BADGE=linux BUILD_TYPE=debug
     - os: linux
       dist: trusty
-      env: BUILD_TYPE=release
+      env: BADGE=linux BUILD_TYPE=release
     - os: linux
       dist: trusty
-      env: BUILD_TYPE=release NO_MACRO_VARARG=1
+      env: BADGE=linux BUILD_TYPE=release NO_MACRO_VARARG=1
 
     # GCC 6
     - os: linux
       dist: trusty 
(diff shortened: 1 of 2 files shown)


CDanU:
It looks broken on your branch: https://github.com/Unity-Technologies/uncrustify/tree/ci-cd/fix-os-badges.
I see two build - unknown and one windows - passing.

mihaipopescu:
Good point. I only tested with

http://badges.herokuapp.com/travis/uncrustify/uncrustify?env=BUILD_TYPE=debug&label=build&branch=master

http://badges.herokuapp.com/travis/uncrustify/uncrustify?env=BUILD_TYPE=foo&label=build&branch=master

which work as expected.

But I'll prove that it works on our fork with https://github.com/Unity-Technologies/uncrustify/pull/149.

mihaipopescu:
seems to work https://github.com/Unity-Technologies/uncrustify/blob/master/README.md

just need a bit of faith 😉

CDanU:
Would it be possible to get a either only _passed_ or _passing_ instead of mixing those terms?

mihaipopescu:
hmm... yeah. You have a point but the status is managed by the badge service... I'm unsure if there is something I can do about that.

mihaipopescu:
In any case, then we need to reflect that we've added OSX as well somehow into the badges... So either add have `Linux&OSX | passing` and use shields.io badges... or use this one instead. As you want.

ianfixes:
> _passed_ or _passing_

https://www.appveyor.com/blog/2014/11/26/customizing-text-of-build-status-badges/

PR #1830, opened 2018-06-22
Sync configs with unity [skip ci]

mihaipopescu:


mihaipopescu:
why does every time github decides to make a default pull request for upstream?

PR #1850, opened 2018-07-03
Improve parallelism of CI builds

mwoehlke:
This changes some (but not all) of the CI builds to use [ninja](https://ninja-build.org/) instead of make to build. Besides improving build system coverage, ninja is usually faster, particularly since it defaults to parallel builds. This also introduces a small(ish) script to wrap invoking CTest that will automatically make a guess at an appropriate level of parallelism, and changes *all* the CI builds (appveyor, particularly) to use it.

While I'm certainly not accusing @mihaipopescu (#1814) of picking `-j4` out of a hat, this approach will always come up with a "good" answer even if the CI hardware changes or is different for different builds. (The main reason for the script, however, is to DTRT on Windows, where — unlike with a proper shell — it's much more awkward to just grab the number of CPU's and use that to form a command line... and since we need it there, we might as well use it everywhere for consistency's sake.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,28 @@ matrix:
     - os: linux
       dist: trusty
       env: BADGE=linux BUILD_TYPE=debug
+      addons:
+        apt:
+          update: true
+          packages:
+            - ninja-build
+
     - os: linux
       dist: trusty
       env: BADGE=linux BUILD_TYPE=release
+      addons:
+        apt:
+          update: true
+          packages:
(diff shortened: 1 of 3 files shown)


mwoehlke:
Alas, the actual build speed-up is less than amazing, but, eh... :slightly_smiling_face:

OTOH, the test time on Windows (appveyor) is *significantly* reduced. #1821 would reduce the difference a fair bit, but this should still be an improvement over #1821 by itself.

Issue #1854, opened 2018-07-06
space before semicolon is removed


guy-maurel:
config:
```.cfg
sp_before_semi = ignore
```
input:
```.cpp
while (*p++ = ' ') ;
```
expected:
the same

get:
```.cpp
while (*p++ = ' '); // the space before the semicolon is removed
```

guy-maurel:
#1855 is a bugfix proposal.

gmaurel:
Fixed

PR #1855, opened 2018-07-06
Preserve the space before semicolon, even if after a virtual brace.

guy-maurel:
Ref. #1854


src/space.cpp
```diff
@@ -2433,8 +2433,17 @@ void space_text(void)
             {
                column += next->orig_col - pc->orig_col_end;
             }
+            else
+            {
+               // preserve the position if virtual brace
+               // Issue #1854
+               if (chunk_is_token(pc, CT_VBRACE_OPEN))
```
CDanU: Ok, but why don't you use else if here ?

mihaipopescu: Looks like @gmaurel forgot to reply to your question

PR #1858, opened 2018-07-07
this block will never been reached

guy-maurel:
Ref. https://coveralls.io/builds/17772219/source?filename=src/space.cpp#L193
The next comments from coveralls will be token as soon as possible.


guy-maurel:
Correct. 
@mihaipopescu 
The changes where introduce at 5453802646ffc96a0c188c16bf90892971df5ea7 on Mar 26 11:03:54 2018.
I make a cherry-pick from the https://github.com/Unity-Technologies/uncrustify/commit/2a204bc7891a9af451467438e46799d0cb9e1e96 
and used the test 60032 `staging/Uncrustify.Cpp.cfg staging/UNI-21729.cpp`
It might be not enough.
I tried to anderstand the 
```.cpp
if (chunk_is_token(first, CT_PAREN_CLOSE) && first->parent_type == CT_DECLSPEC)
```
and think it could never be true

mihaipopescu:
nice work

Issue #1838, opened 2018-06-26
Uncrustify-0.67_f goes does not finish, infinite loop suspected


falko-strenzke:
Uncrustify apparently goes into an endless loop:

Version: Uncrustify-0.67_f

the parsed c-file:
[tls_rec_prot.txt](https://github.com/uncrustify/uncrustify/files/2137082/tls_rec_prot.txt)

the uncrustify config:
[uncrustify.txt](https://github.com/uncrustify/uncrustify/files/2137087/uncrustify.txt)

CDanU:
I broke it down a bit, it certainly is an interesting one. 

C:
```slq
# The number of columns to indent per level.
# Usually 2, 3, 4, or 8. Default=8.
indent_columns                  = 2        # unsigned number

# Whether to add newline after '(' in a function call if '(' and ')' are in different lines.
nl_func_call_start_multi_line   = true     # false/true

# Whether to force a newline after a multi-line comment.
nl_after_multiline_comment      = true     # false/true

# If False, disable all multi-line comment changes, including cmt_width. keyword substitution and leading chars.
# Default=True.
cmt_indent_multi                = false    # false/true
```

I (spaces/newlines need to be exactly as stated):
```
  b(/*
*/)
```

PR #1859, opened 2018-07-08
prevent underflow in output_comment_multi_simple 

CDanU:
ref: #1838


src/output.cpp
```diff
@@ -2011,99 +2012,106 @@ static void do_kw_subst(chunk_t *pc)
...
-   else
-   {
-      // The comment starts after something else
-      col_diff = 0;
-   }
+   // The multiline comment is saved inside one chunk. If the comment is
+   // shifted all lines of the comment need to be shifter by the same amount.
+   // Save the difference of initial and current position to apply it on every
+   // line_column
+   const int col_diff = [pc](){
```
mihaipopescu: Is the lambda really required here?

src/output.cpp
```diff
@@ -2011,99 +2012,106 @@ static void do_kw_subst(chunk_t *pc)
...
          else if (ch == '\t')
          {
-            ccol = calc_next_tab_column(ccol, cpd.settings[UO_input_tab_size].u);
+            line_column = calc_next_tab_column(line_column, cpd.settings[UO_input_tab_size].u);
             continue;
          }
-         else
+      }
+
+      // 2: add chars to line, handle the CRLF and CR endings (convert both to LF)
```
mihaipopescu: > convert both to LF

why? isn't that controlled by an option?

CDanU: I just refactored the function to make it more understandable (mostly for my own sake).  While doing that I didn't gave too much thought into actually changing its behaviour outside of what I needed to change to fix the reported issue. 

Here the `\r\n` are converted into `\n`, this most certainly isn't needed here. 
The actual determination which line ends are going to be used in the end are happening inside the end of the tokenization function (using the `UO_newlines` option) and the write char function which actually is writing the correct line ends.

src/output.cpp
CDanU: Well not really but I wanted the `col_diff` to be const.

src/output.cpp
CDanU: Dropping the check will force me to make nasty checks in other places. 
Why do we keep the \r around anyways? I will try to remove them in the tokenization and see if that causes a total mess or simplifies things even more.

CDanU:
I am not entirely sure why coveralls doesn't show that all new lines are covered ... I stepped locally through the code, I do hit this block and it works as it should. The only reason that I can think of is that git somehow eats the '\r' in the push process (or fetch for the travis run) but that shouldn't happen.

PR #1879, opened 2018-07-11
Change the value of option sp_oc_catch_paren

guy-maurel:
This test was missed bei https://coveralls.io/builds/17931530/source?filename=src%2Fspace.cpp#L683


guy-maurel:
Some more explanations about the work on the space.cpp file and the informations we are getting from coveralls.
space680
The picture shows what corveralls detected: The sign ! at the right of the picture: Not any test produce
a trace at those lines. We get many of those warnings.

I now take each one, look for reason, and make some changes to get a trace at those lines.
It is sure we don't have any other errors: travis would have show them.

gmaurel:
The file .travis.yml must have:
```.cfg
sudo: required
```

nivekkagicom:
The implementation for the sp_oc_catch_* options uses ignore to fall through to the sp_catch_* options if the value for the first setting is ignore. The Objective-C test case 50512 tested that 'ignore" was properly handled and no longer does so. Is there a preferred way to provide coverage tests that also test options that fall through to other settings if the value is ignore?

PR #1883, opened 2018-07-12
Change the value of option sp_enum_after_assign

guy-maurel:
To get a better trace of coveralls.

This trace was missed at https://coveralls.io/builds/17772219/source?filename=src/space.cpp#L795


guy-maurel:
I don't anderstand! travis produces yesterday an error saying sudo where needed.

mihaipopescu:
What was the error? Usually sudo should not be required.

PR #1884, opened 2018-07-12
Change the value of option sp_angle_colon

guy-maurel:
To get a better trace of coveralls

This trace was missed at https://coveralls.io/builds/17772219/source?filename=src/space.cpp#L971


tests/config/indent_with_tabs-0.cfg
```diff
@@ -1 +1,2 @@
-indent_with_tabs                = 0
\ No newline at end of file
+indent_with_tabs      = 0
+sp_enum_before_assign = remove
```
CDanU: Could you either make a new config for that or rename this config file? 

Way back when I did the clean up of the test configs I spend some effort to name single option configs according to the pattern that you see in the config name (`<optionName>-<1stChar:optionValue>`).
I wanted to have that pattern in order to clearly see which option/value pairs are already having tests and which ones don't.

gmaurel: Thanks.

PR #1889, opened 2018-07-13
false test of type for the option sp_before_byref_func

guy-maurel:
5453802646ffc96a0c188c16bf90892971df5ea7 introduced an error.
We must test the parent_type


guy-maurel:
As we don't have enough tests, the error wasn't detected.
Thanks to @mihaipopescu to have introduced coveralls.

PR #1890, opened 2018-07-13
Change the test source to get a trace for the option sp_before_semi

guy-maurel:
Change the value of option sp_before_semi

To get a better trace of coveralls

This trace was missed at https://coveralls.io/builds/17772219/source?filename=src/space.cpp#L1038

### DIFF
--- a/tests/input/cpp/bug_1158.cpp
+++ b/tests/input/cpp/bug_1158.cpp
@@ -1,4 +1,4 @@
 void Class1::Func(void)
 {
-        while (Next());
+        while (Next()) ;
 }


CDanU:
Did you forget to push a new config for this test?  The input file `cpp/bug_1158.cpp` is called with an `empty.cfg`.

gmaurel:
This is correct. It uses the default value REMOVE.

CDanU:
I am a bit confused, wasn't the goal of this to change the tests so that the lines in your first linked coveralls run are tested? That is currently still not happening.

guy-maurel:
The goal is to get a real change of the test data. This will be shown by coveralls.
This is the most important part of this work (to day on space.cpp) with coveralls.
This way should we be (after) sure that all the part of uncrustify are used.
coveralls shows us the duplicate code #1898

Issue #1894, opened 2018-07-14
A question about objective C with the option sp_oc_catch_brace


guy-maurel:
The description is:
_Add or remove space between `@catch` and '{' if on the same line._

I never found any example for that. But something such as:
https://www.tutorialspoint.com/objective_c/objective_c_exception_handling.htm

```.oc
   @try  {
      NSString *string = [array objectAtIndex:10];
   } @catch (NSException *exception) {
      NSLog(@"%@ ",exception.name);
      NSLog(@"Reason: %@ ",exception.reason);
   }
```

I see always a _(NSException *exception)_ after `catch` and before the brace.

If this is correct, we could remove the case.
Any comment from "OC" people?

mihaipopescu:
cc @aet

nivekkagicom:
An example for sp_oc_catch_brace can be found in Objective-C test 50512. In the bar message implementation is `@catch (NSException* const inException){` and the configuration specifies that uncrustify should add a space before the '{' to produce `@catch (NSException* const inException) {`. Functionally it is identical to the C++ options for catch statements (e.g., `catch(std::exception const& inException){`). Because some people like different formatting for the two exception statements uncrustify has sets of options for C++ and Objective-C to allow this customization. Because other people like the same formatting for either exception if any of the sp_oc_catch_* settings is ignore then the uncrustify behavior falls through to the more generic sp_catch_* setting. For example, sp_oc_catch_brace falls through to sp_catch_brace if the first is set to ignore.

Is this the information you are looking for?

guy-maurel:
Yes and no.
The description is:
_Add or remove space between '@catch' and '{' if on the same line._
The code is:
```.cpp
   if (  language_is_set(LANG_OC)
      && chunk_is_token(first, CT_CATCH)
      && chunk_is_token(second, CT_BRACE_OPEN)
      && (cpd.settings[UO_sp_oc_catch_brace].a != AV_IGNORE))
   {
      log_rule("sp_oc_catch_brace");
      return(cpd.settings[UO_sp_oc_catch_brace].a);
   }
```
You speak about:
_@catch (NSException* const inException){_

The description, the space.cpp code work on:
_@catch {_

This is **NOT** the same.
Following (for C++) https://en.cppreference.com/w/cpp/language/try_catch

We could introduce **two new options**, such as:
1 _Add or remove space between '@catch' and the '(' of the declarator if on the same line._
[...]

PR #1900, opened 2018-07-16
correct handling of sp_catch_brace and sp_oc_catch_brace

nivekkagicom:
This change corrects the handling of `sp_catch_brace` and `sp_oc_catch_brace` to control the space in a catch statement between the expression and the brace if on the same line (i.e., not placed on separate lines as `nl_catch_brace` or `nl_oc_catch_brace` would dictate). Specifically:
```
try {
catch (std::exception&)<controls this spacing>{
}
```
and
```
@try
@catch (NSException*)<controls this spacing>{
}
```


nivekkagicom:
The Objective-C case is already covered by test 50512 as mention in the original issue #1894. I believe the C++ case is already covered in test 30015 though I did not write that particular test so I cannot speak authoritatively for it.

mihaipopescu:
If this PR is fixing a broken behaviour, I'd like to be able to see that also captured in a test. I don't understand how an old passing test still covers this new fix as well. Maybe I'm missing out something.

gmaurel:
You need to make a little change to config/sp_oc_catch.cfg
```.cfg
sp_oc_catch_brace = remove
```
to force a change (remove the space before the brace).
Please update the expected file(s)

nivekkagicom:
Using `sp_oc_catch_brace - remove` would accomplish nothing different during the test. The input line in question is:
`@catch (NSException* const inException){
`
Or in other words it is already removed. The setting in the test config is:
` sp_oc_catch_brace               = add      # ignore/add/remove/force
`
Which adds the space per the expected output line:
`@catch (NSException* const inException) {`

src/options.cpp
```diff
@@ -611,9 +611,9 @@ void register_options(void)
    unc_add_option("sp_brace_typedef", UO_sp_brace_typedef, AT_IARF,
                   "Add or remove space between '}' and the name of a typedef on the same line.");
    unc_add_option("sp_catch_brace", UO_sp_catch_brace, AT_IARF,
-                  "Add or remove space between 'catch' and '{' if on the same line.");
+                  "Add or remove space between 'catch' and '{' in 'catch (something) { }' if on the same line.");
```
mwoehlke: I'm confused, how is a space added between `catch` and `{` if there is *already* non-space text between those? I guess the description should be revised to clarify?

If it's before the brace, I recommend "Add or remove space before '{' in 'catch (...) { }' if on the same line.".

src/options.cpp
```diff
@@ -1110,9 +1110,9 @@ void register_options(void)
    unc_add_option("nl_for_brace", UO_nl_for_brace, AT_IARF,
                   "Add or remove newline between 'for' and '{'.");
    unc_add_option("nl_catch_brace", UO_nl_catch_brace, AT_IARF,
-                  "Add or remove newline between 'catch' and '{'.");
```
mwoehlke: Similar question here; where is the newline added?

tests/config/sp_catch_brace.cfg
```diff
@@ -0,0 +1 @@
+sp_catch_brace = force
```
mwoehlke: BTW, I know we have a *lot* of files that are missing newlines, but can we please try to stop adding new ones?

tests/cpp.test
```diff
@@ -60,6 +60,7 @@
 30051  nl_namespace-a.cfg                   cpp/nl-namespace.h
 
 30052  nl_try-a.cfg                         cpp/try-catch-nl.cpp
+30053  sp_catch_brace.cfg                   cpp/exception.cpp
```
mwoehlke: Reusing an existing test input is good, but it would be better if you also changed the existing input to have two spaces where `sp_catch_brace` is meant to trigger, and add two configs; one `force` and one `remove`. This way you ensure the rule is actually triggering where expected. Right now, because the input and output are the same, the test could be passing by accident.

tests/config/sp_catch_brace.cfg
mwoehlke: Thank you!

src/space.cpp
```diff
@@ -1027,11 +1027,27 @@ static argval_t do_space(chunk_t *first, chunk_t *second, int &min_sp)
 
    if (chunk_is_token(first, CT_SPAREN_CLOSE))
    {
-      if (  chunk_is_token(second, CT_BRACE_OPEN)
-         && (cpd.settings[UO_sp_sparen_brace].a != AV_IGNORE))
+      if (chunk_is_token(second, CT_BRACE_OPEN))
       {
-         log_rule("sp_sparen_brace");
-         return(cpd.settings[UO_sp_sparen_brace].a);
+         chunk_t *prev = chunk_get_prev_type(first, CT_CATCH, first->level);
```
mwoehlke: Ick... not entirely your fault, but this is inadequately sophisticated. Consider:
```c++
try { ... }
catch (int) foo();
while (true) { ... }
```

This is affecting the space before the `{` of the `while`. (I suspect it may do that with or without braces around the `catch` block.) You need to instead find the `SPAREN_OPEN` that matches the `SPAREN_CLOSE` (I'd hope we have a sane way of doing that) and check if the preceding token is `CATCH`.

nivekkagicom: What language are you using? This seems like a contrived example since it is not valid C++, Objective-C++, or Java syntax. For these languages the syntax is `catch ( ... ) compound-statement` where `compound-statement` is defined to be `brace-enclosed sequence of statements`.

mwoehlke: C++. Okay, so I guess for some inexplicable reason, C++ requires braces for a `catch`, even though it almost never requires them otherwise. However, as noted, the issue exists whether or not braces are present. The problem is that `chunk_get_prev_type` tries *really hard* to find the token you requested. AFAICT, the only thing that will stop it is running out of tokens, which means it could match a `catch` in an entirely different *function*... as long as the brace level is the same.

(Maybe it won't span lines, but even so, I could still create a pathological case with two functions on the same line. Granted, I'd probably keel-haul any coder that asked to integrate such code, but... [BLIWYA](https://en.wikipedia.org/wiki/Robustness_principle) :slightly_smiling_face:.)

mwoehlke: > [`chunk_get_prev_type`] could match a catch in an entirely different *function*...

So... yeah. I tried that, just for fun. And, it does :scream:. Even across multiple lines.

nivekkagicom:
FWIW I will not be making further changes to this PR as I have already spent more time on it than I can afford and certainly more than it warrants. Accept it "as is" or reject it and have someone else fix the issue(s).

WRT `nl_catch_*` setting descriptions (and the `sp_catch_*` for that matter) see #1894. In that issue the question was raised about the meaning of "Add or remove space between @catch and '{' if on the same line." The wording and code were simply copies of the standard catch handler to allow separate formatting of `catch` statements and `@catch` statements in Objective-C++ code. I did not implement the original `catch` handling. I believe that was Ben. However, I did add the Objective-C handling and, therefore, attempted to explain its usage. In the process of answering the issue I realized that Ben's original code and thus my implementation for `sp_catch_brace` were not working as expected. Both `sp_catch_brace` and `sp_oc_catch_brace` were essentially ignored. I reworked the code for the `sp_catch_brace` and `sp_oc_catch_brace` settings to correct the behavior. The `sp_catch_brace ` and `nl_oc_catch_brace` behaviors were already correct.

As the original issue questioned the wording of the `sp_catch_brace` setting, as the setting is directly related to the `nl_catch_brace` setting and for consistency the wording for `nl_catch_brace` settings were changed to that used for the `sp_catch_brace` settings. As stated previously the behavior of `nl_catch_brace` was already correct so there are no need for code changes.

mwoehlke:
Makes sense; thanks for clarifying!

@mihaipopescu, are your concerns addressed? (I think the addition of a unit test does this?)

src/brace_cleanup.cpp
```diff
@@ -760,9 +760,16 @@ static bool check_complex_statements(ParseFrame &frm, chunk_t *pc)
    {
       if (chunk_is_token(pc, CT_CATCH) || chunk_is_token(pc, CT_FINALLY))
       {
-         // Replace CT_TRY with CT_CATCH on the stack & we are done
+         // Replace CT_TRY with CT_CATCH or CT_FINALLY on the stack & we are done
          frm.top().type  = pc->type;
-         frm.top().stage = (chunk_is_token(pc, CT_CATCH)) ? brace_stage_e::CATCH_WHEN : brace_stage_e::BRACE2;
+         if (language_is_set(LANG_CS))
+		 {
```
mwoehlke: CI does not like your indent on this brace... same a few lines lower...

mwoehlke: Looks like you have a stray extra space, also:

```
Problem with brace_cleanup.cpp
use: diff ./src/brace_cleanup.cpp ./results/brace_cleanup.cpp to find why
764c764
<          frm.top().type  = pc->type;
---
>          frm.top().type = pc->type;
766c766
< 		 {
---
>          {
770c770
< 		 {
---
>          {
```

nivekkagicom:
As it turns out the whole catch statement processing was broken 2 years ago (2194f94) when code was added to handle C# `when` clauses. Restoring the original behavior for sources other than C# allows some of the logic for this PR to be greatly simplified. Ideally a unit test would have caught the problem when the C# changes were introduced; however, as uncrustify only has output checking there is no way to validate the internal parsing state when changes are made.

mwoehlke:
> Ideally a unit test would have caught the problem when the C# changes were introduced; however, as uncrustify only has output checking there is no way to validate the internal parsing state when changes are made.

That's... mostly true. In fact, there is *some* validation of at least the parse tree dump that `-p` produces done by `test_cli_options.py`, but yeah, we could be doing much better. Eventually I want to split those tests out into a framework more like the formatting tests, and especially to get as many of them as possible working for both debug and release. (The reason they are currently release-only is because the debug output has "stack" traces, so the output is different between debug and release for some tests. Some of the tests, and in particular, the more critical ones that verify the config dumps, don't *actually* care about which configuration you built. The others are open for discussion if we want to keep them release-only, or have separate expected output for debug and release. The latter is more maintenance effort, but OTOH those output files change much more rarely.)

All of the above is more or less why #1910 exists. If/when we can add non-format tests more easily, one of the things I want to do is start adding *unit* tests of `-p`, which, per your comment, will help us monitor when parsing changes and give us more ability to verify that the *parsing* is working correctly, not just that the formatting (to the extent it's covered) "seems" to be producing the expected results.

Don't get me wrong; we need the formatting tests also, but they are of widely varying quality. There are some that do a good job of targeting specific options and inputs for testing, but there are others that probably would still pass even if something breaks under the hood. (This, FYI, is why I favor paired tests that set a single option to `force` and `remove` and operate on files where both values will produce a change between input and output. Those IMO tend to be the best tests.) *Even so*, negative tests have their own value... Ultimately, no single test type is a panacea.

mihaipopescu:
@nivekkagicom yes. There are still some opened questions from @mwoehlke you may want to address.

mwoehlke:
> There are still some opened questions from @mwoehlke you may want to address.

There are/were? I think my comments were addressed. (Unless you mean the last one, but that doesn't really apply to this PR.) I might fiddle with the descriptions later, but #1852 is doing that sort of thing anyway...

PR #1915, opened 2018-07-19
Missing test for sp_before_sparen

guy-maurel:
For the option sp_before_sparen

This trace was missed at https://coveralls.io/builds/17772219/source?filename=src/space.cpp#L1974
Add a new test for it.


guy-maurel:
Yes. I have added a test. Later coveralls shows me that my test never use the lines I was working on.
This shows me that the code was duplicate. I have to remove the test. Is now done.

PR #1944, opened 2018-08-06
Create accessors for options

mwoehlke:
Create namespaced accessor functions for obtaining the value of an option. This will simplify the code slightly, and is also notionally type safe.

On its own, this is a really stupid change (see also commit message for more). The real objective here is to create something that is API-compatible with #1852, so that the nosiest parts of that can be done in a separate PR before the final form of #1852 goes up for review.

I *could* go ahead and start replacing uses in this PR, but I'm inclined to not do so, and instead have the next PR be an almost¹-pristine mechanical transformation, since the amount of churn is going to be impractical to review. This would then be followed by a third PR that cleans up any spots the mechanical transform misses. However, if y'all would prefer one great big PR, let me know.

(¹ There are some instances of code like `cpd.settings[expr ? UO_A : UO_B]` that won't be in such PR. I'm not 100% certain if the transform will butcher these, or if there are other cases that will get butchered. In any case, "almost" means I may back out some of the mechanical changes to avoid broken code, but the intent is to not make any changes *in addition* to the mechanical transform.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,6 +204,7 @@ add_executable(uncrustify
   src/logmask.cpp
   src/md5.cpp
   src/newlines.cpp
+  src/option.cpp
   src/options.cpp
   src/options_for_QT.cpp
   src/output.cpp
@@ -236,6 +237,7 @@ add_executable(uncrustify
   src/logmask.h
   src/log_levels.h
   src/md5.h
+  src/option.h
   src/options.h
   src/options_for_QT.h
   src/ParseFrame.h
--- a/src/option.cpp
+++ b/src/option.cpp
@@ -1,0 +1,2889 @@
+/**
+ * @file option.cpp
+ * Parses the options from the config file.
+ *
+ * @author  Ben Gardner
+ * @author  Guy Maurel since version 0.62 for uncrustify4Qt
+ *          October 2015, 2016
+ * @license GPL v2+
+ */
+#include "uncrustify_types.h"
+#include "args.h"
+#include "prototypes.h"
+#include "uncrustify_version.h"
+#include "uncrustify.h"
+#include "error_types.h"
+#include "keywords.h"
+#include "defines.h"
+#include <cstring>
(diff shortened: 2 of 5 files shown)


mwoehlke:
Unfortunately, github is rather stupid about copied and rewritten files. Here's the *real* diff of `options.{cpp,h}`:
```diff
diff --git a/src/options.cpp b/src/option.cpp
similarity index 99%
copy from src/options.cpp
copy to src/option.cpp
index 4a3b3ea2..ba01e418 100644
--- a/src/options.cpp
+++ b/src/option.cpp
@@ -1,5 +1,5 @@
 /**
- * @file options.cpp
+ * @file option.cpp
  * Parses the options from the config file.
  *
  * @author  Ben Gardner
diff --git a/src/options.h b/src/option.h
similarity index 99%
copy from src/options.h
copy to src/option.h
index 275a8427..28245ed5 100644
--- a/src/options.h
+++ b/src/option.h
@@ -1,14 +1,14 @@
 /**
- * @file options.h
- * Enum and settings for all the options.
+ * @file option.h
+ * Enumerations and data types for options.
  *
  * @author  Ben Gardner
  * @author  Guy Maurel since version 0.62 for uncrustify4Qt
  *          October 2015, 2016
  * @license GPL v2+
  */
-#ifndef OPTIONS_H_INCLUDED
-#define OPTIONS_H_INCLUDED
+#ifndef OPTION_H_INCLUDED
+#define OPTION_H_INCLUDED
 
 #ifdef EMSCRIPTEN
 #include <vector>
@@ -1108,4 +1108,4 @@ typedef group_map_value_options_t::iterator                        option_list_i
 typedef group_map_value_options_t::const_iterator                  option_list_cit;
 
 
-#endif /* OPTIONS_H_INCLUDED */
+#endif /* OPTION_H_INCLUDED */
```
...which is not scary at all :wink:. I also recommend just looking at the new `options.{cpp,h}` and not the diffs. (Note that most of the content of these files is generated.)

PR #1955, opened 2018-08-09
Test the change of newlines ending

guy-maurel:
The type of line endings may be change to the values: _Newlines Unix|Newlines Win|Newlines Mac_
This tests the changes to the possible values.

### DIFF
--- a/tests/config/lineEndings-to-Mac.cfg
+++ b/tests/config/lineEndings-to-Mac.cfg
@@ -1,0 +1,19 @@
+indent_columns   = 3
+indent_with_tabs = 0
+# (284)Align continued statements at the '='. Default=True
+# If False or the '=' is followed by a newline, the next line is indent one tab.
+#indent_align_assign = true
+
+# (527)The span for aligning on '=' in assignments (0=don't align)
+align_assign_span = 2
+# (528)The threshold for aligning on '=' in assignments (0=no limit)
+#align_assign_thresh = 2
+
+# (521)The threshold for aligning variable definitions (0=no limit)
+#align_var_def_thresh=1
+# (522)The gap for aligning variable definitions.
+#align_var_def_gap=2
+
+# (0)The type of line endings. Default=Auto.
+# Newlines Unix|Newlines Win|Newlines Mac|Newlines Auto
+newlines = cr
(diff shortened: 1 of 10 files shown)


tests/config/lineEndings-to-Mac.cfg
```diff
@@ -0,0 +1,19 @@
+indent_columns   = 3
+indent_with_tabs = 0
+# (284)Align continued statements at the '='. Default=True
```
CDanU: Do those comment have a meaning? I know that the text is the option description but what about the numbers?

tests/expected/cpp/30073-lineEndings-Unix.cpp
```diff
@@ -0,0 +1 @@
+int main (){   a    = 5;   bbbb = 6.0;   int a      = 5;   float bbbb = 6.0;   bbbb = 1.0}
```
CDanU: Interesting, Github now (selectively) is showing line ends? 
`0xD` = `13` = `\r` = `CR` = `␍`  ... they could use any one of those but no they'd rather use: 13th letter ->`^M`? 

Well enough ranting. Why is a \r used here instead of \n?

mwoehlke: ...because the test is 'lineEndings-to-Mac.cfg'?

(`^M`, i.e. control-M, is how one types 0x0D in a terminal. And how vi[m] — and I suspect, git — displays a 0x0D in a file with mixed line endings. You can try it yourself; type `echo -e '`, then control-V, control-M, then `' | od -x`... Note that your command line will *look* like `echo -n '^M' | od -c`, but the `^M` will actually be one character.)

CDanU: Ahh ok, everything works as intended then.

Nice example, I knew that it had to do something with the key sequence ctrl+m, so far I never had to actually use that. In my opinion that feels somewhat arcane and it would be more helpful to use any of the other possibilities. As you mentioned they most likely just grab that from git.

tests/config/lineEndings-to-Mac.cfg
gmaurel: any approvment?

Issue #1873, opened 2018-07-11
Unwanted changes inside of comments (tabs to space conversion)


MKoep:
Version is: Uncrustify-0.67_f

Original Code is:
```cs
/// <summary>
/// 	Event handler.
/// </summary>
```
There is a **space** followed by a **tab** before the word "Event".

Output Code is:
```cs
/// <summary>
///     Event handler.
/// </summary>
```
The **tab** is replaced by **spaces**.

There is no option to prevent these changes. The desired output should have **NO** changes inside of comments.

mihaipopescu:
Not that I know of but can you add your config as well?

MKoep:
Sure, this is our config: [config.txt](https://github.com/uncrustify/uncrustify/files/2184271/config.txt)

MKoep:
I think the problem is:
````cpp
// explicitly disallow a tab after a space
if (ch == '\t' && cpd.last_char == ' ')
{
	size_t endcol = next_tab_column(cpd.column);
	while (cpd.column < endcol)
	{
		add_char(' ');
	}
	return;
}
````
in output.cpp, line 275, function _add_char_

mihaipopescu:
Nice. Can you propose the fix in a PR?

guy-maurel:
PR #2248 is a bugfix proposal.
Please test it and report.

PR #1963, opened 2018-08-13
Fix string literal output

oleg-liatte:
Prevent replacing of tabs in string literals to spaces.
Example:
auto test = " 	  	abcdefg  		 ";

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -99,13 +99,13 @@ static void do_kw_subst(chunk_t *pc);
 
 
 //! All output text is sent here, one char at a time.
-static void add_char(UINT32 ch);
+static void add_char(UINT32 ch, bool is_literal = false);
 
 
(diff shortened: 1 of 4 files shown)


oleg-liatte:
It appears that there are tests for the fixed case, and they expectedly fail, as they expect that uncrustify modifies string literals (which is incorrect for my point of view).

CDanU:
Thanks for the PR, I get what you want to solve but I am not entirely convinced if passing an arg through multiple layers is the best solution. 
Honestly I think that add_char should not deal with removing the tab char in the first place. I will see if this situation can be improved somehow.

Related: #1873

gmaurel:
@CDanU May I merge?

CDanU:
Yes, you can actually merge it. I will refactor the output.cpp file soon after that.

PR #1971, opened 2018-08-19
To test bom

guy-maurel:
And the case char_encoding_e::e_UTF8: at write_bom


gmaurel:
You cannot see with vim. Use od instead, or file command.
```
file bom.cpp 
bom.cpp: UTF-8 Unicode (with BOM) text
```
```
od -ch bom.cpp 
0000000 357 273 277   /   /       t   h   e       f   i   l   e       i
           bbef    2fbf    202f    6874    2065    6966    656c    6920
```
Here, the first 3 bytes.

PR #1976, opened 2018-08-21
Change some config files

guy-maurel:
To test the function convert_value with values: t, f, 1, 0


tests/config/oc18.cfg
```diff
@@ -1,6 +1,7 @@
 indent_columns                  = 3
 indent_continue                 = 3
-indent_align_assign             = false
+#indent_align_assign             = false
+indent_align_assign             = true
```
CDanU: Why is the value of the option changed here from false to true?

PR #1977, opened 2018-08-21
Add target to run uncrustify on its own sources

mwoehlke:
Add a custom target (not built by default) to run uncrustify on its own sources. Somewhat like `update-cli-options`, this "assumes" you have a git repository and clobbers the sources in-place (which, for developers, is usually desired).

This makes it much easier to check/enforce code style while hacking on uncrustify. It is also slightly faster, since it is set up as one main target that depends on a sub-target per source, which allows each source to be formatted in parallel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,7 +181,7 @@ add_custom_command(
 #
 # Uncrustify
 #
-add_executable(uncrustify
+set(uncrustify_sources
   src/align.cpp
   src/align_stack.cpp
   src/args.cpp
@@ -223,6 +223,9 @@ add_executable(uncrustify
   src/unicode.cpp
   src/universalindentgui.cpp
   src/width.cpp
+)
+
+set(uncrustify_headers
   src/align_stack.h
   src/args.h
   src/backup.h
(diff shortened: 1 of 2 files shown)


mwoehlke:
I thought also about making this a test, and, while it would be nice to eventually do that, it would actually be *more* complicated to do the in-place reformatting as a test script than this approach.

gmaurel:
What is the difference(s) to using `scripts/Run_uncrustify_for_sources.sh` ?

mwoehlke:
Well, for one, `ninja format-sources` is less typing :wink:.

Joking aside, the main difference is the same as the difference between the CLI options *test* and the `update-cli-options` *target*. One checks, the other makes changes. This target *applies* changes in-place; `Run_uncrustify_for_sources.sh` *checks* if changes are needed.

Also:
- `Run_uncrustify_for_sources.sh` has an assumption about the location of the build directory that fails for my builds. (So, for me, that script just doesn't run.)
- Each file is processed as a separate build edge, which allows reformatting to be done in parallel, which should make it at least slightly faster.
- Doesn't need a POSIX shell (so, no separate `.sh` and `.bat`).

Eventually, I'd like to replace `Run_uncrustify_for_sources.sh` with something using the Python testing framework (and make it a CTest), which will help with those additional points, but that's for later... Anyway, the `format-sources` target would still be useful.

PR #1978, opened 2018-08-21
Fix return brace

mwoehlke:
Add `sp_return_brace`  to control space between return and brace, mirroring `sp_return_paren`. Prevent `mod_paren_on_return` from being applied around braced initializers, as it breaks the code.

See also #1917. (I'm not going to claim this as "fixing" that, since the issue partly deals with how we format uncrustify's own sources, which remains an open question. In particular, even with this, I anticipate needing a change to `forUncrustifySources.cfg` — even if it's just be adding `sp_return_brace=??` — before I would consider #1917 "closed".)

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -2624,6 +2624,19 @@ static chunk_t *process_return(chunk_t *pc)
       }
    }
 
+   // Issue #1917
+   // Never add parens to a braced init list; that breaks the code
+   //   return {args...};    // C++11 type elision; okay
+   //   return ({args...});  // ill-formed
+   if (  language_is_set(LANG_CPP) && chunk_is_token(next, CT_BRACE_OPEN)
+      && next->parent_type == CT_BRACED_INIT_LIST)
+   {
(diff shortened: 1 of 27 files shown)


mwoehlke:
> So the old behaviour was force and now it is ignore

Right.

> I guess that is an acceptable change in terms of breaking changes.

I tend to think so... add/force/remove → ignore won't cause the new version with existing configs to change your source files, though it means it may not force as much formatting as before. The other way — enforcing a different format than before — is obviously not so good... although I guess in this case, in theory, defaulting `sp_return_brace` to `force` would be the same as existing behavior? OTOH we usually prefer to default options to 'no change'.

I don't mind either way; let me know what you think...

mwoehlke:
(Y'all probably know this, but... we usually only default to add/force where remove would be ill-formed, e.g. we can't remove the space in `return 1`. In that sense, the old behavior was something of an aberration; probably because `return {` would have been an error prior to C++11, and so no one thought to allow no space for that case.)

PR #1981, opened 2018-08-22
Add 'nl_for_leave_one_liners' config

Spark-NF:
## Description
Add `nl_for_leave_one_liners` option to preserve one-line `for` statements.
All the logic was copied from other `*_leave_one_liners` options.

## Example
Given `nl_for_brace` is set to `add` or `force`.

**Source:**
```cpp
for (int i = 0; i < 10; ++i) i++; 
```

**Result without:**
```cpp
for (int i = 0; i < 10; ++i)
    i++; 
```

**Result with:**
```cpp
for (int i = 0; i < 10; ++i) i++; 
```


src/options.cpp
```diff
@@ -1945,6 +1945,12 @@ bool &nl_while_leave_one_liners()
 }
 
 
+bool &nl_for_leave_one_liners()
```
mwoehlke: Hehe... yeah, this is a pain right now; sorry. I'm working toward landing #1852, which will make this all *much* simpler. Meanwhile, thanks for putting up with this :slightly_smiling_face:.

PR #1993, opened 2018-08-25
Add a test case for the option -p

guy-maurel:
with a backslash at the end of some lines.


CDanU:
You did overwrite `testSrc.cpp` with `testSrcP.cpp` right? You forgot to also push the changes of `testSrc.cpp`.

guy-maurel:
No.
The input/testSrcP.cpp is used by the test with the option -p, at line 598.
The input/testSrc.cpp is used by the tests with the option -L, at line 614.

Issue #1810, opened 2018-06-16
Uncrustify Issue Reporter Bot


mihaipopescu:
The idea is to have a GitHub app that lets users report formatting bugs using a webapp instead of creating issues here. The app should thereafter create the issues here with a template (similar to what we had used for issues automatically created from our side #1223) or how @CDanU used to write them with: `I:... O:... E:... C:...` blocks (i.e. Input, Output, Expected, Config) (please correct me here).

Anyway, the app should:
- require people all this input fields before creating the issue
- validate that's a real issue (if the issue is fixed in the latest master)
- reduce the user's config
- bisect the issue if the issue is failing now but it was working in an older version
- if the issue is reproducible, the app creates a PR with the incoming test files placed in the right folders and in a `staging.test` testing suite which is ran by the CI builds but here we expect all tests to fail (i.e. tests from the staging will fail if they pass - condition is reversed)
- we fix the issue and the CI will fail with that test passing
- we move the test to a dedicated folder language or to another testing suite `incoming.test` or `regression.test` (what we have in our fork).
- when the test is fixed, the issue will be closed by the app

With this new pipeline:
- increase our testing coverage
- top people from creating non-issues
- help ourselves improving our workflow
- focus more on fixing issues and implementing new features rather than doing a lot of help desk work

What do you guys think?

PS: I remembered about this when I saw a message from @guy-maurel closing an unfinished conversation we had here #1463 that I forgot about. Sorry Daniel about that. You should create a new separate issue with your idea and I'll chip in there.

gmaurel:
It would be fine.

CDanU:
Yes I'd like to have that as well, so far I haven't found the server for that. 
I have to admit that I didn't look that much, there are apparently some services that provide a nodejs server (netlify, zeit.co). The zeit.co also provides a docker service but I don't know how much access we actually get there, those seem to be more geared towards web deployment.

For the ideas that I had I needed to provide those services a key so that they could alter a submitted PRs, I am skeptical about doing that.

PR #2036, opened 2018-10-07
Expected to fail support for tests

mihaipopescu:
As previously proposed, this adds support for staging tests that are expected to fail. This allows reporters and maintainers quickly add bugs as to this suite to be able to separate between reporting and fixing process. This is better for keeping track of known issues as well.

They should not contribute to the coverage but should fail when the test passes. When that happens, the bug should be moved to its dedicated language suite and the related bug closed.

This PR also moves the test added by #2008 to be checked by Uncrustify as expected failure.

This is also a preliminary step for #1810.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 2.8)
 
 find_package(Git QUIET)
 
+# sorted by lexicographic order
 set(test_suites
     c-sharp.test
     c.test
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,16 +3,16 @@ cmake_minimum_required(VERSION 2.8)
 find_package(Git QUIET)
 
 set(test_suites
-    c-sharp.test
     c.test
+    c-sharp.test
     cpp.test
     d.test
-    ecma.test
-    imported.test
-    java.test
     objective-c.test
     pawn.test
     vala.test
+    java.test
+    ecma.test
+    imported.test
 )
 
 option("UNCRUSTIFY_SEPARATE_TESTS"
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -15,6 +15,10 @@ set(test_suites
     imported.test
 )
 
+if (NOT ENABLE_CODECOVERAGE)
+  list(APPEND test_suites staging.test)
+endif()
+
 option("UNCRUSTIFY_SEPARATE_TESTS"
   "Create a separate CTest test for each test case;\
 \ this is slower, especially with Python 3" OFF
@@ -50,24 +54,26 @@ else()
   endforeach()
 endif()
 
-add_test(
-  NAME sources_format
-  COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
-    --executable $<TARGET_FILE:uncrustify>
-    -d --git ${GIT_EXECUTABLE}
-    --result-dir ${CMAKE_CURRENT_BINARY_DIR}
-  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-)
+if (NOT ENABLE_CODECOVERAGE)
+  add_test(
+    NAME sources_format
+    COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
+      --executable $<TARGET_FILE:uncrustify>
+      -d --git ${GIT_EXECUTABLE}
+      --result-dir ${CMAKE_CURRENT_BINARY_DIR}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  )
 
-add_test(
-  NAME cli_options
-  COMMAND ${PYTHON_EXECUTABLE}
-    test_cli_options.py
-    --build ${uncrustify_BINARY_DIR}
-    --diff
-  ${_configs}
-  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cli
-)
+  add_test(
+    NAME cli_options
+    COMMAND ${PYTHON_EXECUTABLE}
+      test_cli_options.py
+      --build ${uncrustify_BINARY_DIR}
+      --diff
+    ${_configs}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cli
+  )
+endif()
 
 add_custom_target(update-cli-options
   COMMAND ${PYTHON_EXECUTABLE}
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -63,17 +63,17 @@ if (NOT ENABLE_CODECOVERAGE)
       --result-dir ${CMAKE_CURRENT_BINARY_DIR}
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   )
-
-  add_test(
-    NAME cli_options
-    COMMAND ${PYTHON_EXECUTABLE}
-      test_cli_options.py
-      --build ${uncrustify_BINARY_DIR}
-      --diff
-    ${_configs}
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cli
-  )
 endif()
+
+add_test(
+  NAME cli_options
+  COMMAND ${PYTHON_EXECUTABLE}
+    test_cli_options.py
+    --build ${uncrustify_BINARY_DIR}
+    --diff
+  ${_configs}
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cli
+)
 
 add_custom_target(update-cli-options
   COMMAND ${PYTHON_EXECUTABLE}
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,16 +3,16 @@ cmake_minimum_required(VERSION 2.8)
 find_package(Git QUIET)
 
 set(test_suites
-    c.test
     c-sharp.test
+    c.test
     cpp.test
     d.test
+    ecma.test
+    imported.test
+    java.test
     objective-c.test
     pawn.test
     vala.test
-    java.test
-    ecma.test
-    imported.test
 )
 
 if (NOT ENABLE_CODECOVERAGE)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -55,16 +55,14 @@ else()
   endforeach()
 endif()
 
-if (NOT ENABLE_CODECOVERAGE)
-  add_test(
-    NAME sources_format
-    COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
-      --executable $<TARGET_FILE:uncrustify>
-      -d --git ${GIT_EXECUTABLE}
-      --result-dir ${CMAKE_CURRENT_BINARY_DIR}
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  )
-endif()
+add_test(
+  NAME sources_format
+  COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
+    --executable $<TARGET_FILE:uncrustify>
+    -d --git ${GIT_EXECUTABLE}
+    --result-dir ${CMAKE_CURRENT_BINARY_DIR}
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+)
 
 add_test(
   NAME cli_options
--- a/tests/run_sources_tests.py
+++ b/tests/run_sources_tests.py
@@ -29,8 +29,7 @@ def main(argv):
         if os.path.splitext(s)[1] in ('.cpp', '.h'):
             t = tu.SourceTest()
             filepath = os.path.join(src_dir, s)
-            t.build(test_input=filepath,
-                    test_lang='CPP', test_config=config,
+            t.build(test_input=filepath, test_lang='CPP', test_config=config,
                     test_expected=filepath)
             tests.append(t)
 
--- a/tests/run_sources_tests.py
+++ b/tests/run_sources_tests.py
@@ -28,8 +28,10 @@ def main(argv):
     for s in os.listdir(src_dir):
         if os.path.splitext(s)[1] in ('.cpp', '.h'):
             t = tu.SourceTest()
-            t.build(test_input=os.path.join(src_dir, s),
-                    test_lang='CPP', test_config=config)
+            filepath = os.path.join(src_dir, s)
+            t.build(test_input=filepath,
+                    test_lang='CPP', test_config=config,
+                    test_expected=filepath)
             tests.append(t)
 
     counts = tu.run_tests(tests, args)
--- a/tests/test_uncrustify/config.py
+++ b/tests/test_uncrustify/config.py
@@ -9,18 +9,19 @@ import os
 
 test_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
+# sorted by lexicographic order
 all_tests = [
     'c-sharp',
     'c',
     'cpp',
     'd',
-    'java',
-    'pawn',
-    'objective-c',
-    'vala',
     'ecma',
     'imported',
-    'staging'
+    'java',
+    'objective-c',
+    'pawn',
+    'staging',
+    'vala',
 ]
 
 FAIL_ATTRS     = {'bold': True}
--- a/tests/test_uncrustify/config.py
+++ b/tests/test_uncrustify/config.py
@@ -10,14 +10,14 @@ import os
 test_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 all_tests = [
-    'c-sharp',
     'c',
+    'c-sharp',
     'cpp',
     'd',
-    'java',
-    'pawn',
     'objective-c',
+    'pawn',
     'vala',
+    'java',
     'ecma',
     'imported'
 ]
--- a/tests/test_uncrustify/config.py
+++ b/tests/test_uncrustify/config.py
@@ -19,7 +19,8 @@ all_tests = [
     'vala',
     'java',
     'ecma',
-    'imported'
+    'imported',
+    'staging'
 ]
 
 FAIL_ATTRS     = {'bold': True}
--- a/tests/test_uncrustify/config.py
+++ b/tests/test_uncrustify/config.py
@@ -10,14 +10,14 @@ import os
 test_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 all_tests = [
-    'c',
     'c-sharp',
+    'c',
     'cpp',
     'd',
-    'objective-c',
-    'pawn',
-    'vala',
     'java',
+    'pawn',
+    'objective-c',
+    'vala',
     'ecma',
     'imported',
     'staging'
--- a/tests/test_uncrustify/failure.py
+++ b/tests/test_uncrustify/failure.py
@@ -58,7 +58,7 @@ class UnstableFailure(Failure):
             self.actual_path, self.expected_path)
 
 # =============================================================================
-class UnexpectedlyPassing(Failure):
+class UnexpectedlyPassingFailure(Failure):
     # -------------------------------------------------------------------------
     def __init__(self, expected, actual):
         self.expected_path = expected
--- a/tests/test_uncrustify/failure.py
+++ b/tests/test_uncrustify/failure.py
@@ -8,6 +8,18 @@
 class Failure(Exception):
     pass
 
+# =============================================================================
+class TestDeclarationParseError(Failure):
+    # -------------------------------------------------------------------------
+    def __init__(self, test_suite, line_number):
+        self.test_suite = test_suite
+        self.line_number = line_number
+
+    # -------------------------------------------------------------------------
+    def __str__(self):
+        return 'Error parsing line {!r} from the {!r} test suite'.format(
+            self.line_number, self.test_suite)
+
 
 # =============================================================================
 class ExecutionFailure(Failure):
--- a/tests/test_uncrustify/failure.py
+++ b/tests/test_uncrustify/failure.py
@@ -66,7 +66,7 @@ class UnstableFailure(Failure):
 
     # -------------------------------------------------------------------------
     def __str__(self):
-        return 'Output {!r} does not match expected output {!r}'.format(
+        return 'Output {!r} unexpectedly matches expected output {!r}'.format(
             self.actual_path, self.expected_path)
 
 # =============================================================================
--- a/tests/test_uncrustify/failure.py
+++ b/tests/test_uncrustify/failure.py
@@ -66,7 +66,7 @@ class UnstableFailure(Failure):
 
     # -------------------------------------------------------------------------
     def __str__(self):
-        return 'Output {!r} unexpectedly matches expected output {!r}'.format(
+        return 'Output {!r} does not match expected output {!r}'.format(
             self.actual_path, self.expected_path)
 
 # =============================================================================
@@ -78,5 +78,5 @@ class UnexpectedlyPassingFailure(Failure):
 
     # -------------------------------------------------------------------------
     def __str__(self):
-        return 'Output {!r} was not expected to match output {!r}'.format(
+        return 'Output {!r} unexpectedly matches expected output {!r}'.format(
             self.actual_path, self.expected_path)
--- a/tests/test_uncrustify/failure.py
+++ b/tests/test_uncrustify/failure.py
@@ -66,5 +66,5 @@ class UnexpectedlyPassingFailure(Failure):
 
     # -------------------------------------------------------------------------
     def __str__(self):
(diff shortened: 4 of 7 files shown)


mihaipopescu:
We should remove uncrustify itself from contributing to the coverage. Otherwise it will continue to report loss in coverage after code refactoring.

tests/CMakeLists.txt
```diff
@@ -50,14 +54,16 @@ else()
...
 
-add_test(
-  NAME sources_format
-  COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
-    --executable $<TARGET_FILE:uncrustify>
-    -d --git ${GIT_EXECUTABLE}
-    --result-dir ${CMAKE_CURRENT_BINARY_DIR}
-  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-)
+if (NOT ENABLE_CODECOVERAGE)
```
mwoehlke-kitware: Um... okay, I guess, but what's the problem this is fixing?

tests/run_sources_tests.py
```diff
@@ -28,8 +28,10 @@ def main(argv):
     for s in os.listdir(src_dir):
         if os.path.splitext(s)[1] in ('.cpp', '.h'):
             t = tu.SourceTest()
-            t.build(test_input=os.path.join(src_dir, s),
-                    test_lang='CPP', test_config=config)
+            filepath = os.path.join(src_dir, s)
+            t.build(test_input=filepath,
```
mwoehlke-kitware: (Nit: saving a line and putting `test_input` and `test_expected` on the same line would be nice :slightly_smiling_face:.)

tests/test_uncrustify/config.py
```diff
@@ -10,16 +10,17 @@
 test_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
 
 all_tests = [
-    'c-sharp',
```
mwoehlke-kitware: Same comment as CMakeLists.txt

tests/test_uncrustify/failure.py
```diff
@@ -56,3 +56,15 @@ def __init__(self, expected, actual):
     def __str__(self):
         return 'Output {!r} does not match expected output {!r}'.format(
             self.actual_path, self.expected_path)
+
+# =============================================================================
+class UnexpectedlyPassing(Failure):
```
mwoehlke-kitware: Pedantic: this is still a test "failure", "UnexpectedlyPassingFailure" might be better :slightly_smiling_face:, even if it *is* a mouthful.

tests/test_uncrustify/failure.py
```diff
@@ -56,3 +56,15 @@ def __init__(self, expected, actual):
...
+    # -------------------------------------------------------------------------
+    def __init__(self, expected, actual):
+        self.expected_path = expected
+        self.actual_path = actual
+
+    # -------------------------------------------------------------------------
+    def __str__(self):
+        return 'Output {!r} was not expected to match expected output {!r}'.format(
```
mwoehlke-kitware: Pedantic: Per [PEP-8](https://www.python.org/dev/peps/pep-0008/#id19)¹ this line is too long (>79 chars). Please try:
```python
m = 'Output {!r} was not expected to match expected output {!r}'
return m.format(self.actual_path, self.expected_path)
```
(...which, conveniently, is the same number of lines. Feel free to use `msg` or whatever you prefer instead of `m`, so long as it fits.)

(¹ I try to mostly follow `pycodestyle` except for places where it is actively "detrimental", e.g. it forbids alignment of assignments.)

tests/test_uncrustify/test.py
```diff
@@ -108,18 +112,30 @@ def run(self, args):
         except subprocess.CalledProcessError as exc:
             msg = '{}: Uncrustify error code {}'
             msg = msg.format(self.test_name, exc.returncode)
-            printc('FAILED: ', msg, **FAIL_ATTRS)
-            raise ExecutionFailure(exc)
+            if not self.test_xfail:
```
mwoehlke-kitware: I was wondering, earlier, if we need to distinguish between tests that are supposed to produce bad output vs. tests that cause uncrustify to *fail*... I think, for cases where failing is the *correct* behavior, we should be using front-end tests, so any failure of a *staging* test is a known bug. That being the case, I am unconvinced that (always) printing a message for these cases is TRTTD; it should just be treated as a pass (IOW, tweak `run_tests` to print either `PASSED` or `XFAILED`, but here, just `return`.)

tests/test_uncrustify/test.py
```diff
@@ -108,18 +112,30 @@ def run(self, args):
...
+                printc('FAILED: ', msg, **FAIL_ATTRS)
+                raise ExecutionFailure(exc)
+            else:
+                printc('XFAILED: ', msg, **FAIL_ATTRS)
+                return
         finally:
             del stderr
 
         try:
-            if not filecmp.cmp(_expected, _result):
```
mwoehlke-kitware: Thinking about the previous comment, I think XFAIL diffs should not be printed "by default". We could add a separate option, but I think it would make sense to print XFAIL diffs iff both `-d` (`args.diff`) and `-p` (`args.show_all`) are passed?

tests/test_uncrustify/test.py
```diff
@@ -210,6 +231,7 @@ def build_from_args(self, args):
         self.test_expected = args.expected
         self.test_rerun_config = args.rerun_config or args.config
         self.test_rerun_expected = args.rerun_expected or args.expected
+        self.test_xfail = False
```
mwoehlke-kitware: It would be nice to not break `UNCRUSTIFY_SEPARATE_TESTS`...

tests/test_uncrustify/test.py
```diff
@@ -176,7 +193,10 @@ def build_from_declaration(self, decl, group):
         else:
             self.test_lang = test_dir
 
-        if parts[0].endswith('!'):
```
mwoehlke-kitware: I'm okay with this. However, this could also be rewritten:
```python
m = re.match(r'(\d+)([~!]*)', parts[0]
num = m.group(1)
is_rerun = ('!' in m.group(2))
is_xfail = ('~' in m.group(2))
```

...which do you like better?

tests/CMakeLists.txt
mwoehlke-kitware: I missed this:

> We should remove uncrustify itself from contributing to the coverage. Otherwise it will continue to report loss in coverage after code refactoring.

I'm not sure I understand why "it will continue to report loss in coverage after code refactoring". I *can* see an argument that we should not include coverage for what uncrustify does operating on its own sources, "because those aren't 'tests'", but OTOH they sort-of are, because we do have 'expected output' and raise an error if the expected and actual don't match.

tests/test_uncrustify/test.py
mihaipopescu: I don't really see the connection. I was just setting a default value and don't expect anything to fail. Otherwise, are you thinking in sending it as an arg?

tests/test_uncrustify/test.py
mihaipopescu: yeah, hard to get it right. If we know that it should fail then nothing should be printed but uncrustify does prints to console by default (we're not calling with `check_output`). But then if we do that and want to capture everything then you lose the stdout/stderr interleaving which probably can be acceptable. But till that happens, I went with printing also a message to explain why something was showing as a failure but it's passing because it's an expected failure.

tests/test_uncrustify/test.py
mihaipopescu: for `XFAIL` you just know that there will be a diff, probably different every time uncrustify changes, but still not the output you are expecting so they are kinda irrelevant. If we find the diff relevant we can always print it on some option.

tests/test_uncrustify/failure.py
mihaipopescu: I had a second thought exactly about this. Sure, let's be pedantic.

tests/test_uncrustify/failure.py
mihaipopescu: Agree, we should enforce this with https://github.com/google/yapf or flake8 for instance and tell everyone how to install the linters locally to be able to check python formatting before pushing.

tests/CMakeLists.txt
```diff
@@ -3,18 +3,22 @@ cmake_minimum_required(VERSION 2.8)
...
     d.test
-    ecma.test
-    imported.test
-    java.test
     objective-c.test
     pawn.test
     vala.test
+    java.test
+    ecma.test
+    imported.test
```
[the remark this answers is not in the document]
mihaipopescu: I did not know about that "standard" and I went for what uncrustify considered to be the order of language ID numbers. Maybe a comment would have helped but this is the commit https://github.com/uncrustify/uncrustify/pull/2036/commits/a0434b522c020ba5664224d471b639bf1005d63f. I can reverse it.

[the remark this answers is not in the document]
mihaipopescu: Same time, a comment would have stopped me from sorting them.

tests/CMakeLists.txt
mihaipopescu: Imagine you're doing some refactoring and removing some code. That piece of code was triggering some other part from uncrustify and that was the only thing that was covering. Suddenly you're getting an unexpected drop in coverage and forced to add some coverage to a totally unrelated piece of uncrustify's code base.

With tests is different. There is a clear input/expected output and nothing should be changing underneath your feet while you are changing the code. The input/expected should remain static in this process.

mihaipopescu:
Thanks @mwoehlke-kitware for feedback. Much appreciated.

tests/CMakeLists.txt
mwoehlke-kitware: Okay, that makes sense. Thanks for the explanation!

tests/test_uncrustify/test.py
mwoehlke-kitware: > but uncrustify does prints to console by default

Hmm... okay, I see the "problem"... This does make me inclined to think we should capture stdout/stderr for an xfail test (or maybe even all tests?) and suppress it. At least, that would be the way to implement my knee-jerk preference, which is to continue to have no output unless a test fails or `-p` is given.

Presumably, `stdout=subprocess.PIPE, stderr=subprocess.STDOUT` would preserve interleaving? (OTOH, do we care about redirecting the handles of `run_tests.py` having 1:1 mapping to redirecting uncrustify's handles?)

tests/test_uncrustify/test.py
mwoehlke-kitware: Right... IOW, it would be nice to see the needed changes to `print_as_ctest` and `add_test_arguments` also.

tests/test_uncrustify/failure.py
mwoehlke-kitware: :smile:

tests/test_uncrustify/failure.py
mwoehlke-kitware: I'm not familiar with yapf; I've always just used [pycodestyle](https://github.com/PyCQA/pycodestyle), but a) I'm ignoring E221, and b) I like separator lines (all the `# ===` and `# ---` comments), though I am using the latter in a way that pycodestyle tolerates.

At least one advantage of pycodestyle is that it is packaged on Fedora; yapf isn't. (Ubuntu seems to have both.) It's also dead simple to run: `pycodestyle tests/test_uncrustify/*.py`. (Or on any other Python files that are trying to be styled... Some of the old ones are a mess, but anything I've hacked on significantly should be conforming, aside from a couple E221 hits.)

I'd be okay with using flake8 if available; it uses pycodestyle under the hood, so it gives similar results, and folks that don't want to install all of it can use just pycodestyle.

tests/test_uncrustify/test.py
mihaipopescu: > unless a test fails or -p is givenI agree with that

mwoehlke-kitware: Ah... well, I'm going to be fiddling in here soon-ish anyway (actually I have a WIP to integrate the front-end tests). If you remind me, I can probably do that conversions while I'm at it :slightly_smiling_face:.

mihaipopescu: Yeah, I'm unsure. In the last update everything looks like it's working, even for debug output which is a lot of text.

mihaipopescu:
@mwoehlke-kitware are you satisfied with the last iteration? You can do other change after we land this one if you don't have the time to do it now.

mwoehlke-kitware:
Aside: would it make sense to squash-merge this?

tests/test_uncrustify/test.py
```diff
@@ -94,36 +100,52 @@ def run(self, args):
...
             msg = msg.format(self.test_name, exc.returncode)
-            printc('FAILED: ', msg, **FAIL_ATTRS)
-            raise ExecutionFailure(exc)
+            if not self.test_xfail:
+                printc(output)
+                printc('FAILED: ', msg, **FAIL_ATTRS)
+                raise ExecutionFailure(exc)
+            else:
+                if args.xdiff:
+                    printc(output)
```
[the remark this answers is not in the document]
mwoehlke-kitware: (Er... note that `end=''` requires `from __future__ import print_function`...)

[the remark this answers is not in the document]
mwoehlke-kitware: (Another option, possibly better, would be `print(output.rstrip())`.)

tests/test_uncrustify/test.py
```diff
@@ -94,36 +100,52 @@ def run(self, args):
...
+                    printc('XFAILED: ', msg, **FAIL_ATTRS)
```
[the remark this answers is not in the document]
mihaipopescu: it's already guarded by `--xdiff`

tests/test_uncrustify/utilities.py
```diff
@@ -128,22 +133,26 @@ def run_tests(tests, args, selector=None):
             continue
 
         try:
-            test.run(args)
+            result = test.run(args)
```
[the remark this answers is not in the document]
mihaipopescu: oh yeah, was a leftover, removed

tests/test_uncrustify/test.py
```diff
@@ -234,6 +268,7 @@ def to_cmake_path(path):
              '    --expected       "{test_expected}"\n' +
              '    --rerun-config   "{test_rerun_config}"\n' +
              '    --rerun-expected "{test_rerun_expected}"\n' +
+             '    --xfail          "{test_xfail}"\n' +
```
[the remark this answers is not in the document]
mihaipopescu: oh, I didn't noticed that... sorry for rushing it.

mihaipopescu:
@mwoehlke-kitware thanks for your comments, I addressed them.

tests/test_uncrustify/test.py
```diff
@@ -7,14 +7,16 @@
...
 import subprocess
 import sys
 
 from .ansicolor import printc
-from .config import (config, test_dir, FAIL_ATTRS,
+from .config import (config, test_dir, FAIL_ATTRS, PASS_ATTRS,
                      MISMATCH_ATTRS, UNSTABLE_ATTRS)
-from .failure import (ExecutionFailure, MissingFailure,
-                      MismatchFailure, UnstableFailure)
+from .failure import (ExecutionFailure, MismatchFailure, MissingFailure,
```
mwoehlke-kitware: :+1: Thanks!

tests/test_uncrustify/test.py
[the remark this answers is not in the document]
mwoehlke-kitware: Maybe... or bite the bullet and give it the axe. I'm on the fence; it's potentially advantageous to be able to pick individual tests at the CTest level, but OTOH I've adjusted my own workflow to use the test runner(s) directly. (It's too bad Python 3 imports are so slow :cry:.)

tests/test_uncrustify/test.py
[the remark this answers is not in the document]
mwoehlke-kitware: Sure, but, if I pass both `--xdiff` and `--show-all`, isn't "XFAILED" going to be printed twice per test? Once here, and once by `run_tests`? (I could be reading the code wrong...)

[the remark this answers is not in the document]
mihaipopescu: mm, that could be correct.. I haven't checked to be honest and I forgot to do it.

mihaipopescu:
@mwoehlke-kitware ok, pushed a fix to the last fix we talked about (although slightly different). Let me know if there is anything else.

tests/test_uncrustify/test.py
```diff
@@ -151,6 +152,10 @@ class FormatTest(SourceTest):
     pass_input = ['test_input', 'test_expected']
     pass_expected = ['test_expected', 'test_rerun_expected']
 
+    re_test_declaration = re.compile(r'^(?P<num>\d+)(?P<mark>[~!])?\s+'
```
mwoehlke-kitware: :+1: Thanks!

tests/test_uncrustify/test.py
```diff
@@ -253,7 +254,9 @@ def print_as_ctest(self, out_file=sys.stdout):
         def to_cmake_path(path):
             if type(path) is dict:
                 return {k: to_cmake_path(v) for k, v in path.items()}
-            return path.replace(os.sep, '/')
+            if type(path) is str:
```
mwoehlke-kitware: I take it this is what was "already broken"? What is the type of `path` if *not* dict or str? (I also wonder if this is correct for Python 2 vs. 3?)

Note: I'd be perfectly happy to drop this from this PR and follow up on it separately.

tests/test_uncrustify/utilities.py
```diff
@@ -181,11 +181,8 @@ def read_format_tests(filename, group):
     tests = []
 
     print("Processing " + filename)
-    line_number = 0
     with open(filename, 'rt') as f:
-        for line in f:
-            line_number += 1
-
+        for line_number, line in enumerate(f, 1):
```
mwoehlke-kitware: :smile:

tests/test_uncrustify/utilities.py
```diff
@@ -133,10 +133,10 @@ def run_tests(tests, args, selector=None):
             continue
 
         try:
-            result = test.run(args)
+            test.run(args)
             if args.show_all:
                 outcome = 'XFAILED' if test.test_xfail else 'PASSED'
-                printc('%s: ' % outcome, test.test_name, **PASS_ATTRS)
+                printc('%s: '.format(outcome), test.test_name, **PASS_ATTRS)
```
mwoehlke-kitware: `s/%s/{}/` :slightly_smiling_face:.

tests/test_uncrustify/test.py
```diff
@@ -100,31 +99,33 @@ def run(self, args):
...
-                printc(exc.output)
+                print(output)
                 printc('FAILED: ', msg, **FAIL_ATTRS)
                 raise ExecutionFailure(exc)
             else:
                 if args.xdiff:
-                    printc(exc.output)
-                    printc('XFAILED: ', msg, **FAIL_ATTRS)
+                    print(output)
+                if not args.show_all:
```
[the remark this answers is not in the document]
mihaipopescu: To preserve consistency, let's print `XFAILED` on `--xdiff` after the diff but suppress it on `--show-all` because it will be printed in `run_tests`.

tests/test_uncrustify/test.py
mihaipopescu: yeah, it's broken because of `self.test_xfail = False` is not a string referring to a path.

mihaipopescu:
@mwoehlke-kitware 
> Open question: does an XFAIL test still do two passes? 

yes, `XFAIL` tests will have two passes as well to ensure stability in failures as well. You don't want to have a failure and then a pass in the second pass due to an instabilitiy.

tests/test_uncrustify/test.py
```diff
@@ -140,6 +140,8 @@ def run(self, args):
             if has_diff and self.test_xfail:
                 if args.xdiff:
                     self._diff(_expected, _result)
+                    if not args.show_all:
```
mwoehlke-kitware: This still isn't *quite* consistent... for a normal test, we print the status, then the diff, but here we print the diff, then the status. Maybe normal tests should flip? (I.e. swap lines 133-134 and 135-136...)

Otherwise, in this file we should "always" print the status (i.e. don't check `args.show_all`) and instead in `run_tests`, don't print the statuf if `args.xdiff`.

mihaipopescu: I don't know anymore how we print stuff for a normal test. I think first we need a test framework for the test framework before changing anything here.

I feel like we've tweaked the output 1000 times back and forth... "use these args, no, let's flip them back".  I'm ok in changing them but I don't feel like we're going anywhere with this. We're talking about rare use-cases, combination of flags that almost no one will use.

I think uncrustify could be better off with this PR landed than with a perfect consistent console output handling all combinations of all possible args.

tests/test_uncrustify/test.py
```diff
@@ -94,36 +100,52 @@ def run(self, args):
...
+                printc('FAILED: ', msg, **FAIL_ATTRS)
+                raise ExecutionFailure(exc)
+            else:
+                if args.xdiff:
+                    printc(output)
+                    printc('XFAILED: ', msg, **FAIL_ATTRS)
+                return
         finally:
-            del stderr
+            if args.debug:
```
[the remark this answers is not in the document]
mihaipopescu: I think it's yes to both.

tests/test_uncrustify/test.py
```diff
@@ -94,36 +100,52 @@ def run(self, args):
...
-            if not filecmp.cmp(_expected, _result):
+            has_diff = not filecmp.cmp(_expected, _result)
+            if has_diff and not self.test_xfail:
                 printc('{}: '.format(self.diff_text),
                        self.test_name, **self.diff_attrs)
                 if args.diff:
                     self._diff(_expected, _result)
                 raise self.diff_exception(_expected, _result)
+            if not has_diff and self.test_xfail:
+                raise UnexpectedlyPassingFailure(_expected, _result)
```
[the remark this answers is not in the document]
mihaipopescu: I think we discussed this previously.

mihaipopescu:
Is there anything else left on the table @mwoehlke-kitware ?

mihaipopescu:
@gmaurel @CDanU any other opinions?

I have another test that I want to add to the staging tests after this PR lands.

mwoehlke-kitware:
Apologies; I am still catching up after WG21... Looking now...

tests/test_uncrustify/failure.py
```diff
@@ -54,5 +66,17 @@ def __init__(self, expected, actual):
 
     # -------------------------------------------------------------------------
     def __str__(self):
-        return 'Output {!r} does not match expected output {!r}'.format(
+        return 'Output {!r} unexpectedly matches expected output {!r}'.format(
```
mwoehlke-kitware: Ack... this text was meant to appear at L81, no? I think this change was unintended.

tests/test_uncrustify/test.py
```diff
@@ -46,12 +48,13 @@ def _diff(self, expected, actual):
         subprocess.call(cmd)
 
     # -------------------------------------------------------------------------
-    def build(self, test_input, test_lang, test_config):
+    def build(self, test_input, test_lang, test_config, test_expected):
```
mwoehlke-kitware: BTW, remind me why is this change needed? AFAICT this method is only used by `run_sources_tests.py` which always passes the same value. (Was this left over from an earlier version?)

mwoehlke-kitware:
Okay, so we can put this to bed :slightly_smiling_face:, I pushed a few minor changes to https://github.com/mwoehlke-kitware/uncrustify/tree/testing/xfail-support; PTAL. Also please feel free to drop especially the last commit (enable both XFAIL and special-rerun); I think it makes sense to do that if XFAIL tests are doing both passes, but we could also always add it later if needed.

I'm going to run some tests to verify things are working as expected; if you don't hear back from me, assume it's good.

p.s. Does it make sense to squash-merge this? There is a lot of history that I'm not sure is actually useful...

mwoehlke-kitware:
Okay, if it works for you, AFAICT with my changes the status is *always* printed last. Note that I did find a minor issue; please see the last commit on my branch.

tests/test_uncrustify/test.py
mihaipopescu: The change is for clarity. Better to have that in the caller.

tests/test_uncrustify/failure.py
mihaipopescu: true, I wonder how this one slipped (shrug)

mihaipopescu:
@mwoehlke-kitware I merged commits from your branch. It's fine to squash history as well.

tests/test_uncrustify/test.py
mwoehlke-kitware: I guess that is okay, then...

mwoehlke-kitware:
> Should it be included in version 0.68 ?

@gmaurel, I suppose that is up to you; I don't think it will hurt either way. Are we blocking merges not intended for 0.68? If so, any objection to this being merged? (If there is none, I will plan to squash-merge tomorrow at latest.)

mwoehlke-kitware:
@mihaipopescu, thanks again!

mihaipopescu:
Thank you for your feedback @mwoehlke 👍

PR #2040, opened 2018-10-14
Remove Uncrustify's own source code from counting to test coverage

mihaipopescu:
While formatting its own source code can be regarded as a test in itself
it is not correct from the coverage point of view.

Imagine you're doing some refactoring and removing some code. That piece
of code was triggering some other part from uncrustify and that was the
only thing that was covering. Suddenly you're getting an unexpected drop
in coverage and forced to add some coverage to a totally unrelated piece
of uncrustify's code base.

With normal tests is different. There is a clear input/expected output
and nothing should be changing underneath your feet while you are
changing the code. The input/expected remain static in this process.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -50,14 +50,16 @@ else()
   endforeach()
 endif()
 
-add_test(
-  NAME sources_format
-  COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
-    --executable $<TARGET_FILE:uncrustify>
-    -d --git ${GIT_EXECUTABLE}
-    --result-dir ${CMAKE_CURRENT_BINARY_DIR}
-  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-)
+if (NOT ENABLE_CODECOVERAGE)
+  add_test(
+    NAME sources_format
+    COMMAND ${PYTHON_EXECUTABLE} run_sources_tests.py
+      --executable $<TARGET_FILE:uncrustify>
+      -d --git ${GIT_EXECUTABLE}
+      --result-dir ${CMAKE_CURRENT_BINARY_DIR}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  )
+endif()
 
 add_test(
   NAME cli_options


mihaipopescu:
I think it is necessary even for code owners to avoid wasting time wondering why coverage drops for unknown reasons. I'm trying to give an example with #2042.

mihaipopescu:
@gmaurel You can see in #2042 that a minor newline change in `enum_flags.h` produces a drop in coverage in `newlines.cpp`. Even if we can force the merge as code owners, you would have to go through the same process of figuring why the coverage dropped. That means time wasted for us and for the other contributors.

 Does that makes more sense now?

mwoehlke:
I'm going to weigh in as "neutral". From a technical perspective the change is okay, Whether or not it's a good idea philosophically is, obviously, debatable.

FWIW though, I'll throw out that one might argue that it would be preferable to have unit tests covering "everything", and not "incidental" coverage via running uncrustify on itself. (On a related note, it would be nice if this didn't cause a coverage drop, e.g. if it was accompanied by new unit tests to cover what is currently only covered by running uncrustify on its own sources.)

mihaipopescu:
> it would be nice if this didn't cause a coverage drop, e.g. if it was accompanied by new unit tests to cover what is currently only covered by running uncrustify on its own sources

That requires additional investigation work to identify what part of the code accounts for the `0,05%` drop in coverage. I can do it for the example from #2042 but the rest of coverage drop of `0,02%` needs to be investigated.

But I'm waiting for any arguments, philosophical or not, against this change.

guy-maurel:
I might have passed something! I might be not enought "up to date".
I thought the use of coveralls is to show us which lines of code are NOT used by the complete test suite we have.
I found with it many parts of code which were not necessary, add some more tests to the suite.
I found it very usefull.
If I follow your proposal, what it the use of coveralls?

mihaipopescu:
As you may know, we are ensuring that uncrustify's own source code is formatted and that is a test that is part of the overall coverage. This is an unreliable test to put in coverage because it's always changing. It's not like a test that has static input and output. One has to change uncrustify source code to fix a bug for instance. For example, because we have the source code to count in coverage, #2042 shows that a change in uncrustify can cause an unexpected drop in coverage in a totally different location for unknown reasons. It would take some time to understand why the coverage drops. This is where this PR comes and takes out the source code from counting in the test coverage. We are not removing the coverage reporting or anything else. We are making a conscious deliberate decision to remove something from coverage to avoid further time wasted. We are talking about 0.05% drop in coverage coming from uncrustify's own source code. I hope that sheds more light.

Issue #1702, opened 2018-04-03
Extraneous newlines added after 7cc8012


smaeul:
Commit 7cc8012254f8dfad8517efcce7dde7f8a92d7a37 "Change XY->type == CT_abc to chunk_is_token(XY, CT_abc)" included the following chunk:
```patch
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -2949,9 +2946,7 @@ static void fix_enum_struct_union(chunk_t *pc)
       next = chunk_get_next_ncnl(next);
    }

-   if (  next != nullptr
-      && !prev
-      && next->type == CT_SEMICOLON)
+   if (chunk_is_token(next, CT_SEMICOLON))
    {
       set_chunk_parent(next, pc->type);
    }
```
which is not equivalent to the original code. It was safe to drop the `next != nullptr` check, but not the `!prev` check.

After updating `uncrustify`, the program now puts a blank line after every `struct` variable or member declaration, e.g.:
```patch
diff --git a/common/dm.c b/common/dm.c
index 9f7f9b5..d4a2dd0 100644
--- a/common/dm.c
+++ b/common/dm.c
@@ -12,6 +12,7 @@
 #include <string.h>

 extern struct device device_list[];
+
 extern struct device device_list_end[];

 static int
@@ -90,6 +91,7 @@ dm_setup_pins(struct device *dev, uint8_t num_pins)
 {
        struct gpio_handle *pins = dev->gpio_pins;
        struct device *gpio_dev;
+
        uint8_t pin_id, mode;
        int     err;

diff --git a/common/scpi.c b/common/scpi.c
index b389349..9572f51 100644
--- a/common/scpi.c
+++ b/common/scpi.c
@@ -28,11 +28,13 @@

 struct scpi_mem {
        struct scpi_msg tx_msg; /**< The reply to be sent to a client. */
+
        struct scpi_msg rx_msg; /**< The request received from a client. */
 };

 struct scpi_buffer {
-       struct scpi_mem mem;    /**< Memory for the request/reply messages. */
+       struct scpi_mem mem; /**< Memory for the request/reply messages. */
+
        uint8_t         client; /**< Client that should receive the reply. */
        bool            busy;   /**< Flag telling if this buffer is in use. */
 };
@@ -205,7 +207,8 @@ static void
 scpi_receive_message(struct device *dev __unused, uint8_t client, uint32_t msg)
 {
        struct scpi_buffer *buffer;
[...]

PR #2057, opened 2018-11-07
Fix nl_after_struct option

larixsoft:
In latest code, the nl_after_struct was broken due to a fix to avoid newlines for forward declarations for struct type variables(Issue #1702 as mentioned in the comment). 
The fix is to check both PCF_IN_STRUCT and PCF_VAR_DEF to determine it is a variable or not.


larixsoft:
will fix the failed test

PR #2060, opened 2018-11-07
Don't remove the space(s)

guy-maurel:
if sp_before_semi_for_empty is set to ignore.
The expected data for test 30713 must be corrected.
The two spaces before the first semi (at the for statement) must stay.


guy-maurel:
ref. #2052

Issue #2064, opened 2018-11-09
indent_relative_single_line_comments endless spaces after if vbrace


antcodd:
``indent_relative_single_line_comments = true`` is working incorrectly for comments after a non-braced if statement (probably any vbrace expression), instead adding large amounts of space for each run.

In Uncrustify 0.67 and at least as far back as 0.59 this happens regardless of the number of spaces.
With git version Uncrustify-0.67-525-ab9eaa20 at least 2 spaces are needed before the comment.

I suspect this is something to do with offset calculations and the fact VBRACE_CLOSE pops the stack.

# uncrustify config
```
# If set, will indent trailing single line ('//') comments relative
# to the code instead of trying to keep the same absolute column.
indent_relative_single_line_comments = true     # false/true
```
# Input
```
void do_something(int foo, int foobar)
{
    if (foobar < foo)
        foobar++;  /* comment after VBRACE_END with >1 spaces */
}

```
# Output
```
Parsing: test-uncrustify.c as language C
void do_something(int foo, int foobar)
{
        if (foobar < foo)
                foobar++;                     /* comment after VBRACE_END with >1 spaces */
}
```
For each run the amount of space increases substantially.

```
Line      Tag          Parent     Columns  Br/Lvl/pp Flag Nl  Text
  1>          TYPE[     FUNC_DEF][ 1/ 1/ 5][0/0/0][     70000][0-0] void
  1>      FUNC_DEF[         NONE][ 6/ 6/18][0/0/0][         0][0-0]      do_something
  1>   FPAREN_OPEN[     FUNC_DEF][18/18/19][0/0/0][ 100000000][0-0]                  (
  1>          TYPE[         NONE][19/19/22][0/1/0][    450008][0-0]                   int
  1>          WORD[         NONE][23/23/26][0/1/0][    800008][0-0]                       foo
  1>         COMMA[         NONE][26/26/27][0/1/0][ 100000008][0-0]                          ,
  1>          TYPE[         NONE][28/28/31][0/1/0][    450008][0-0]                            int
  1>          WORD[         NONE][32/32/38][0/1/0][    800008][0-0]                                foobar
  1>  FPAREN_CLOSE[     FUNC_DEF][38/38/39][0/0/0][ 100000000][0-0]                                      )
  1>       NEWLINE[         NONE][39/39/ 1][0/0/0][         0][1-0]
[...]

PR #2066, opened 2018-11-10
[DO NOT MERGE]: If token is CT_VBRACE_CLOSE, we must take …

guy-maurel:
…the token before to calculate the value for column.
Ref. #2064

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -354,7 +354,7 @@ void reindent_line(chunk_t *pc, size_t column)
          const auto tmp_col = static_cast<int>(pc->column) + col_delta;
          pc->column = max(tmp_col, static_cast<int>(min_col));
 
-         LOG_FMT(LINDLINED, "   set column of ");
+         LOG_FMT(LINDLINED, "   %s(%d): set column of ", __func__, __LINE__);
          if (chunk_is_token(pc, CT_NEWLINE))
          {
             LOG_FMT(LINDLINED, "<Newline>");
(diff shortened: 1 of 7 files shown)


antcodd:
Is there a more generic solution to this problem than special casing in space.c? It's quite possible there could be other cases. For example I report another one in the same bug #2064 with VBRACE_OPEN, and I've seen infinite growing in unsupported cases (like nested struct table alignment).

Why doesn't the original code break with normal ``{`` braces? Whatever was used there should ideally be used here rather than special casing it. Why does VBRACE_CLOSE in indent.c update pc? I think that may be the source of the inconsistency.

The way uncrustify is written, the fallback and complicated mix of option cases seems to be to mess up the alignment through invalid offset calculations rather than leaving the line as is relative to the basic indent. Maybe something like not popping indent token alignment column changes until after a newline, or a another method of offset calculation (such as separating the concepts of line indent from extra column offset from the previous token in the same line)? There seems to be some infrastructure to this kind of effect in indent.c but there are so many special cases it's hard to follow exactly how it actually works.

mihaipopescu:
@antcodd Thank you for your comment about uncrustify's design.

> Is there a more generic solution to this problem than special casing in space.c?

Now without a massive rewrite but I don't have any _silver bullet_ solution to this.

> infinite growing in unsupported cases (like nested struct table alignment).

I don't think this qualifies for _infinite growing_. Even if something like that is possible in theory, it doesn't mean uncrustify has to support it.

In any case, do not hesitate to make a PR with your proposal in a better direction.

PR #2104, opened 2018-11-27
correct the type of token by a catch... when statement

guy-maurel:
} catch (Exception e) when (DateTime.Now.DayOfWeek == DayOfWeek.Saturday)
The type must be CT_SPAREN_CLOSE

### DIFF
--- a/tests/input/cs/exception-filters.cs
+++ b/tests/input/cs/exception-filters.cs
@@ -24,7 +24,7 @@ try {
 }
 try {
   int a = (int)when.foo();
-} catch (Exception e) when (DateTime.Now.DayOfWeek == DayOfWeek.Saturday)
+} catch (Exception e)when (DateTime.Now.DayOfWeek == DayOfWeek.Saturday)
 {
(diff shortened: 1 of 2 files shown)


tests/input/cs/exception-filters.cs
```diff
@@ -24,7 +24,7 @@ void TestExceptionFilter()
 }
 try {
   int a = (int)when.foo();
-} catch (Exception e) when (DateTime.Now.DayOfWeek == DayOfWeek.Saturday)
+} catch (Exception e)when (DateTime.Now.DayOfWeek == DayOfWeek.Saturday)
```
CDanU: The above `CT_SPAREN_CLOSE` change seems correct. 
I don't understand why the space was removed here. Did you forget to push a spacing option for the `ben_036.cfg` file and a corresponding test output?

guy-maurel: The line 1162 at _space.cpp_ was not covered.
The rule is "FORCE".
I changed the test data _exception-filters.cs_ to prove it is now working.

PR #2111, opened 2018-11-28
Provide test data for the option sp_deref

guy-maurel:


tests/input/cs/1822.cs
```diff
@@ -0,0 +1,2 @@
+                foreach (Cat c in *cats)
```
CDanU: It would be nice if you could preformat that with an empty config.

PR #2114, opened 2018-11-28
Remove the code

guy-maurel:
The condition **chunk_is_comment(second)** is never reached because at line 408:
```.cpp
   if (second->parent_type == CT_COMMENT_END)
   {
      switch (second->orig_prev_sp)
      {
      case 0:
         log_rule("orig_prev_sp-REMOVE");
         return(IARF_REMOVE);

      case 1:
         log_rule("orig_prev_sp-FORCE");
         return(IARF_FORCE);

      default:
         log_rule("orig_prev_sp-ADD");
         return(IARF_ADD);
      }
```
is reached before.


CDanU:
Check if this is still the case with something like:
```cpp
int i = 0; /* a b *//* a b */  int b = 0;
```

guy-maurel:
Yes. It is still not used with your example.
The output is the same.
Thanks.

guy-maurel:
Sorry for closing.
Any other comments?

PR #2120, opened 2018-11-29
Add config option: nl_fdef_brace_cond

Benio101:
## Pull request
Add config option: `nl_fdef_brace_cond`.

### Description
`nl_fdef_brace_cond` allows to set _iarf_ for `nl_fdef_brace` if function signature ends with cparen _(skipping new lines and comments)_.

### Rationale
`nl_fdef_brace_cond` allows removing redundant new line between function signature and opening bracket if function signature ends with a closing cparen. This option is essential for dangling cparen (`indent_paren_close = 2`) with `nl_fdef_brace = add|force`, eg for c++ fdef without cv qualifiers; compare `void f(){` vs `void f() const\n{`.

#### Config
```.cfg
nl_fdef_brace                   = force
indent_paren_close              = 2
```
#### Input
```.cpp
unsigned int& function(
	unsigned int& x,
	unsigned int& y
){
	return blah(x, y);
}
```

#### Output without `nl_fdef_brace_cond`
```.cpp
unsigned int& function(
	unsigned int& x,
	unsigned int& y
)
{
	return blah(x, y);
}
```

#### `nl_fdef_brace_cond`
Option `nl_fdef_brace_cond = remove` allows removing redundant space between `)` and `{` for dangling cparen, while still adding it when there is a cv qualifier after `)`, eg `const`.

#### Output with `nl_fdef_brace_cond = remove`
```.cpp
unsigned int& function(
	unsigned int& x,
	unsigned int& y
){
	return blah(x, y);
}
```

### Compatibility
Option is backwards compatible.

### Note
Local tests (`100% tests passed, 0 tests failed out of 14`): [tests.log](https://pastebin.com/raw/geuRYn4v)


CDanU:
As I understood you want to have two options, one for your _normal_ function syntax `f(){}` and another for your in whatever way qualified function like `f() const {}`.

In the code you first get the value for your new function and then if that is not defined you use the value for the existing `nl_fdef_brace` option. 

Generally we use the option value from the most generic option (in this case `nl_fdef_brace`) and then overwrite that with more specific ones. So your new option should become a `nl_fdef_qulified_brace` (or hopefully something better named). After that we have the possibility to add even more specific options like `nl_fdef_ref_qulified_brace` or `nl_fdef_const_qulified_brace`.

Benio101:
Yes, looks good enough.

Benio101:
I'd like to have an option to overwrite `nl_fdef_brace` if function signature ends with _cparen_ (following `parameter-declaration-clause`).
```.cpp
// Function signature ends with cparen.
// Use `nl_fdef_brace_cond` instead of `nl_fdef_brace`.
unsigned int& function(
	unsigned int& x,
	unsigned int& y
){
	return blah(x, y);
}


// Function signature does not end with cparen.
// `cv-qualifier`, `noexcept-specifier` and `attribute-specifier` are present.
// If at least one of them is present, don't use `nl_fdef_brace_cond`, and use `nl_fdef_brace`.
// Note the above list is not full, eg `virt-specifier-seq` xor `requires-clause` can be present.
// ※ http://eel.is/c++draft/dcl.fct.def.general#nt:function-definition
unsigned int& function(
	unsigned int& x,
	unsigned int& y
) const noexcept
	[[excepts audit: y > 0]]
	[[ensures axiom r: r > 1]]
{
	return blah(x, y);
}
```

PR #2121, opened 2018-11-29
D language has the same syntax

guy-maurel:
A extra condition for D is not necessary.
Ref. https://dlang.org/spec/struct.html
Static Initialization of Structs
```.d
struct S { int a, b, c, d = 7; }
S s = { a:1, b:2 };           // s.a = 1, s.b = 2, s.c = 0, s.d = 7
```

### DIFF
--- a/src/align.cpp
+++ b/src/align.cpp
@@ -580,8 +580,7 @@ void align_struct_initializers(void)
    {
       chunk_t *prev = chunk_get_prev_ncnl(pc);
       if (  chunk_is_token(prev, CT_ASSIGN)
-         && (  chunk_is_token(pc, CT_BRACE_OPEN)
-            || (language_is_set(LANG_D) && chunk_is_token(pc, CT_SQUARE_OPEN))))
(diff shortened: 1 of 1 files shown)


CDanU:
While I agree that according to the provided link D structs are for the most part like C++ structs (esp with C++20 designated initialization) I think that there is more to that bracket check. 
It got introduced with 46bc5762b4836cdc8ca6f927c5c3ca2dd27d1d7a, the tests have stuff like:
```cpp
int[ 3 ] a = [ 2:0, 0:3, 1:2 ];
```
I suspect that the (super atrocious) `align_init_brace` function was abused to handle such cases as well, I honestly don't understand what it does.

guy-maurel:
The function cannot work, because the loop is looking for CT_BRACE_OPEN for the next chunk.

I propose to adopt the PR and make a new issue for the D-language #2127

gmaurel:
It was my error to close it.
Any other comments?

Issue #2102, opened 2018-11-26
1ui32 syntax not supported


chris-mich:
```
# Uncrustify-0.68
indent_columns   = 4
indent_with_tabs = 0

```
// Test.cpp (no changes are expected)
`unsigned __int32 b = 1ui32;`

// Output
`unsigned __int32 b = 1ui 32;`

This syntax is probablement ms-specific, but this code will not compile anymore.

guy-maurel:
Please make a test and report.

PR #2133, opened 2018-12-17
Recognize MS suffix

guy-maurel:
Ref. #2102 
A bugfix proposal


guy-maurel:
Well catch.
I preffer  "Issue_#".
And propose to make the changes over the xyz.test files.

Issue #2090, opened 2018-11-24
Vala nullable types should be handled the same as C#


brendanlong:
In Vala, types can end with `?` to indicate nullability the same as C#:

```
class Example {
    int? some_int = null;
}
```

But uncrustify treats all `?` as the ternary operator and formats it like this:

```
class Example {
    int ? some_int = null;
}
```

It seems like this was fixed in C#, see ba5671d892116479bb6c3dc31782119495146c96, #1628, #1290, #533 and probably others. I'm not sure what's necessary to make this work the same for Vala.

(I tried this on both 0.66 and git master with the same results, and I don't see any options that would effect this unless I want to remove spaces around the ternary operator too)

guy-maurel:
#2137 is a bugfix proposal.
Please test and report

guy-maurel:
A new option sp_type_question was introduced with the PR #2148

phako:
I tried running master (69f3834adae4f929ca4d50b2fc3e15453bc852ad) with sp_type_question = remove and it still keeps adding a space between the type and the ? - probably there's some bad interaction with one of the tenary operator configs?
[uncrustify.vala.txt](https://github.com/uncrustify/uncrustify/files/4161998/uncrustify.vala.txt)

phako:
Ah the option is force, not remove. works for me then.

PR #2137, opened 2018-12-27
types can end with ? to indicate nullability

guy-maurel:
Ref. #2090


brendanlong:
I just tried this on FeedReader and it produced the expected results (no extra spaces before `?` in types). It seems like this causes it to ignore spaces before ? in type names. Am I correct in assuming that there's currently not an option to remove them if they exist? (This change alone would be enough to make Uncrustify usable for me so I'm not complaining)

PR #2167, opened 2019-01-17
Update package.json version for semver

shazron:
npm doesn't like versions without major.minor.patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "A highly configurable, easily modifiable source code beautifier for C, C++, C#, ObjectiveC, D, Java, Pawn and VALA",
   "main": "uncrustify",
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uncrustify",
-  "version": "0.68",
+  "version": "0.68.1",
```
mihaipopescu: Why not `.0` then?

shazron: Can't publish to npm without the .patch version number. 
I thought the version was 0.68.1 already? (at least what I see in the Releases tab in GH and also the tags for the repo)

mihaipopescu: My mistake then! You are right, there has been a precedent having a patch version number as well.

shazron:
> Please make some changes to the file **release-steps.txt** to help us keep the right version number.

Done. I added an extra commit for cleaning up the package.json

Issue #2170, opened 2019-01-18
Fix for #1340 breaks align_assign_decl_func (and others?)


rwiesenfarth:
Failing version: Uncrustify-0.68-2-8c80bd84

Running uncrustify with align_assign_span = 1 on this snippet:

```
class Foo
{
  public:
    Foo( int bar = 1 );
    Foo( const Foo & ) = delete;
    Foo &operator= ( const Foo & ) = delete;
    ~Foo();
};
```

results in

```
class Foo
{
  public:
    Foo( int bar                   = 1 );
    Foo( const Foo & )             = delete;
    Foo &operator= ( const Foo & ) = delete;
    ~Foo();
};
```

even if align_assign_decl_func is set to 1.

Removing the comment in align.cpp line 717 (fix for #1340) brings back the intended behavior.

rwiesenfarth:
Above mentioned location corresponds to align_assign.cpp line 56 on master d3feedbc01...

guy-maurel:
Thank for the message.
#2177 is a proposal for a bugfix.
Please test it and report.

Issue #1340, opened 2017-09-08
Alignment issue when passing lambda as function parameter


ttnghia:
Hi,

I have an alignment issue when using lambda. Usually, I have successfully set up uncrustify config and get the alignment like this:
```
double t    = 0; 
double t1   = 0; 
double t111 = 0;
```
When I write a lambda, I also get the desired effect:
```
auto f = [](double x) -> double
         {
             double t    = 0;
             double t1   = 0;
             double t111 = 0;

             return ...;
         };
```

However, when I write a lambda as a parameter passed to a function, I cannot get the assignment '=' of variables t, t1, t111 aligned anymore:
```
std::transform(v1.begin(), v1.end(), v2.begin(),
               [](double x) -> double
               {
                   double t = 0;
                   double t1 = 0;
                   double t111 = 0;

                   return ...;
               }; );
```

For nested lambda, I don't have such problem:
```
auto f = [](double x) -> double
         {
             auto f = [](double x) -> double
                      {
                          auto f = [](double x) -> double
                                   {
                                       double t    = 0;
                                       double t1   = 0;
                                       double t111 = 0;

                                       return ...;
                                   };

                          return ...;
                      };

             return ...;
         };
```

I have tried to change many parameters in the config file including _align_assign_span_ and _align_assign_thresh_, but could not succeed. Did I miss something?

guy-maurel:
Are you using an empty configuration file?
If not, push it just now.

ttnghia:
I have a lot of configs in my config file. Here it is: [MyUncrustify.txt](https://github.com/uncrustify/uncrustify/files/1349678/MyUncrustify.txt)

guy-maurel:
I am working on it.

guy-maurel:
#1375 is a bugfix proposal.
Please test and report.

ttnghia:
Tested. That works! Thanks very much 👍

PR #2177, opened 2019-01-28
The alignment at the sign '=' must be divided in different cases.

guy-maurel:
Case  **CT_ASSIGN**
        The most common case such as:
```.cpp
        a = 1;
        bbb = 13;
```

Case  **CT_ASSIGN_DEFAULT_ARG**
        The default value for an argument such as:
```.cpp
        Foo( int Foo = 5 );
```

Case  **CT_ASSIGN_FUNC_PROTO**
        The function prototype modifier such as:
```.cpp
        void* operator new(std::size_t) = delete;
        Foo( const Foo & ) = default;
        Foo( const Foo & ) = 0;
```

Introduce the new option align_assign_func_proto

Ref. #1340, #2170

### DIFF
--- a/src/align_assign.cpp
+++ b/src/align_assign.cpp
@@ -38,9 +38,13 @@ chunk_t *align_assign(chunk_t *first, size_t span, size_t thresh, size_t *p_nl_c
    vdas.Start(span, thresh);
    vdas.m_right_align = as.m_right_align;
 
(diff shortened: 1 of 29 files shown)


guy-maurel:
Use the options:
```.cfg
align_assign_decl_func = 1
align_assign_span      = 6
```

Issue #2173, opened 2019-01-23
Option align_func_proto_thresh missing


p4pyru5:
The threshold option for aligning function prototypes is not implemented.
Suitable option name would be `align_func_proto_thresh`.

uncletall:
Indeed, I came across the same issue. This is the code that I have in my class definition and it is not what I would expect. could you please assist in implementing this feature?

void                                                                                    f1();
void                                                                                    f2();
void                                                                                    f3();
void                                                                                    f4();
void                                                                                    f5();
void                                                                                    f6();
void                                                                                    f7();
void                                                                                    f8();
SomeLongClassNameThatIsReturnedByAnotherFunction GetIt();

uncletall:
Fixed it but am not sure what tests are required to be changed.

uncletall:
Can this be closed?

PR #2246, opened 2019-04-05
Added the align_func_proto_thresh option

uncletall:
When aligning function prototype that have very long return type
it can result in all prototype function names to be indented very
far. This option limits this.

Ref. #2173

### DIFF
--- a/src/align_func_proto.cpp
+++ b/src/align_func_proto.cpp
@@ -18,9 +18,11 @@ void align_func_proto(size_t span)
 {
    LOG_FUNC_ENTRY();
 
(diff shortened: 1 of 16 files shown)


guy-maurel:
You need to update some tests files.
You can you the script tests/cli/test_cli_options.py to show that.

uncletall:
> I miss one (or more) test(s) example(s) to show how it works

I have added some tests, please let me know if this is sufficient and if the tests were added in the correct place

gmaurel:
Thanks for the PR

Issue #2270, opened 2019-04-29
Request: Add option to handle i18n gettext paarenthesis


Alecaddd:
This is not an issue as it's more of a feature request.

Would it be possible to add an option to remove the space before the parenthesis when dealing with i18n gettext in Vala?

Example:
`new Gtk.Label (_("Label text"));`
turns into
`new Gtk.Label (_ ("Label text"));`

I'd like to keep the space before the parenthesis for pretty much everything, except for translatable strings.

Thank you.

Current version: Uncrustify-0.66.1_f

gmaurel:
Please provide a link to a part of the language description.
What is necessary to recognize the place?
What is the name of "_" in this context?

Philip-Scott:
@gmaurel the `_()` is a vala helper function to make strings translatable using gettext. There are also others such as `N_()` and `C_()`. 

The space is being requested by uncrustify because it IS technically a function, but we use it as if it was a macro or just part of the string. 

See https://elementary.io/docs/code/getting-started#Adding-Translations for more details on how the translatable strings are formatted. 

I think that this is the file where it's defined in the glib vapi
https://github.com/GNOME/vala/blob/f83ecd81d6ede886e1d2573c60da28b993db89e3/vapi/glib-2.0.vapi#L5381 It contains the reference of `_()` along the other functions. 

The rule could probably be that functions that end with `_` do not need the space, so it would cover the other formating options as well

guy-maurel:
PR #2273 is a proposal.
Please test it and report.

albfan:
@guy-maurel just mention that this `_()` is used in python, c, so stick to vala `sp_vala_after_translation` is not accurate.

Seems this is released in 0.69, so I will add to our vala uncrustify.cfg and see how it works

guy-maurel:
Fine. I wait for your next comment.

albfan:
Not sure what you mean. I post exactly that comment 1 month ago (trying to polish PR):

https://github.com/uncrustify/uncrustify/pull/2273#issuecomment-489391497

Now PR is merged I post here as this is still open.

@guy-maurel, Do you think we can deprecate that property and add a general new one?

PR #2273, opened 2019-04-30
Introduce a new option sp_vala_after_translation

guy-maurel:
to control the space after the translation function _
Ref. #2270


albfan:
Hi @guy-maurel, this looks nice, but see _() is a gettext common function: C, python:

https://docs.python.org/3/library/gettext.html#changing-languages-on-the-fly

You can find everywhere:

https://www.gnu.org/software/gettext/manual/html_node/Mark-Keywords.html

PR #2282, opened 2019-05-03
Change arrows to ASCII characters

guy-maurel:
Thanks to @oldunixguy
Ref. #2054


oldunixguy:
uncrustify-68-268-0be1b1e is what just got pulled by me.
I dont know how references like "Ref. #2054" relate to what gets pulled by the git clone... I'm too new to git.

What ever happened to SCCS which I used to use all the time?

Anyway, the non-ASCII chars have been expunged from the default config output by --update-config-with-doc.

thanks
oldunixguy

PR #2283, opened 2019-05-03
Try a better documentation.

guy-maurel:
Ref. #2084

Thanks for the proposals.

### DIFF
--- a/src/options.h
+++ b/src/options.h
@@ -2607,6 +2607,7 @@ align_var_def_span;
 // 0: Part of the type     'void *   foo;' (default)
 // 1: Part of the variable 'void     *foo;'
 // 2: Dangling             'void    *foo;'
+// Dangling: the '*' will not be taken into account when aligning.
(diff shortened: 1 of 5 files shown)


arneforlie:
IMHO these are confusing options, and I'm not sure that the proposed changes make them easier to understand. I had to make a small example and then observe the results for each style. I used the following input:
```
long  *lptr;
int  *iptr;
long  lptr;
int  iptr;
```
With a config file of:
```
align_var_def_span       = 1
align_var_def_star_style = 0
```
I get
```
long *lptr;
int * iptr;
long  lptr;
int   iptr;
```
But although the documentation says:
```
# How to align the '*' in variable definitions.
# 0: Part of the type     'void *   foo;' (default)
```
there is no actual alignment of the '*', it is the **variables** that are aligned.

After having pondered this for a while, I think the sentence "How to align the * in variable definitions" is misleading me. Perhaps if this was rewritten as "How to consider (or treat) the '*' in the alignment of variables" would improve matter.

Issue #2285, opened 2019-05-08
C++ __attribute__ specifiers in class definition cause constructors being tagged as function calls.


patrickrohr:
When `__attribute__` specifiers are used in class definitions, constructors are wrongly tagged as function calls (FUNC_CALL), rather than function declarations (FUNC_CLASS_PROTO) / definitions (FUNC_CLASS_DEF). This causes some formatting issues -- specifically relating to formatting member initializer lists.

A similar issue seems to have previously been fixed for `__declspec` specifiers in #1289.
I have a fix prepared in https://github.com/patrickrohr/uncrustify/commit/657de64184afe6c3c5e1addbb1cc03ae41d51453 that I will be happy to merge into master.

```
# Uncrustify-0.68-263-8454ca48
# option(s) with 'not default' value: 0
#
# -=====-
# Line                Tag              Parent          Columns Br/Lvl/pp     Flag   Nl  Text
#   1>              CLASS[               NONE][  1/  1/  6/  0][0/0/0][     70000][0-0] class
#   1>          ATTRIBUTE[               NONE][  7/  7/ 20/  1][0/0/0][         0][0-0]       __attribute__
#   1>        FPAREN_OPEN[          ATTRIBUTE][ 21/ 21/ 22/  1][0/0/0][ 100000000][0-0]                     (
#   1>         PAREN_OPEN[               NONE][ 22/ 22/ 23/  0][0/1/0][ 100040000][0-0]                      (
#   1>          FUNC_CALL[               NONE][ 23/ 23/ 33/  0][0/2/0][     40000][0-0]                       visibility
#   1>        FPAREN_OPEN[          FUNC_CALL][ 34/ 34/ 35/  1][0/2/0][ 100000000][0-0]                                  (
#   1>             STRING[               NONE][ 35/ 35/ 44/  0][0/3/0][  10040010][0-0]                                   "default"
#   1>       FPAREN_CLOSE[          FUNC_CALL][ 44/ 44/ 45/  0][0/2/0][ 110000010][0-0]                                            )
#   1>        PAREN_CLOSE[               NONE][ 45/ 45/ 46/  0][0/1/0][ 110000000][0-0]                                             )
#   1>       FPAREN_CLOSE[          ATTRIBUTE][ 46/ 46/ 47/  0][0/0/0][ 110000000][0-0]                                              )
#   1>               TYPE[               NONE][ 48/ 48/ 55/  1][0/0/0][  10060000][0-0]                                                TestBad
[...]

PR #2286, opened 2019-05-08
Change mark_class_ctor to Skip Attributes Specifiers

patrickrohr:
This is the fix to the issue described in #2285. It looks like this was previously fixed for `__declspec` specifiers, but `__attribute__` specifiers still cause `mark_class_ctor` to return prematurely.


patrickrohr:
Let me work on adding a test case to this too.

PR #2308, opened 2019-06-04
Fix cast-align warnings in md5.cpp on armhf 

emersonknapp:
When building this package using GCC for `armhf` platform, the `-Wcast-align` warning is raised on the `(UINT32 *)` casts from `UINT8 *`, because the compiler can't guarantee the access is word-aligned. 

By having the default view of this `m_in` variable be a `UINT32`, and casting the other way to `UINT8` where needed, we get some assurance that `UINT32` accesses are aligned.

### DIFF
--- a/src/md5.cpp
+++ b/src/md5.cpp
@@ -72,6 +72,7 @@ MD5::MD5()
     * The MD5 stuff is written for little endian.
     */
 
+   m_in8           = (UINT8 *)m_in32;
    m_need_byteswap = *(UINT8 *)m_buf != 4;
    m_big_endian    = *(UINT8 *)m_buf == 1;
 }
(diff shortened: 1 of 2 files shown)


gmaurel:
Please, have a look at
https://ci.appveyor.com/project/gmaurel/uncrustify/builds/25019170/job/iw5xvulcv7g6l8g9

emersonknapp:
@gmaurel thanks for pointing me to the CI build. It looks like the VS2017 Release build timed out, but the rest of the configurations passed, in approximately the same amount of time as past builds. I wonder if it was an infrastructure problem - to get more information either way, how would we re-run the build?

emersonknapp:
I pushed the same commit with a new hash to re-trigger the build and it looks fine. What do you think?

PR #2349, opened 2019-07-24
 Add sp_before_square_asm_block (sp_before_square for asm macros)

pietvr:
I added two commits.
- 1 for the code
- 1 for an additional test for the new option.

This is my first commit to github (and uncrustify of course). If something is wrong I will correct ...


pietvr:
I used number 30060 for the c test. This seems wrong. Which number should I use? 10006 ?

gmaurel:
Complete your account https://github.com/pietvr, with address and some more...

src/space.cpp
```diff
@@ -1073,6 +1073,10 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
       {
          return(IARF_FORCE);
       }
+      if (chunk_is_token(first, CT_ASM_COLON) && options::sp_before_square_ignore_asm_block())
```
mihaipopescu: there's nothing saying here that we are in a macro

src/options.h
```diff
@@ -416,6 +416,11 @@ sp_after_semi_for_empty;
 extern Option<iarf_e>
 sp_before_square;
 
+// Whether to ignore sp_before_square for 'asm' blocks (i.e. assume they
+// have their own indentation).
+extern Option<bool>
```
mihaipopescu: why not `Option<iarf_e>` ?

src/space.cpp
pietvr: CT_ASM_COLON in first means we are in a C ASM macro. Otherwise another token would have been found. Should I add a comment?

src/options.h
pietvr: I saw another option implement something alike in indent_ignore_asm_block implemented as a bool.
Using iarf_e is also possible. And is more like the default behavior of course.

Just tell me what the uncrustify maintainers like most!

guy-maurel: I like iarf

guy-maurel: Please use the same comment as the description from options.h

gmaurel: Please change the source as proposed at my comment just before.
It is important to have the same text at all the places

pietvr: I think I have the comment "Add or remove space before '[' for asm block." everywhere.
(space.cpp, options.h, mini_nd_ucwd.txt, mini_d_ucwd.txt, universalindent.cfg, show_config.txt and defaults.cfg).

src/space.cpp
```diff
@@ -1073,6 +1073,12 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
       {
          return(IARF_FORCE);
       }
+      if (chunk_is_token(first, CT_ASM_COLON))
```
mihaipopescu: I don't understand how the assembly colom is related with the open bracket? I don't have access to code so I can't check it myself.

pietvr:
@gmaurel What changes should I still do? I see none (except merging main again ...)

gmaurel:
Can you answer the comment from @mihaipopescu 20 days ago?

src/space.cpp
pietvr: I just solved my main issue with asm here.
When the '[' follows the ':' in asm. I exclude rule sp_before_square.
Because the ': [' sequence is typically in asm (with the space).

Maybe this should be made more generic. (Check we are in asm in a more generic fashion ...)

PR #2361, opened 2019-08-03
Add newline handling for multiline if/while/... conditions

### LINKED ISSUE #2360: No option to add newline before multiline condition
I would like to have uncrustify add a newline after '(' in an if/while/for condition if '(' and ')' are on different lines. Basically the same behavior as `nl_func_call_start_multi_line = true`, but for conditionals. I would expect such an option to be called something like `nl_cond_start_multi_line`, but I don't think one exists.

This is to meet the ROS2 style guide:

https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#open-versus-cuddled-braces

> * When a function call cannot fit on one line, wrap at the open parenthesis (not in between arguments) and start them on the next line with a 2-space indent. Continue with the 2-space indent on subsequent lines for more arguments. (Note that the Google style guide is internally contradictory on this point.)
> 
>     * Same goes for `if` (and `while`, etc.) conditions that are too long to fit on one line.
> 

Simple example:
```
if (this &&
    that ||
    both) {
    // ...
}
```
should uncrustify to:
```
if (
    this &&
    that ||
    both) {
    // ...
}
```

current version: Uncrustify-0.69.0_f

rotu:
fix #2360


tests/c.test
```diff
@@ -377,6 +377,8 @@
 09615  i1564.cfg                            c/i1564.c
 
 09616  enum_comma_ifdef.cfg                 c/enum_comma_ifdef.c
+09617  09617.cfg                            c/issue_2360.c
+09618  09618.cfg                            c/issue_2360.c
```
gmaurel: Please use the current way for the file name:
Change at tests/c.text:
09617.cfg to Issue_2360-a.cfg
09618.cfg to Issue_2360-b.cfg
c/issue_2360.c to c/Issue_2360.c

src/newlines.cpp
```diff
@@ -782,6 +782,47 @@ void newline_del_between(chunk_t *start, chunk_t *end)
 } // newline_del_between
 
 
+void newlines_sparens()
+{
+   LOG_FUNC_ENTRY();
+
```
gmaurel: Make a comment which link to meet the ROS2 style guide, as written at your #2360

src/newlines.cpp
```diff
@@ -782,6 +782,47 @@ void newline_del_between(chunk_t *start, chunk_t *end)
...
+        open_paren != nullptr;
+        open_paren = chunk_get_next_type(open_paren, CT_SPAREN_OPEN, ANY_LEVEL))
+   {
+      chunk_t *close_paren = chunk_get_next_type(open_paren, CT_SPAREN_CLOSE, open_paren->level);
+      if (!close_paren)
+      {
+         continue;
+      }
+
+      if (
```
gmaurel: Comment with the text from options.h

src/newlines.h
```diff
@@ -19,6 +19,9 @@
  */
 void newlines_remove_newlines(void);
```
gmaurel: Make a comment to the ROS2 style guide

src/newlines.cpp
rotu: Done

gmaurel:
Could you provide a configuration file for ROS2 ?
It would be fine

rotu:
The configuration file can be found here:

https://github.com/ament/ament_lint/blob/master/ament_uncrustify/ament_uncrustify/configuration/ament_code_style.cfg

I'm working on a few changes since it doesn't totally agree with our developer guide:

https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#language-versions-and-code-format

I'll be happy to add it here when it actually agrees with the written developer guide :-). https://github.com/uncrustify/uncrustify/tree/master/etc

PR #2390, opened 2019-09-07
Fix indentation of statement which goes after '<<'

andrey-starodubtsev:
If statement continues on the next line after `<<` and starts with `(`, it's indented incorrectly, i.e.

```c++
void foo() {
    cout <<
        (a + 1);
}
```

turns into

```c++
void foo() {
    cout <<
    (a + 1);
}
```


gmaurel:
@andrey-starodubtsev 
Please, have a look to the conflict

Issue #2383, opened 2019-08-30
Newline added in //-style comment block


PerMalmberg:
Uncrustify_d-0.69.0-116-1bc07039

Given the following file:
```
// Smooth - C++ framework for writing applications based on Espressif's ESP-IDF.
// Copyright (C) 2017 Per Malmberg (https://github.com/PerMalmberg)
//
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.
//
// This program is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.
//
// You should have received a copy of the GNU General Public License
// along with this program. If not, see <http://www.gnu.org/licenses/>.

#include <smooth/core/io/i2c/I2CCommandLink.h>
#include <smooth/core/io/i2c/I2CMasterDevice.h>

namespace smooth::core::io::i2c
{
    I2CCommandLink::I2CCommandLink(const I2CMasterDevice& master_device)
            : master_device(master_device)
    {
        // Acquire the guard to ensure noone else is trying to use the I2C bus.
        master_device.get_guard().lock();

        cmd_link = i2c_cmd_link_create();
    }

    I2CCommandLink::~I2CCommandLink()
    {
        i2c_cmd_link_delete(cmd_link);

        // Release the guard
        master_device.get_guard().unlock();
    }

    void I2CCommandLink::reset()
    {
        i2c_cmd_link_delete(cmd_link);
        cmd_link = i2c_cmd_link_create();
    }
}
```

When running `/path/to/uncrustify -c uncrustify.cfg --no-backup --replace I2CCommandLink.cpp` a new line is inserted in the comment, like so:

```
// Smooth - C++ framework for writing applications based on Espressif's ESP-IDF.

// Copyright (C) 2017 Per Malmberg (https://github.com/PerMalmberg)
//
```

I've attached .p and .A files as a zip.

Is there a setting that affects this, or is it a bug?

[uncrustify-debug.zip](https://github.com/uncrustify/uncrustify/files/3560953/uncrustify-debug.zip)

guy-maurel:
PR #2398 is a bugfix proposal.
Please test it and report.

PR #2398, opened 2019-09-11
Newline before comment block is now correct

guy-maurel:
Correct the parentheses.
Ref. #2383


PerMalmberg:
@guy-maurel Seems to work nicely, thank you!

PR #2412, opened 2019-09-21
Fix #2411 - preserve leading tabs after brace level in multiline C-style comments when comment formatting is disabled

### LINKED ISSUE #2411: Leading tabs after brace level aren't preserved in multiline C-style comments when comment formatting is disabled
On uncrustify 0.69.

When using tabs for indenting and disabling all comment formatting as such:
```
input_tab_size = 4
output_tab_size = 4
indent_with_tabs = 1 or 2
cmt_width = 0
cmt_reflow_mode = 0
cmt_convert_tab_to_spaces = false
cmt_indent_multi = false
(every other cmt_* is false or 0 or "")
```

In a C-style comment, leading tabs after the brace level aren't preserved and are replaced with spaces.

For this example input:
```c
			/*
			if(cond)
				more commented out code
			*/
```

The current output as of uncrustify 0.69 is:
```c
			/*
			if(cond)
			    more commented out code
			*/
````

Notice how the tab before "more commented out code" is replaced with 4 spaces.

The expected output is the same as the input.

### PR DESCRIPTION
**Pull request by lbmaian**
Fix #2411

In multiline c-comment, when cmt_indent_multi and cmt_convert_tab_to_spaces are false, don't change leading whitespace after brace level.

This only addresses that particular use case. I'm not going to attempt to change the case where cmt_indent_multi=true, since this is already way out of my expertise.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -2101,7 +2101,8 @@ static void output_comment_multi_simple(chunk_t *pc)
       cmt_idx++;
 
       // 1: step through leading tabs and spaces to find the start column
-      if (line.size() == 0 && (line_column < cmt.base_col || options::cmt_convert_tab_to_spaces()))
+      if (  line.size() == 0
+         && (line_column < cmt.base_col || options::cmt_convert_tab_to_spaces()))
       {
          if (ch == ' ')
          {
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -2101,7 +2101,7 @@ static void output_comment_multi_simple(chunk_t *pc)
       cmt_idx++;
 
       // 1: step through leading tabs and spaces to find the start column
-      if (line.size() == 0)
+      if (line.size() == 0 && (line_column < cmt.base_col || options::cmt_convert_tab_to_spaces()))
       {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by andrey-starodubtsev** [comment]
Could you add the test which demonstrates the issue?

**Comment by lbmaian** [comment]
I added a test by the way.

File: src/output.cpp [hunk 1]
```diff
@@ -2101,7 +2101,7 @@ static void output_comment_multi_simple(chunk_t *pc)
       cmt_idx++;
 
       // 1: step through leading tabs and spaces to find the start column
-      if (line.size() == 0)
+      if (line.size() == 0 && (line_column < cmt.base_col || options::cmt_convert_tab_to_spaces()))
```
gmaurel: Please break the `if `after the zero to get two lines

lbmaian: Done, replicated formatting of other such conditions in the file.

lbmaian:
Is there anything else that needs to be done for this to be merged in?

lbmaian:
I already did that though. See my reply in the code review and the additional commit.

PR #2418, opened 2019-09-27
Revert " Add sp_before_square_asm_block (sp_before_square for asm macros)"

gmaurel:
Reverts uncrustify/uncrustify#2349
I need some more tests


gmaurel:
no more necessary.

Issue #2460, opened 2019-10-20
sp_trailing_ret_t = add does not work for noexcept


arneforlie:
Option sp_trailing_ret_t = add fails when noexcept is used on a method.
As we see from the test case, the spaces around the trailing return operator are deleted. The spaces were there before, so they are not ignored, they are actually removed.

[debug.txt](https://github.com/uncrustify/uncrustify/files/3747807/debug.txt)

Test case
```
class X34
{
   public:
      auto f1() const -> bool;
      auto f2() noexcept -> bool;
      auto f3() const noexcept -> bool;
};
```
Expected output
```
class X34
{
public:
auto f1() const -> bool;
auto f2() noexcept -> bool;
auto f3() const noexcept -> bool;
};
```
Generated output
```
class X34
{
public:
auto f1() const -> bool;
auto f2() noexcept->bool;
auto f3() const noexcept->bool;
};
```

guy-maurel:
PR #2463 is a bugfix proposal.
Please test it and report.

arneforlie:
Thanks, it is better but doesn't quite cover all cases noexcept (and throw()) can be used. The following is a more complete test case:
```
class X34
{
   public:
      auto f11() -> bool;
      auto f02() noexcept -> bool;
      auto f03() noexcept(true) -> bool;
      auto f04() noexcept(false) -> bool;
      auto f05() noexcept -> bool = delete;
      auto f06() noexcept(true) -> bool = delete;
      auto f07() noexcept(false) -> bool = delete;

      auto f11() const -> bool;
      auto f12() const noexcept -> bool;
      auto f13() const noexcept(true) -> bool;
      auto f14() const noexcept(false) -> bool;
      auto f15() const noexcept -> bool = delete;
      auto f16() const noexcept(true) -> bool = delete;
      auto f17() const noexcept(false) -> bool = delete;

      auto f21() throw() -> bool;
      auto f22() throw() -> bool = delete;
      auto f23() const throw() -> bool;
      auto f24() const throw() -> bool = delete;
};
```

f03, f04, f06, f07, f13, f14, f16, f17, f21, f22, f23 and f24 still loose the space around ->

guy-maurel:
Thanks for the test data.
A new version is available. Please test again.

arneforlie:
Thanks, it now passes all of above test cases.

PR #2463, opened 2019-10-21
The option sp_trailing_ret_t now works with noexcept

guy-maurel:
Ref. #2460


arneforlie:
There are still some cases that are not handled.
#2460 has been updated with a more complete test case.

gmaurel:
Could you please fill up you profile https://github.com/arneforlie ?
Thanks

PR #2471, opened 2019-10-24
nl for if (after/before), 2. run

guy-maurel:
PR #2465 will be removed


guy-maurel:
Don't merge anything else before THIS is OK.

PR #2535, opened 2019-11-19
Teach indent the `indent_macro_brace` option

jmesmon:
Allows controlling the indent effect of an initial `{` in a macro.
Setting `indent_macro_brace` false removes 1 level (the brace level)
indent when the brace is opened on the same line as the macro.

true:

	#define X() do { \
			bar(); \
			baz(); \
	} while (0)

false:

	#define X() do { \
		bar(); \
		baz(); \
	} while (0)

Default is set to `true` to match current behavior.

### DIFF
--- a/forUncrustifySources.cfg
+++ b/forUncrustifySources.cfg
@@ -37,6 +37,7 @@ indent_class                              = true
 indent_columns                            = 3
 indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
+indent_macro_brace                        = true
 indent_member                             = 3
 indent_with_tabs                          = 0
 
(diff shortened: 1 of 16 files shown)


guy-maurel:
Choose a adequat value for the new option to be add in the file ./forUncrustifySources.cfg

jmesmon:
Ok. I've set `indent_macro_brace = true` in `forUncrustifySources.cfg`. This avoids reformatting the macros in `chunk_list.h`, and matches the default value for the option. Is this a reasonable value to choose for the uncrustify sources?

forUncrustifySources.cfg
```diff
@@ -39,6 +39,7 @@ indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
 indent_member                             = 3
 indent_with_tabs                          = 0
+indent_macro_brace                        = true
```
gmaurel: Please, sort the option before _indent_member_

PR #2538, opened 2019-11-19
Add option `indent_compound_literal_return`

jmesmon:
Setting this to `false` (default is `true`) makes indent treat returned
compound literals like c++ init lists.

Previously, we'd apply an indent from both the `return` and the brace
when returning a compound literal.

Now, we can treat them like just c++ initializers lists (and avoid the
extra indent level).

indent_compound_literal_return = true (current default):

	struct f z(void) {
		return (struct f){
				.z = 1,
		};
	}

indent_compound_literal_return = false:

	struct f z(void) {
		return (struct f){
			.z = 1,
		};
	}

For uncrustify, we've set this to `false` to match the behavior with c++
initializer lists.

### DIFF
--- a/forUncrustifySources.cfg
+++ b/forUncrustifySources.cfg
@@ -35,6 +35,7 @@ align_var_struct_span                     = 3
 indent_bool_paren                         = true
 indent_class                              = true
 indent_columns                            = 3
+indent_compound_literal_return            = false
 indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
 indent_macro_brace                        = true
--- a/etc/defaults.cfg
+++ b/etc/defaults.cfg
@@ -1284,6 +1284,13 @@ indent_token_after_brace        = true     # true/false
 # Whether to indent the body of a C++11 lambda.
 indent_cpp_lambda_body          = false    # true/false
 
+# How to indent compound literals that are being returned.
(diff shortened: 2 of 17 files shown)


jmesmon:
Not sure if we'd want an option to change this behavior as it does change the defaults. Given that this is really building on the existing support for returned c++ initializer lists (where returning compound literals is a strong C analog), I did not add an option initially.

gmaurel:
Yes, a new option would be better.

jmesmon:
I've updated this PR to add a new option `indent_compound_literal_return` to control this option. It defaults to `true` (ie: existing behavior adds an indent for compound literals in addition to the return indent), and I've made it `false` for uncrustify` sources (resulted in no formatting changes, and matches the behavior of c++ initializer lists).

forUncrustifySources.cfg
```diff
@@ -39,6 +39,7 @@ indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
 indent_member                             = 3
 indent_with_tabs                          = 0
+indent_compound_literal_return            = false
```
gmaurel: Please, sort the option before _indent_cpp_lambda_only_once_

gmaurel:
@jmesmon could you resolve the conflict

jmesmon:
I've rebased to resolve the conflict.

PR #2539, opened 2019-11-19
Add indent_sparen_extra to adjust indent in ctrl statements

jmesmon:
When using `indent_continue` set to the same value as `indent_columns`
(I'll use `indent_continue = 8` in examples here), allow adjusting the
indent added due to a SPAREN_OPEN.

In otherwords, when we have a control statement, allow adding (or
removing) indent to continuations.

Existing:

	// indent_continue = 8 // non-default
	// indent_sparen_extra = 0 // default

	void f(void) {
	if (long__________________________x
		&& longger______x) {
		h();
	}

With this change:

	// indent_continue = 8 // non-default
	// indent_sparen_extra = 8 // non-default

	void f(void) {
	if (long__________________________x
			&& longger______x) {
		h();
	}

### DIFF
--- a/forUncrustifySources.cfg
+++ b/forUncrustifySources.cfg
@@ -40,6 +40,7 @@ indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
 indent_macro_brace                        = true
 indent_member                             = 3
+indent_sparen_extra                       = 0
 indent_with_tabs                          = 0
 
 # Newline adding and removing options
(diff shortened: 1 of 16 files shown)


gmaurel:
Same as #2535 Choose a adequat value for the new option to be add in the file ./forUncrustifySources.cfg

jmesmon:
I've updated to add the option to `forUncrustifySources.cfg`. I've set it to zero here, which matches the default and is likely a good choice given that we're not setting `indent_continue`.

forUncrustifySources.cfg
```diff
@@ -39,6 +39,7 @@ indent_cpp_lambda_only_once               = true
 indent_first_bool_expr                    = true
 indent_member                             = 3
 indent_with_tabs                          = 0
+indent_sparen_extra                       = 0
```
gmaurel: Please, sort the option before _indent_with_tabs_

gmaurel:
@jmesmon : please, have a look about the conflict.

PR #2602, opened 2020-01-10
Add a setter

guy-maurel:
Will be used by #2281


src/output.cpp
```diff
@@ -449,22 +449,23 @@ void output_parsed(FILE *pfile)
    // and              430: ... fprintf(pfile, "%s# %3zu>%19.19s[%19.19s] ...
    // here                                                xx xx   xx xx
 #ifdef WIN32
-   fprintf(pfile, "# Line                Tag              Parent          Columns Br/Lvl/pp     Nl  Text");
+   fprintf(pfile, "# Line                Tag         Parent_type  Type of the paren          Columns Br/Lvl/pp     Nl  Text");
```
chrisemb: I guess it should be
Parent_type  Type of the parent
(with 't' in the end)

Issue #2616, opened 2020-01-27
C: nl_after_struct affects prototype definitions including struct parameters


jaguarondi:
The setting `nl_after_struct = 2` will affect prototype definitions with struct in the parameter list.

This:
```
void ethernetif_input(struct netif netif);
void ethernetif_update_config(struct netif netif);
```
becomes this:
```
void ethernetif_input(struct netif netif);

void ethernetif_update_config(struct netif netif);
```

- Uncrustify-0.70.1_f
- https://pastebin.com/hTSj0Ubx

guy-maurel:
PR #2619 is a bugfix proposal.
Please test it and report.

jaguarondi:
Tested and the issue is fixed  in PR #2619.

PR #2619, opened 2020-01-31
The setting nl_after_struct = 2 may NOT affect prototype definitions …

guy-maurel:
…with struct in the parameter list.

Ref. #2616


guy-maurel:
will b later splited

PR #2624, opened 2020-02-03
Fix issue 2623 with nested template syntax

samishchandra:
The template token parsing logic has a bug when `static_cast` is present. `static_cast` already takes care of the angle_close token, hence during the check for template we need to take one less angle_close into account.

`tok_split_gte` option is added as a hack to parse nested template syntax like `Class1<Class2<Class3>>>`, but it breaks the simple syntax `assert(x<0 && y>=3)`. This commit correctly fixes the template parsing logic, hence the option `tok_split_gte` option may become obsolete and can be removed.

### DIFF
--- a/src/tokenize_cleanup.cpp
+++ b/src/tokenize_cleanup.cpp
@@ -58,7 +58,7 @@ static void check_template_args(chunk_t *start, chunk_t *end);
  *
  * @param start  chunk to start check at
  */
-static void check_template(chunk_t *start);
+static void check_template(chunk_t *start, bool in_type_cast);
 
 
 /**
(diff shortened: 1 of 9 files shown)


gmaurel:
Please have a look to the source test to make some little changes to the source code.

samishchandra:
@gmaurel sorry for my noob questions. I did not get what you mean. Do you mean to look at TESTING documentation ?

guy-maurel:
Have a look to https://github.com/uncrustify/uncrustify/pull/2624
At the end, you can see:  continuous-integration/travis-ci/pr — The Travis CI build could not complete due to an error
Follow the link Details Most of the links are red.
Follow the first one: https://travis-ci.org/uncrustify/uncrustify/jobs/645338531?utm_medium=notification&utm_source=github_status
Scroll down to the line 463 and see the problem.
This is the output of the source test. You can repeat the test with
ctest --output-on-failure make -I 1273,1273

samishchandra:
Thanks for the info. Updated the code.

gmaurel:
Please make a "git rebase master" on your branch

PR #2632, opened 2020-02-04
Fix wrongly tagged parent type for braces inside OC_CLASS

samishchandra:
Open and Close braces are wrongly tagged with parent type `OC_CLASS` incase of parent type `ASSIGN`. This should fix the issue 2631.


gmaurel:
Please the source test:
`/uncrustify/tests/run_sources_tests.py`
to see ... (again)
or use the commands:
```.txt
cd build
ctest --output-on-failure make
```

PR #2635, opened 2020-02-05
Fix ptr type in trailing return syntax to correctly apply rules

samishchandra:
The `STAR` is not correctly tagged as `PTR_TYPE` in trailing return syntax. This causes several rules in config related to ptr not apply.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -1788,6 +1788,16 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
             //    using AbstractLinkPtrPtr = AbstractLink**;
             set_chunk_type(pc, CT_PTR_TYPE);
          }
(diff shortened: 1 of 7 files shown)


src/combine.cpp
```diff
@@ -1788,6 +1788,16 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
             //    using AbstractLinkPtrPtr = AbstractLink**;
             set_chunk_type(pc, CT_PTR_TYPE);
          }
+         else if (  (  get_chunk_parent_type(pc) == CT_FUNC_DEF
+                    && (chunk_is_opening_brace(next) || chunk_is_star(pc->next)))
```
gmaurel: Please write the issue number as comment at the begin of your work, line 1791.
Look at some examples, such at line 522:
// Issue # 2166

PR #2644, opened 2020-02-06
Fix OC_BLOCK_EXPR parent tag not recognized for OC blocks inside C method calls

samishchandra:
When OC block is used as a param inside C function calls, they are not recognized as OC_BLOCK_EXPR. This can cause the rules around OC block fail. This fixes the issue 2643.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -6481,6 +6481,12 @@ static void handle_oc_block_type(chunk_t *pc)
          return(handle_oc_block_literal(pc));
       }
 
+      // Check apo is '(' or else this might be a block literal. Issue 2643.
+      if (!chunk_is_paren_open(apo))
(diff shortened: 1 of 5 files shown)


src/combine.cpp
```diff
@@ -6469,6 +6469,12 @@ static void handle_oc_block_type(chunk_t *pc)
       chunk_t *apo = chunk_get_next_ncnl(tpc);   // arg open paren
       chunk_t *apc = chunk_skip_to_match(apo);   // arg close paren
 
+      // Check apo is '(' or else this might be a block literal.
```
gmaurel: Please add a comment with the issue number

samishchandra: Thanks for the review. Added.

PR #2645, opened 2020-02-06
Fix issues with OC block indentation

samishchandra:
There are multiple issues with OC block indentations, the added test cases should explain the different code syntaxes that are supported with the fixes.


src/indent.cpp
```diff
@@ -454,9 +450,32 @@ static chunk_t *oc_msg_block_indent(chunk_t *pc, bool from_brace,
       tmp = chunk_get_prev_nc(chunk_skip_to_match_rev(tmp));
    }
 
+   if (chunk_is_token(tmp, CT_PTR_TYPE))
```
gmaurel: Please add a comment with the issue number,
here and some more places

PR #2646, opened 2020-02-07
Add option indent_off_after_assign to turn off indent similar to indent_off_after_return_new

samishchandra:
The added option will help turn off indentation after assign expression. This is how Xcode does formatting by default. Hence this option would provide a great value in staying close to regular formatting rules which most people use.


gmaurel:
Please, introduce the new option in forUncrustifySources.cfg, as explained before,
some more comments with the issue number.
I restart the test which failed

PR #2647, opened 2020-02-07
Add option indent_off_after_return to turn off indent similar to indent_off_after_return_new

samishchandra:
The added option will help turn off indentation after 'return' expression. This is how Xcode does formatting by default. Hence this option would provide a great value in staying close to regular formatting rules which most people use.


gmaurel:
Same things as PR #2646

PR #2651, opened 2020-02-10
Fix OC tests after the merge #2645

samishchandra:


samishchandra:
@gmaurel, thanks for the patience on this. :-)

PR #2659, opened 2020-02-12
Add option indent_oc_inside_msg_sel to indent code inside OC message selectors

samishchandra:
The indentation inside OC messages is not correctly handled currently. We need to correctly indent the statements inside OC message or else option `align_oc_msg_colon_span` will make the indentation of code inside OC message selectors look off. This is because there is no parse frame maintained for OC messages. Added optional configuration option to enable indentation inside OC message selectors. Issue 2658.

### DIFF
--- a/forUncrustifySources.cfg
+++ b/forUncrustifySources.cfg
@@ -42,6 +42,7 @@ indent_macro_brace                        = true
 indent_member                             = 3
 indent_sparen_extra                       = 0
 indent_with_tabs                          = 0
+indent_oc_inside_msg_sel                  = false
 indent_off_after_assign                   = false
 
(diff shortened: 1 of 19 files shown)


forUncrustifySources.cfg
```diff
@@ -42,6 +42,7 @@ indent_macro_brace                        = true
 indent_member                             = 3
 indent_sparen_extra                       = 0
 indent_with_tabs                          = 0
+indent_oc_inside_msg_sel                  = false
```
gmaurel: as this option will be never used for uncrustify itself, please put the line as comment with the other ones.
Please, make the change(s) at the next PR

Issue #2386, opened 2019-09-05
Enhancement: keep form feed characters; ASCII 12


oneptr:
### Enhancement request:  

**keep Form feed characters**

It is not that uncommon to use a formfeed character (FF, ASCII 12) to divide a source file into sections; e.g. one section per function. An emacs extention displays a single line consisting of a FF (ASCII-12) followed by a \n (newline) as a colored horizontal line. This make the source file much better readable for me. (Just FYI; a C/C++ compiler is fine with these characters)

Currently those FF characters are removed from the source file; possibly because there's an option set to remove white-space before a \n. But I suspect that those characters are removed unconditionally.

In any case, I would like to remove white space before \n with the exception of "FF \n". Of course this behavior needs to be on/off switchable.

guy-maurel:
Could you push such a source file?

oneptr:
attached a very simple file that contains some of those lines. btw. the emacs extention is called page-break-lines
[hw.cpp.txt](https://github.com/uncrustify/uncrustify/files/3935482/hw.cpp.txt)

see line 0020 0050 00b0: sequence 0c 0a

```> xxd -g 1 hw.cpp.txt    
00000000: 23 69 6e 63 6c 75 64 65 20 3c 69 6f 73 74 72 65  #include <iostre
00000010: 61 6d 3e 0a 23 69 6e 63 6c 75 64 65 20 3c 73 74  am>.#include <st
00000020: 72 69 6e 67 3e 0a 0a 0c 0a 73 74 64 3a 3a 73 74  ring>....std::st
00000030: 69 6e 67 20 67 65 74 54 65 78 74 28 29 0a 7b 0a  ing getText().{.
00000040: 20 20 20 72 65 74 75 72 6e 20 22 48 65 6c 6c 6f     return "Hello
00000050: 20 57 6f 72 6c 64 22 3b 0a 7d 0a 0c 0a 69 6e 74   World";.}...int
00000060: 20 6d 61 69 6e 28 69 6e 74 20 61 72 67 63 2c 20   main(int argc, 
00000070: 63 68 61 72 20 2a 61 72 67 76 5b 5d 29 0a 7b 0a  char *argv[]).{.
00000080: 20 20 20 73 74 64 3a 3a 63 6f 75 74 20 3c 3c 20     std::cout << 
00000090: 67 65 74 54 65 78 74 28 29 20 3c 3c 20 73 74 64  getText() << std
[...]

PR #2674, opened 2020-02-21
Remove FF from the list

guy-maurel:
Ref. #2386 
It is now possible to remove the Form Feed character from the list of whitespace characters.
  https://en.cppreference.com/w/cpp/string/byte/isspace
The character goes through and can be use to separate parts of the source, as wished.


tests/expected/cpp/30721-Issue_2386.cpp
```diff
@@ -0,0 +1,19 @@
+// A
+
+
+std::sting getText()
```
oneptr: std::string, not std::sting, not sure if it matters here or not !?

gmaurel: It doesn't matter.

Issue #2693, opened 2020-03-17
C/C++: nl_before_return and switch/case behavior


geimer:
If `nl_before_return` is set to `true`, an extra blank line is inserted between a `case` label of a `switch` statement and an immediately following `return`.  This seems somehow related to #1257, though the fix for this issue doesn't seem to work properly (at least not for C/C++).

Uncrustify version: Uncrustify_d-0.70.1-253-86143624
Debug log: [debug.txt](https://gist.github.com/geimer/3310cd29ddbefefb67aaf37e19a841ee)

Input:
```C++
int foo(int arg)
{
    switch (arg)
    {
        case 1:
            return 42;
        default:
            break;
    }
    return 0;
}
```

Actual output:
```C++
int foo(int arg)
{
	switch (arg)
	{
	case 1:

		return 42;
	default:
		break;
	}

	return 0;
}
```

Expected output:
```C++
int foo(int arg)
{
	switch (arg)
	{
	case 1:
		return 42;
	default:
		break;
	}

	return 0;
}
```

PR #2694, opened 2020-03-17
Avoid blank line between case and return statement

geimer:
This PR proposes a fix for issue #2693. I'm not familiar enough with the `uncrustify` code base to judge whether this could have any adverse side effects, though the test suite still passes. However, this means that this particular case isn't covered by the test suite yet. Please let me know whether I should include my test case from the issue as a new test case, or whether I should modify one of the existing test cases.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -2515,7 +2515,7 @@ static void newline_before_return(chunk_t *start)
    if (  pc == nullptr
       || (  chunk_is_token(pc, CT_BRACE_OPEN)
          || chunk_is_token(pc, CT_VBRACE_OPEN)
-         || get_chunk_parent_type(start) == CT_CASE)) // Issue #1257
+         || chunk_is_token(pc, CT_CASE_COLON)))
    {
       return;
    }
--- a/tests/c.test
+++ b/tests/c.test
@@ -124,6 +124,8 @@
 00202  case-2.cfg                           c/case.c
 00203  case-3.cfg                           c/case.c
 00204  bug_1718.cfg                         c/bug_1718.c
+00205  nl_before_return_false.cfg           c/case-nl_before_return.c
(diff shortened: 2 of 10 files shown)


geimer:
It turns out that this is exactly what was proposed already in https://github.com/Unity-Technologies/uncrustify/commit/3b41d6ef7c32114ba9605ffd8b141830dca8212d (though using the helper function rather than using `pc->type` directly), but somehow a different solution got committed in https://github.com/uncrustify/uncrustify/commit/db6adb6c69b238746c60233a423eac3df02c4073. However, the test that was added at the same time does not use `nl_before_return`, and therefore the fact that the issue still persists went by unnoticed...

MattBussing:
What is the ETA on this being merged?

MattBussing:
Just a heads up this doesn't fix default in C#.
It edited a file and left this:

```
default:
      log....

      return;
```

So, this still needs work.

geimer:
@MattBussing Could you please elaborate what you are expecting? I'm not entirely sure I understand your comment. With the following input
```
default:
    return;
```
and `nl_before_return = true`, current `master` inserts an extra blank line:
```
default:

    return;
```
Now, with the proposed change, this blank line is gone:
```
default:
    return;
```
If there is anything in between the `default` and the `return`, the `nl_before_return` still applies, that is,
```
default:
    Console.WrieLine("hello world\n");
    return;
```
will result in
```
default:
    Console.WrieLine("hello world\n");

    return;
```
if `nl_before_return = true` with both `master` and the proposed change, and stay unchanged if `nl_before_return = false`.  I believe that this is the intended behavior.

Having said that, extra newlines that are already in the input will indeed be kept, that is,
```
default:

    return;
```
will stay unchanged regardless of the `nl_before_return` setting. Likewise, for
```
default: return;
```
no newline will be inserted even with `nl_before_return = true`. It seems that `nl_after_case = false` prevents this.

PR #2700, opened 2020-03-21
Open stdin in binary mode to preserve newline characters

### LINKED ISSUE #2686: Converts CRLF to LF when newlines=auto and input is from STDIN
The config option `newlines=auto` is supposed to preserve the existing newline style, but the behaviour is equivalent to `newlines=lf` if the input comes from STDIN.

With:

- [infile.c](https://github.com/uncrustify/uncrustify/files/4304729/infile.c.txt)
- [newlines.cfg](https://github.com/uncrustify/uncrustify/files/4304730/newlines.cfg.txt)

```bash
$ file infile.c
infile.c: C source, ASCII text, with CRLF line terminators

$ <infile.c uncrustify -c newlines.cfg -o outfile.c

$ file outfile.c
outfile.c: C source, ASCII text
```

Notice that the output file is no longer flagged as having CRLF line terminators.

__Uncrustify Version:__ Uncrustify-0.70.1_f (Windows 10)
__Relevant config:__ `newlines=auto`

shoogle:
Fixes #2686

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -1031,6 +1031,13 @@ static bool read_stdin(file_mem &fm)
    fm.data.clear();
    fm.enc = char_encoding_e::e_ASCII;
 
+   // Re-open stdin in binary mode to preserve newline characters
+#ifdef WIN32
+   _setmode(_fileno(stdin), _O_BINARY);
+#else
+   freopen(NULL, "rb", stdin);
+#endif
+
    while (!feof(stdin))
    {
       int len = fread(buf, 1, sizeof(buf), stdin);


src/uncrustify.cpp
```diff
@@ -1031,6 +1031,13 @@ static bool read_stdin(file_mem &fm)
    fm.data.clear();
    fm.enc = char_encoding_e::e_ASCII;
 
+   // Re-open stdin in binary mode to preserve newline characters
+#ifdef WIN32
+   _setmode(_fileno(stdin), _O_BINARY);
+#else
+   freopen(NULL, "rb", stdin);
+#endif
```
shoogle, line 1039: @guy-maurel [asked](https://github.com/uncrustify/uncrustify/issues/2686#issuecomment-638931567):

> Why is it necessary to freopen(NULL, "rb", stdin); for the NOT Windows OS?

The short answer is that it is *probably* not necessary, but it *might* be necessary in some cases and *including it does no harm*.

CPlusPlus.com's [entry for `fopen`](http://www.cplusplus.com/reference/cstdio/fopen/) says:

> Depending on the environment where the application runs, some special character conversion may occur in input/output operations in text mode to adapt them to a system-specific text file format. Although on some environments no conversions occur and both text files and binary files are treated the same way, using the appropriate mode improves portability.

Notice that it __does not__ say:

- That Windows is the only platform that does character conversion.
- Or that newlines are the only characters that could get converted.

It *does* say that some platforms don't do conversions, and you might happen to know that Linux is one of those platforms, but it recommends using the same mode on all platforms. There could be an obscure platform that converts all characters beyond the ASCII range, or that uses `\r` (CR) instead of `/n` (LF) for newlines.

In any case, there's little cost to using `freopen`, and if `freopen` literally does nothing on Linux then hopefully it will be optimized out by the compiler anyway.

mikedld: @shoogle, @guy-maurel, it does at least some harm on Linux unfortunately. I know this got fixed already, but decided to mention for historical purposes (and since I spent some time on it without realizing it was already fixed as my distro still has 0.71.0_f). Maybe It'll also help someone else someday.

There is (or rather was) [Uncrustify extension for Visual Studio code](/LaurentTreguier/vscode-uncrustify); now, I don't know the exact reason it has been discontinued this year, but the extension got pulled from the Marketplace and repository is currently archived. Because of that, I tried setting up a couple of other extensions that allow you to do generic code formatting based on a configurable command that reads stdin and writes to stdout. And what a surprise, none of them worked with Uncrustify either.

After a short debugging, I managed to reproduce the issue. It has something to do with NodeJS (have no idea what, probably some kind of isolation) and leads to parent process getting `EPIPE` when trying to write to stdin, and child process getting `ENXIO` on `freopen()`. Below are the test programs (in C and NodeJS) and sample runs with/without strace, with/without `freopen()`.

<details><summary>test.c</summary>

```c
#include <stdio.h>
#include <string.h>

int main(int argc, char** argv)
{
  if (argc == 2 && strcmp(argv[1], "--reopen-stdin") == 0) {
    if (freopen(NULL, "rb", stdin) == NULL) {
      perror("freopen");
      return 1;
    }
  }

  char buf[32];
  while (!feof(stdin)) {
    fwrite(buf, 1, fread(buf, 1, sizeof(buf), stdin), stdout);
  }

  fflush(stdout);
  return 0;
}
```
</details>

<details><summary>test.js</summary>

```js
(async () => {
  let args = process.argv.slice(2);
  let reopenStdin = args.indexOf('--reopen-stdin') >= 0;
  let strace = args.indexOf('--strace') >= 0;

  let c = require('child_process').spawn(`${strace ? 'strace ' : ''}./test${reopenStdin ? ' --reopen-stdin' : ''}`, { shell: true });

  let r = new require('stream').Readable();
  r._read = () => {};
  r.push('Hello, world!');
  r.push(null);
  r.pipe(c.stdin).on('error', err => {
    console.log(`error (stdin): ${err.message}`);
  });

  let stdout = '';
  c.stdout.on('data', chunk => stdout += chunk);

  let stderr = '';
  c.stderr.on('data', chunk => stderr += chunk);

  c.on('close', code => {
    console.log(`code: ${code}`);
    console.log(`stdout:\n${stdout.replace(/^/mg, '> ')}`);
    console.log(`stderr:\n${stderr.replace(/^/mg, '> ')}`);
  });
})();
```
</details>

<details><summary>Output</summary>

```
# gcc test.c -otest
# node test.js
code: 0
stdout:
> Hello, world!
stderr:
> 
# node test.js --reopen-stdin
error (stdin): write EPIPE
code: 1
stdout:
> 
stderr:
> freopen: No such device or address
> 
# node test.js --reopen-stdin --strace
error (stdin): write EPIPE
code: 1
stdout:
> 
stderr:
> execve("./test", ["./test", "--reopen-stdin"], 0x7ffdb874ee98 /* 90 vars */) = 0
> brk(NULL)                               = 0x55ac280d0000
> access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
> openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
> fstat(3, {st_mode=S_IFREG|0644, st_size=173570, ...}) = 0
> mmap(NULL, 173570, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7fd3f0d75000
> close(3)                                = 0
> openat(AT_FDCWD, "/lib64/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
> read(3, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0p?\2\0\0\0\0\0"..., 832) = 832
> fstat(3, {st_mode=S_IFREG|0755, st_size=1802424, ...}) = 0
> mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fd3f0d73000
> mmap(NULL, 1815824, PROT_READ, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7fd3f0bb7000
> mmap(0x7fd3f0bd9000, 1331200, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x22000) = 0x7fd3f0bd9000
> mmap(0x7fd3f0d1e000, 307200, PROT_READ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x167000) = 0x7fd3f0d1e000
> mmap(0x7fd3f0d69000, 24576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x1b1000) = 0x7fd3f0d69000
> mmap(0x7fd3f0d6f000, 13584, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7fd3f0d6f000
> close(3)                                = 0
> mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7fd3f0bb5000
> arch_prctl(ARCH_SET_FS, 0x7fd3f0d74580) = 0
> mprotect(0x7fd3f0d69000, 16384, PROT_READ) = 0
> mprotect(0x55ac26833000, 4096, PROT_READ) = 0
> mprotect(0x7fd3f0dca000, 4096, PROT_READ) = 0
> munmap(0x7fd3f0d75000, 173570)          = 0
> openat(AT_FDCWD, "/proc/self/fd/0", O_RDONLY) = -1 ENXIO (No such device or address)
> close(0)                                = 0
> dup(2)                                  = 0
> fcntl(0, F_GETFL)                       = 0x2 (flags O_RDWR)
> brk(NULL)                               = 0x55ac280d0000
> brk(0x55ac280f1000)                     = 0x55ac280f1000
> fstat(0, {st_mode=S_IFSOCK|0777, st_size=0, ...}) = 0
> write(0, "freopen: No such device or addre"..., 35freopen: No such device or address
> ) = 35
> close(0)                                = 0
> exit_group(1)                           = ?
> +++ exited with 1 +++
> 
```
</details>

shoogle: It turns out that attempting to reopen `stdin` causes problems when the program is run by a NodeJS script on Linux (other platforms were not affected). See #2798.

PR #2707, opened 2020-03-27
Improve sorting for 'import' and 'include' statements

samishchandra:
Improve sorting for 'import' and 'include' statements. Additional options are provided to be able to control the sorting behaviour of these directives.


samishchandra:
Trying to submit individual PRs for each commit on this like #2710.

PR #2715, opened 2020-04-01
Improve add features include import sorting

samishchandra:
These series of commits are to improve sorting of `#import` and `#include` directives. Added bunch of features that makes code beautiful. Additional options are provided to be able to control the sorting behaviour of these directives.


samishchandra:
Hi @guy-maurel,

Updated to make all tests pass.

Issue #1813, opened 2018-06-17
Extra indent in lambda body with indent_cpp_lambda_only_once


nsubiron:
I'm using `indent_cpp_lambda_only_once = true`, and `indent_namespace_single_indent = true` (Uncrustify 0.67); and the indentation of a lambda body depends on how many nested namespaces I have

```cpp
// This works, input = output.
namespace n1 {

  void func() {
    another_func([]() {
      return 42;
    });
  }

}
```

```cpp
// Lambda body gets indented.
namespace n1 {
namespace n2 {

  void func() {
    another_func([]() {
        return 42;
      });
  }

}
}
```

```cpp
// Lambda body gets indented twice.
namespace n1 {
namespace n2 {
namespace n3 {

  void func() {
    another_func([]() {
          return 42;
        });
  }

}
}
}
```

Here my full config file: [uncrustify.cfg.zip](https://github.com/uncrustify/uncrustify/files/2109226/uncrustify.cfg.zip).

rrsettgast:
I have a similar problem. with the following config:

```
indent_with_tabs = 0
output_tab_size = 2
indent_continue = output_tab_size
indent_columns = output_tab_size
indent_cpp_lambda_body = false
nl_cpp_ldef_brace = force
indent_cpp_lambda_only_once = true
```

the following code
```
namespace blah
{
void func0()
{
  functionThatTakesALambda( [&]()->void
  {
    lambdaBody();
  });
}

int func()
{
  const auto f = [](int i, int j) -> int
  {
    return i + j;
  };
}
}
```
becomes
```
namespace blah
{
void func0()
{
  functionThatTakesALambda( [&]() -> void
    {
      lambdaBody();
    });
}

int func()
{
  const auto f = [](int i, int j) -> int
    {
      return i + j;
    };
}
}
```

mihaipopescu:
There were some fixes for C# delegates indentation but I'm unaware if the options should have covered also C++ lambdas.

rolevax:
The same problem also occurs when `indent_namespace = false`

rrsettgast:
Here is a better example:
Running with 
```
indent_cpp_lambda_body = false
indent_namespace= false
```

on the following
```
void func0()
{
functionThatTakesALambda( [&]()
{
lambdaBody();
});
}
```
gives what I would call "correct behavior", of which the output running with the ```-p``` option gives
```
# Uncrustify-0.68-172-81f03dc
newlines                        = lf
[...]

PR #2716, opened 2020-04-01
allow word to be present in between [] and () when indentifying a  c++ lambda

rrsettgast:
This PR is related to #1813 . The code change seeks to allow for the inclusion of a lambda specifier between `[]` and `()` in a lambda function. Specifically this allows for proper identification of a lambda when a lambda specifier is included. Without this, the following example was incorrectly formatted:
```
Uncrustify_d-0.70.1-287-0625f89e

../uncrustify -c uncrustify.cfg -f test.cpp 
Parsing: test.cpp as language CPP
namespace blah0
{
namespace blah
{
void func0()
{
  functionThatTakesALambda( [&] __device__ ()->void
      {
        lambdaBody;
      });
}
}
}
```

This change results in:
```
namespace blah0
{
namespace blah
{
void func0()
{
  functionThatTakesALambda( [&] __device__ () -> void
  {
    lambdaBody;
  });
}
}
}
```

I do have concerns about how this change will conflict with the `sp_cpp_lambda_square_paren` option as it conflicts with the premise that is only whitespace possible between `[] ()`

### DIFF
--- a/documentation/how_to_format_cpp_lambda_functions.rst
+++ b/documentation/how_to_format_cpp_lambda_functions.rst
@@ -3,64 +3,37 @@ Formatting C++ Lambda Expressions
 #################################
 
 Uncrustify supports some formatting of c++ lambda expressions, although the support is incomplete.
-A complete description of the c++ lambda expressions 
 The parts of c++ lambda expressions that are currently recognized by Uncrusitify are:
 
 .. code-block:: c++
 
   [ captures ] execution_context ( params ) specifiers -> ret { body }
   
-Explanations for all the tokens in the above lambda expression (except for execution_context) are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
-The `execution_context` token is a non-standard addition to allow for specification of the execution space (e.g. host or device in CUDA).
-The native specifiers for the `execution_context` for lambda expression in CUDA are ``__device__`` and ``__host__ __device__``.
+Explanations for all the tokens in the above lambda expression (except for execution_context),
+as well as a complete description of c++ lambda expressions are found
+`here <https://en.cppreference.com/w/cpp/language/lambda>`_.
+The `execution_context` token is a non-standard addition to allow for specification of the 
+execution space (e.g. host or device in CUDA).
+The native specifiers for the `execution_context` for lambda expression in CUDA are 
+``__device__`` and ``__host__ __device__``.
 However, it is common for code to use a preprocessor variable in place of the native specifiers.
 
-The Uncrustify options for formatting of c++ lambda expressions are as follows:
+The Uncrustify options for formatting of c++ lambda expressions are:
 
 .. code-block::
 
-  # Add or remove space around '=' in C++11 lambda capture specifications.
-  #
-  # Overrides sp_assign.
-  sp_cpp_lambda_assign            = ignore   # ignore/add/remove/force
-  
-  # Add or remove space after the capture specification of a C++11 lambda when
-  # an argument list is present, as in '[] <here> (int x){ ... }'.
-  sp_cpp_lambda_square_paren      = ignore   # ignore/add/remove/force
-  
-  # Add or remove space after the capture specification of a C++11 lambda with
-  # no argument list is present, as in '[] <here> { ... }'.
-  sp_cpp_lambda_square_brace      = ignore   # ignore/add/remove/force
-  
-  # Add or remove space after the argument list of a C++11 lambda, as in
-  # '[](int x) <here> { ... }'.
-  sp_cpp_lambda_paren_brace       = ignore   # ignore/add/remove/force
-  
-  # Add or remove space between a lambda body and its call operator of an
-  # immediately invoked lambda, as in '[]( ... ){ ... } <here> ( ... )'.
-  sp_cpp_lambda_fparen            = ignore   # ignore/add/remove/force
-  
-  # Whether to indent the body of a C++11 lambda.
-  indent_cpp_lambda_body          = false    # true/false
-  
-  # Don't split one-line C++11 lambdas, as in '[]() { return 0; }'.
-  nl_cpp_lambda_leave_one_liners  = false    # true/false
-  
-  # Add or remove newline between C++11 lambda signature and '{'.
-  nl_cpp_ldef_brace               = ignore   # ignore/add/remove/force
-  
-  # The value might be used twice:
-  # - at the assignment
-  # - at the opening brace
-  #
-  # To prevent the double use of the indentation value, use this option with the
-  # value 'true'.
-  #
-  # true:  indentation will be used only once
-  # false: indentation will be used every time (default)
-  indent_cpp_lambda_only_once     = false    # true/false
+  sp_cpp_lambda_assign
+  sp_cpp_lambda_square_paren
+  sp_cpp_lambda_square_brace
+  sp_cpp_lambda_paren_brace
+  sp_cpp_lambda_fparen
+  nl_cpp_lambda_leave_one_liners
+  nl_cpp_ldef_brace
+  indent_cpp_lambda_body
+  indent_cpp_lambda_only_once
 
-  
+Please refer to the example configuration file at 
+`uncrustify/documentation/htdocs/default.cfg <https://github.com/uncrustify/uncrustify/blob/master/documentation/htdocs/default.cfg>`_ for an explanation of the options.
 Additionally, a multiple number of ``execution_context`` tokens may be set in the configuration file:
 
 .. code-block::
--- a/documentation/how_to_format_cpp_lambda_functions.rst
+++ b/documentation/how_to_format_cpp_lambda_functions.rst
@@ -5,22 +5,20 @@ Formatting C++ Lambda Expressions
 Uncrustify supports some formatting of c++ lambda expressions, although 
 the support is incomplete.
 The parts of c++ lambda expressions that are currently recognized by Uncrusitify
- are:
+are:
 
 .. code-block:: c++
 
   [ captures ] execution_context ( params ) specifiers -> ret { body }
   
-Explanations for all the tokens in the above lambda expression (except for 
-execution_context),
-as well as a complete description of c++ lambda expressions are found
-`here <https://en.cppreference.com/w/cpp/language/lambda>`_.
-The `execution_context` token is a non-standard addition to allow for 
-specification of the 
-execution space (e.g. host or device in CUDA).
+Explanations for all the tokens in the above lambda expression
+(except for ``execution_context``),
+as well as a complete description of c++ lambda expressions,
+are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
+The ``execution_context`` token is a non-standard addition to allow for 
+specification of the execution space (e.g. host or device in CUDA).
 The native specifiers for the `execution_context` for lambda expression in CUDA 
-are 
-``__device__`` and ``__host__ __device__``.
+are ``__device__`` and ``__host__ __device__``.
 However, it is common for code to use a preprocessor variable in place of the 
 native specifiers.
 
--- a/documentation/how_to_format_cpp_lambda_functions.rst
+++ b/documentation/how_to_format_cpp_lambda_functions.rst
@@ -4,7 +4,7 @@ Formatting C++ Lambda Expressions
 
 Uncrustify supports some formatting of c++ lambda expressions, although the support is incomplete.
 A complete description of the c++ lambda expressions 
-The parts of c++ lambda expressions that is currently recognized by Uncrusitify is:
+The parts of c++ lambda expressions that are currently recognized by Uncrusitify are:
 
 .. code-block:: c++
 
@@ -13,11 +13,11 @@ The parts of c++ lambda expressions that is currently recognized by Uncrusitify 
 Explanations for all the tokens in the above lambda expression (except for execution_context) are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
 The `execution_context` token is a non-standard addition to allow for specification of the execution space (e.g. host or device in CUDA).
 The native specifiers for the `execution_context` for lambda expression in CUDA are ``__device__`` and ``__host__ __device__``.
-However it is common for codes to use a preprocessor variable in place of the native specifiers.
+However, it is common for code to use a preprocessor variable in place of the native specifiers.
 
 The Uncrustify options for formatting of c++ lambda expressions are as follows:
 
-.. code-block:: c++
+.. code-block::
 
   # Add or remove space around '=' in C++11 lambda capture specifications.
   #
@@ -61,16 +61,19 @@ The Uncrustify options for formatting of c++ lambda expressions are as follows:
   indent_cpp_lambda_only_once     = false    # true/false
 
   
-Additionally, a multiple number of  "execution_context" token may be set in the configuration file using the ``EXECUTION_CONTEXT`` token.
-Each specification of the ``EXECTION_CONTEXT`` populate a list of allowable keywords to be used at the "execution_context" location.
-For example, the following shows setting 4 values (2 default, and 2 custom) for the execution_context.
+Additionally, a multiple number of ``execution_context`` tokens may be set in the configuration file:
 
-.. code-block:: c++
+.. code-block::
 
-  set EXECUTION_CONTEXT __device__
   set EXECUTION_CONTEXT __host__ __device__
-  set EXECUTION_CONTEXT DEVICE_LAMBDA_CONTEXT
-  set EXECUTION_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT
+  set EXECUTION_CONTEXT DEVICE_LAMBDA_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT
 
-The effect of these lines in the confiuration file is that any of the strings (__device__, __host__ __device__, DEVICE_LAMBDA_CONTEXT, HOST_DEVICE_LAMBDA_CONTEXT) will be recognized by uncrusitfy and will allow the lambda to be properly identified.
-
+The effect of these lines in the configuration file is that any of the strings
+(``__host__``, ``__device__``,
+``DEVICE_LAMBDA_CONTEXT``, ``HOST_DEVICE_LAMBDA_CONTEXT``)
+will be recognized by uncrusitfy
+and will allow the lambda to be properly identified.
+Note that each word after the token name
+(``EXECUTION_CONTEXT`` in this instance) is a separate token.
+This means that uncrustify will parse ``__host__`` and ``__device__``
+as separate tokens, and there is no need to specify ``__device__`` twice.
--- a/documentation/how_to_format_cpp_lambda_functions.rst
+++ b/documentation/how_to_format_cpp_lambda_functions.rst
@@ -1,0 +1,76 @@
+#################################
+Formatting C++ Lambda Expressions
+#################################
+
+Uncrustify supports some formatting of c++ lambda expressions, although the support is incomplete.
+A complete description of the c++ lambda expressions 
+The parts of c++ lambda expressions that is currently recognized by Uncrusitify is:
+
+.. code-block:: c++
+
+  [ captures ] execution_context ( params ) specifiers -> ret { body }
+  
+Explanations for all the tokens in the above lambda expression (except for execution_context) are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
+The `execution_context` token is a non-standard addition to allow for specification of the execution space (e.g. host or device in CUDA).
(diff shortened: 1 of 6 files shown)


rrsettgast:
> why allowing "WORD" and not a new defined keyword "**device**" ?
> Please join one/some test(s) example(s) to prove it is working.
> Choose a number in tests/cpp.test

Hello. A common thing would be to specify a preprocessor macro in place of `__device__` so the "word"  could be anything. For example something like this is appropriate:
```
#ifdef CUDA
#define MYCODE_DEVICE_LAMBDA __device__
#else 
#define MYCODE_DEVICE_LAMBDA
#endif 
```
Then usage in some lambda would be:
```
function( [=] MYCODE_DEVICE_LAMBDA () 
{ 
  lambda body...
} );
```

mwoehlke:
Interesting. I take it this is a CUDA extensions? (AFAIK it is *not* legal C++.)

It might be better to do as @gmaurel suggests and only accept a specific token type. I would then suggest to use your configuration to make whatever arbitrary text your project uses parse as that token type.

mwoehlke:
Also, please add a unit test.

rrsettgast:
> Interesting. I take it this is a CUDA extensions? (AFAIK it is _not_ legal C++.)
> 

Oh, yeah. It is not standard C++. This is probably a CUDA extension, but will also be the case for HIP. It is also a requirement for many of the platform portability abstraction layers (LLNL/RAJA, SANIDA/KOKKOS, etc.).

> It might be better to do as @gmaurel suggests and only accept a specific token type. I would then suggest to use your configuration to make whatever arbitrary text your project uses parse as that token type.

We have several codes at my institution that manage this in their own way (i.e. use their own preprocessor variables and treatment), so using a hard coded keyword might not be flexible enough to be useful to others. Perhaps we can specify a list of keywords in the configuration file...but it does seem to pass tests allowing a single "word".

rrsettgast:
> Also, please add a unit test.

I added this to the test for `issue_1813`.

gmaurel:
I prefer another way.

1. Define a new token such as CT_CUDA (or anything else) see src/token_enum.h
2. Use it where necessary with a preprocessor guard such as
```.cpp
  #ifdef USE_WITH_CUDA
     some code
  #endif 
```
     to control the compilation, in the file token_enum.h and where you are using it.
3. The user may complete his configuration file with some more lines such as
```.cfg
sp_arith   = force
...
set CUDA ___device___
set some more ...
```
4. Prepare a file "documentation/how_to_use_CUDA.txt" to help the new users.

mwoehlke:
> Use it where necessary with a preprocessor guard

Requiring a custom build of uncrustify itself seems a bit much, especially if the token doesn't match anything unless specified in the user's config. Otherwise, however, I agree. I'd suggest `CT_CUDA_LAMBDA_CONTEXT` though, or maybe just `CT_CUDA_CONTEXT` if `__host__`, `__device__` and `__global__` are usable elsewhere. (Ideally, we'd add a CUDA language and only recognize those in CUDA mode, in which case recognizing the default values should be reasonably safe... though honestly it probably is anyway.)

rrsettgast:
> I prefer another way.
> 
> 1. Define a new token such as CT_CUDA (or anything else) see src/token_enum.h

> Otherwise, however, I agree. I'd suggest `CT_CUDA_LAMBDA_CONTEXT` though, or maybe just `CT_CUDA_CONTEXT` if `__host__`, `__device__` and `__global__` are usable elsewhere. (Ideally, we'd add a CUDA language and only recognize those in CUDA mode, in which case recognizing the default values should be reasonably safe... though honestly it probably is anyway.)

So that I understand what is being asked here. Are you asking that we create a new token to signify when we are specifying a lambda specifier? So this would be something like `CT_LAMBDA_CONTEXT` since  I don't think this will be a `CUDA` specific thing. So then how do we signify the strings that are allowed to serve as `CT_LAMBDA_CONTEXT`?

> 2. Use it where necessary with a preprocessor guard such as
> 
> ```c++
>   #ifdef USE_WITH_CUDA
>      some code
>   #endif 
> ```
> 
> ```
>  to control the compilation, in the file token_enum.h and where you are using it.
> ```
> 

I don't think this is a good idea. 

> 3. The user may complete his configuration file with some more lines such as
> 
> ```haproxy-config
> sp_arith   = force
> ...
> set CUDA ___device___
> set some more ...
> ```
> 
I don't understand this at all. Where are the `set CUDA __device__` done?

> 4. Prepare a file "documentation/how_to_use_CUDA.txt" to help the new users.

This is slightly more scope that I am expecting. Wouldn't it be better to have a unified strategy for CUDA which was not authored by someone who knows very little about how uncrustify works? (i.e. me :) )


As a counter-proposal can we specify a set of strings that are allowed to be used for a `CT_LAMBDA_CONTEXT` in the configuration file? Maybe a comma delimited string? Then it would be straight forward to check if the chunk following the `]` was one of these strings. As you may notice, I would like to make this general enough to be applied to several different strings in a single code, as some lambda context statements may be named various things in the source, even though the preprocessor would ultimately make them `__device__` or `__host__ __device__`

mwoehlke:
> So that I understand what is being asked here. Are you asking that we create a new token to signify when we are entering a lambda context statement?

This doesn't make sense. Probably because...

> So this would be something like `CT_LAMBDA_CONTEXT` since I don't think this will be a CUDA specific thing.

...this is wrong. Specifying a context for a lambda *is* a CUDA extension. It is not presently allowed by C++, nor am I aware of any active proposals to add such a thing.

Based on this and the preceding, I think you are misunderstanding what I mean by "context". I am talking about a keyword that *specifies* the context of a lambda. Think "similar to `CT_QUALIFIER`".

> So then how do we signify the strings that are allowed to serve as CT_LAMBDA_CONTEXT?

Using the configuration: `set CUDA_CONTEXT MYCODE_DEVICE_LAMBDA`. Which (per your other question) is done in your `uncrustify.cfg` (or however you name it).

> As a counter-proposal can we specify a set of strings that are allowed to be used for a `CT_LAMBDA_CONTEXT` in the configuration file?

Yes. That is *exactly* what we are suggesting :slightly_smiling_face:. (I'm not sure if it is comma-delimited; I believe not, but I'd have to check how argument splitting works. Anyway, you don't need to implement this stuff, it is already implemented! For example, you could say `set QUALIFIER MYLIB_CONST` today and it will cause uncrustify to treat `MYLIB_CONST` as if it were e.g. `const`.)

rrsettgast:
@mwoehlke OK. I get it i think. I added the `CT_CUDA_LAMBDA_CONTEXT ` token, and check for the `CT_CUDA_LAMBDA_CONTEXT` instead of `CT_WORD`. Adding two `set` statements in the config file seem to do the trick automagically. I added two more examples with illustrating the use of the two different `set` statements. 

Is this what you and @gmaurel had in mind?

tests/config/Issue_1813.cfg
```diff
@@ -4,3 +4,6 @@ indent_namespace               = true
 indent_namespace_single_indent = true
 use_indent_continue_only_once  = true
 indent_cpp_lambda_only_once    = true
+sp_cpp_lambda_paren_brace      = add
+set CUDA_LAMBDA_CONTEXT DEVICE_LAMBDA_CONTEXT
+set CUDA_LAMBDA_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT
```
mwoehlke, line 9: Missing newline

mwoehlke:
> Is this what you and @gmaurel had in mind?

I think so. (LGTM anyway, aside from nits; was how I understood Guy's comments also, but since I am not Guy, I am not 100% certain :wink:. Aside from the one review comment, I wonder about the churn from changing the namespace names; is there a benefit?)

Can `__device__`, etc. appear in contexts *other* than a lambda? If yes, `CT_CUDA_CONTEXT` might be preferred over `CT_CUDA_LAMBDA_CONTEXT`.

rrsettgast:
> > Is this what you and @gmaurel had in mind?
> 
> I think so. (LGTM anyway, aside from nits; was how I understood Guy's comments also, but since I am not Guy, I am not 100% certain 😉. Aside from the one review comment, I wonder about the churn from changing the namespace names; is there a benefit?)
> 
Aside from the silliness that was in there previously. I had posted that example on the issue, but didn't expect it to go into the code! At any rate, I just though it would be better to give the namespaces reasonable names.

> Can `__device__`, etc. appear in contexts _other_ than a lambda? If yes, `CT_CUDA_CONTEXT` might be preferred over `CT_CUDA_LAMBDA_CONTEXT`.

Yes. I am not expect on the subject, but `__global__`, `__device__`, `__host__` and __host__ __device__ are specifiers used to indicate whether a function is callable globally, on host, and/or device. I am sure my explanation is clumsy. See https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#c-language-extensions
Uncrustify seems to handle the function specifiers just fine.

The usage for lambda is discussed here: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#extended-lambda

There are a lot of places that `__device__` and other specifiers are used...I think it is a pretty large undertaking to properly support CUDA extensions in `Uncrustify`.

This of course illustrates that I have failed to account for the case where `__host__ __device__` is the specifier. Uncrustify will chunk these individually, so it has to be checked for. 

One other thing that I will keep going back to is that these sort of specifiers may originate from CUDA, but they are not unique to CUDA. They are in Hip as well: https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_kernel_language.md#__device__


Would you be OK with calling it: `CT_EXECUTION_SPACE_SPECIFIER`?

guy-maurel:
I take the second part of my comment back. A preprocessor guard is not a solution.
The c_token_t CT_EXECUTION_SPACE_SPECIFIER is pretty long.
Please look for a shorter one.
I am still missing some file "documentation/how_to_use_xxx.txt" to help the users.
Thanks.

mwoehlke:
> Please look for a shorter one.

`CT_EXECUTION_CONTEXT`? I'd be okay with that. (Guy? Randolph?)

> I have failed to account for the case where __host__ __device__ is the specifier.

This is probably as simple as using `while` instead of `if` to skip these:

```c++
   while (pa_o->type == CT_EXECUTION_CONTEXT)
   {
      // set lambda-declarator to next chunk after specifier
      pa_o = chunk_get_next_ncnl(pa_o);
   }
```

(Also, I didn't notice earlier, but you don't really need the extra local `lambda_specifier`. Besides replacing that with `pa_o`, literally the only changes I made were replacing `if` with `while` and using my suggestion for the chunk type name.)

rrsettgast:
> I take the second part of my comment back. A preprocessor guard is not a solution.
> The c_token_t CT_EXECUTION_SPACE_SPECIFIER is pretty long.
> Please look for a shorter one.
> I am still missing some file "documentation/how_to_use_xxx.txt" to help the users.
> Thanks.

I changed it to CT_EXECUTION_CONTEXT.

Also added a ho_to document for formatting of c++ lambda expressions.

rrsettgast:
> > Please look for a shorter one.
> 
> `CT_EXECUTION_CONTEXT`? I'd be okay with that. (Guy? Randolph?)
> 
done.

> > I have failed to account for the case where **host** **device** is the specifier.
> 
> This is probably as simple as using `while` instead of `if` to skip these:
> 
> ```c++
>    while (pa_o->type == CT_EXECUTION_CONTEXT)
>    {
>       // set lambda-declarator to next chunk after specifier
>       pa_o = chunk_get_next_ncnl(pa_o);
>    }
> ```
> 
> (Also, I didn't notice earlier, but you don't really need the extra local `lambda_specifier`. Besides replacing that with `pa_o`, literally the only changes I made were replacing `if` with `while` and using my suggestion for the chunk type name.)

I removed the `lambda_specifier`. I had thought it might not be desirable to use `pa_o` for this, since I read `pa_o` as "Parenthesis Open". But it is better to just cut out the middle man.

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -0,0 +1,76 @@
...
+The parts of c++ lambda expressions that is currently recognized by Uncrusitify is:
+
+.. code-block:: c++
+
+  [ captures ] execution_context ( params ) specifiers -> ret { body }
+  
+Explanations for all the tokens in the above lambda expression (except for execution_context) are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
+The `execution_context` token is a non-standard addition to allow for specification of the execution space (e.g. host or device in CUDA).
+The native specifiers for the `execution_context` for lambda expression in CUDA are ``__device__`` and ``__host__ __device__``.
+However it is common for codes to use a preprocessor variable in place of the native specifiers.
```
mwoehlke, line 16:
```suggestion
However, it is common for code to use a preprocessor variable in place of the native specifiers.
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -0,0 +1,76 @@
...
+
+The Uncrustify options for formatting of c++ lambda expressions are as follows:
+
+.. code-block:: c++
```
mwoehlke, line 20: This isn't C++, please omit that.
```suggestion
.. code-block::
```

Also, I'm a little dubious of copying the documentation here. While the added value is obvious, the chance for it to drift out of sync is high. I might replace `s/are as follows/are/` above and then just list the relevant options, either as a (not-C++) `code-block`, or perhaps even as an actual list of literals, i.e.:

```
 - ``sp_cpp_lambda_assign``
 - ``sp_cpp_lambda_square_paren``
 ...
```

(If you do the latter, don't forget a blank line before and after the list, or docutils won't format it properly.)

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -0,0 +1,76 @@
...
+
+  set EXECUTION_CONTEXT __device__
+  set EXECUTION_CONTEXT __host__ __device__
+  set EXECUTION_CONTEXT DEVICE_LAMBDA_CONTEXT
+  set EXECUTION_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT
+
+The effect of these lines in the confiuration file is that any of the strings (__device__, __host__ __device__, DEVICE_LAMBDA_CONTEXT, HOST_DEVICE_LAMBDA_CONTEXT) will be recognized by uncrusitfy and will allow the lambda to be properly identified.
+
```
mwoehlke, line 76: "tokens" s.b. plural here. Also, maybe use ``` ``execution_context`` ``` for consistency?

Actually, I'd be inclined to rewrite this whole section:
```suggestion
Additionally, a multiple number of ``execution_context`` tokens may be set in the configuration file:

.. code-block::

  set EXECUTION_CONTEXT __host__ __device__
  set EXECUTION_CONTEXT DEVICE_LAMBDA_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT

The effect of these lines in the configuration file is that any of the strings
(``__host__``, ``__device__``,
``DEVICE_LAMBDA_CONTEXT``, ``HOST_DEVICE_LAMBDA_CONTEXT``)
will be recognized by uncrusitfy
and will allow the lambda to be properly identified.
Note that each word after the token name
(``EXECUTION_CONTEXT`` in this instance) is a separate token.
This means that uncrustify will parse ``__host__`` and ``__device__``
as separate tokens, and there is no need to specify ``__device__`` twice.
```

(Note also my later comment for more explanation. Also, please omit the extra blank line at the end of the file.)

tests/config/Issue_1813.cfg
```diff
@@ -4,3 +4,8 @@ indent_namespace               = true
 indent_namespace_single_indent = true
 use_indent_continue_only_once  = true
 indent_cpp_lambda_only_once    = true
+sp_cpp_lambda_paren_brace      = add
+set EXECUTION_CONTEXT DEVICE_LAMBDA_CONTEXT
+set EXECUTION_CONTEXT HOST_DEVICE_LAMBDA_CONTEXT
+set EXECUTION_CONTEXT __device__
+set EXECUTION_CONTEXT __host__ __device__
```
mwoehlke, line 11: Nitpick: As (sort-of) explained above, this is redundant; the syntax of `set` is `set <TYPE> <TOKEN> [<TOKEN>...]`. Each word is a separate token. So, here, you are making `__host__` a token (desired), and also making `__device__` a token *again*, which you already did in the previous line.

```suggestion
set EXECUTION_CONTEXT __host__ __device__
```

mwoehlke:
> I had thought it might not be desirable to use `pa_o` for this, since I read `pa_o` as "Parenthesis Open". But it is better to just cut out the middle man.

Heh. I read it as "oddly chosen name for a local" :smile:, but you may be right. Since the old code still *initially* set `pa_o` to the next token (which might be `CT_EXECUTION_CONTEXT`, and since we never actually need that chunk pointer... Anyway, I'm pretty sure this pattern of 'guess where the chunk we want will be and fix it if we're wrong' also happens elsewhere.

documentation/how_to_format_cpp_lambda_functions.rst
rrsettgast: Oh. I didn't know you could set multiple strings on the same set command.

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
 Formatting C++ Lambda Expressions
 #################################
 
-Uncrustify supports some formatting of c++ lambda expressions, although the support is incomplete.
-A complete description of the c++ lambda expressions 
-The parts of c++ lambda expressions that is currently recognized by Uncrusitify is:
+Uncrustify supports some formatting of c++ lambda expressions, although 
+the support is incomplete.
+The parts of c++ lambda expressions that are currently recognized by Uncrusitify
+ are:
```
mwoehlke, line 8:
```suggestion
are:
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
...
 .. code-block:: c++
 
   [ captures ] execution_context ( params ) specifiers -> ret { body }
   
-Explanations for all the tokens in the above lambda expression (except for execution_context) are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
-The `execution_context` token is a non-standard addition to allow for specification of the execution space (e.g. host or device in CUDA).
-The native specifiers for the `execution_context` for lambda expression in CUDA are ``__device__`` and ``__host__ __device__``.
-However it is common for codes to use a preprocessor variable in place of the native specifiers.
+Explanations for all the tokens in the above lambda expression (except for 
+execution_context),
```
mwoehlke, line 15:
```suggestion
Explanations for all the tokens in the above lambda expression
(except for ``execution_context``),
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
...
+as well as a complete description of c++ lambda expressions are found
+`here <https://en.cppreference.com/w/cpp/language/lambda>`_.
```
mwoehlke, line 17:
```suggestion
as well as a complete description of c++ lambda expressions,
are found `here <https://en.cppreference.com/w/cpp/language/lambda>`_.
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
...
+The `execution_context` token is a non-standard addition to allow for 
+specification of the 
+execution space (e.g. host or device in CUDA).
```
mwoehlke, line 20:
```suggestion
The ``execution_context`` token is a non-standard addition to allow for 
specification of the execution space (e.g. host or device in CUDA).
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
...
+The native specifiers for the `execution_context` for lambda expression in CUDA 
+are 
+``__device__`` and ``__host__ __device__``.
```
mwoehlke, line 23:
```suggestion
are ``__device__`` and ``__host__ __device__``.
```

documentation/how_to_format_cpp_lambda_functions.rst
```diff
@@ -2,75 +2,59 @@
...
+  sp_cpp_lambda_square_brace
+  sp_cpp_lambda_paren_brace
+  sp_cpp_lambda_fparen
+  nl_cpp_lambda_leave_one_liners
+  nl_cpp_ldef_brace
+  indent_cpp_lambda_body
+  indent_cpp_lambda_only_once
+
+Please refer to the example configuration file at 
+`uncrustify/documentation/htdocs/default.cfg <https://github.com/uncrustify/uncrustify/blob/master/documentation/htdocs/default.cfg>`_
```
mwoehlke, line 42: Ick. Oh, well, there really isn't anything that can be done with this :slightly_smiling_face:.

gmaurel:
@mwoehlke : can you remove your request change?

Issue #2720, opened 2020-04-02
Modernized raw strings are not handled correctly in OC/OC+ languages


samishchandra:
[Modernized raw strings](https://clang.llvm.org/extra/clang-tidy/checks/modernize-raw-string-literal.html) are not handled correctly in OC/OC+ languages.

### ⬇️Input:
```
NSString *predicate = [NSString stringWithFormat:@R"Something"];
```

### ✅Expected Output:
```
NSString *predicate = [NSString stringWithFormat:@R"Something"];
```

### ❌Actual Output:
```
NSString *predicate = [NSString stringWithFormat:@R "Something"];
```

### ⚙️Debugging info:
```
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>               TYPE|               NONE|     PARENT_NOT_SET[  1/  1/  9/  0][0/0/0][   10460000][0-0] NSString
#   1>           PTR_TYPE|               NONE|     PARENT_NOT_SET[ 10/ 10/ 11/  1][0/0/0][  110400000][0-0]          *
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 11/ 11/ 20/  0][0/0/0][   11840000][0-0]           predicate
#   1>             ASSIGN|               NONE|     PARENT_NOT_SET[ 21/ 21/ 22/  1][0/0/0][  100000000][0-0]                     =
#   1>        SQUARE_OPEN|             OC_MSG|     PARENT_NOT_SET[ 23/ 23/ 24/  1][0/0/0][  100044000][0-0]                       [
#   1>       OC_MSG_CLASS|               NONE|     PARENT_NOT_SET[ 24/ 24/ 32/  0][0/1/0][      74000][0-0]                        NSString
#   1>        OC_MSG_FUNC|               NONE|     PARENT_NOT_SET[ 33/ 33/ 49/  1][0/1/0][     184000][0-0]                                 stringWithFormat
#   1>           OC_COLON|               NONE|     PARENT_NOT_SET[ 49/ 49/ 50/  0][0/1/0][  100104000][0-0]                                                 :
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 50/ 50/ 52/  0][0/1/0][      64000][0-0]                                                  @R
#   1>             STRING|         PP_INCLUDE|     PARENT_NOT_SET[ 53/ 52/ 63/  0][0/1/0][       4000][0-0]                                                     "Something"
[...]

PR #2721, opened 2020-04-02
Fix issue on handling modernized raw string literals in OC/OC+

samishchandra:
Modernized raw strings are introduced in CPP 11 https://clang.llvm.org/extra/clang-tidy/checks/modernize-raw-string-literal.html. Fixed the issue to handle the raw string correctly and added required tests. Fixes issue #2720.


samishchandra:
Some inconsistency with the AppVeyor build.?

Issue #2727, opened 2020-04-04
Issue with OC block formatting in typedefs


samishchandra:
Issue with OC block formatting in typedefs. As noticed below `Handler` is incorrectly tokenized as `CT_WORD`.

### ⬇️Input:
```
typedef NSArray<NSString *> * (^ Handler)(void);
```

### ✅Expected Output:
```
typedef NSArray<NSString *> * (^ Handler)(void);
```

### ❌Actual Output:
```
typedef NSArray<NSString *> *(^ Handler)(void);
```

### ⚙️Debugging info:
> sp_after_oc_block_caret = remove

### ⚙️Debugging info:
```
# -=====-
# number of loops               = 4
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>            TYPEDEF|               NONE|     PARENT_NOT_SET[  1/  1/  8/  0][0/0/0][      70000][0-0] typedef
#   1>               TYPE|               NONE|     PARENT_NOT_SET[  9/  9/ 16/  1][0/0/0][         80][0-0]         NSArray
#   1>         ANGLE_OPEN|           TEMPLATE|     PARENT_NOT_SET[ 16/ 16/ 17/  0][0/0/0][  1000000c0][0-0]                <
#   1>               TYPE|               NONE|     PARENT_NOT_SET[ 17/ 17/ 25/  0][0/1/0][      400c0][0-0]                 NSString
#   1>           PTR_TYPE|               NONE|     PARENT_NOT_SET[ 26/ 26/ 27/  1][0/1/0][  1000000c0][0-0]                          *
#   1>        ANGLE_CLOSE|           TEMPLATE|     PARENT_NOT_SET[ 27/ 27/ 28/  0][0/0/0][  1000000c0][0-0]                           >
#   1>           PTR_TYPE|               NONE|     PARENT_NOT_SET[ 29/ 29/ 30/  1][0/0/0][  100040080][0-0]                             *
#   1>         PAREN_OPEN|               NONE|     PARENT_NOT_SET[ 30/ 31/ 32/  1][0/0/0][  100040080][0-0]                              (
#   1>              CARET|               NONE|     PARENT_NOT_SET[ 31/ 32/ 33/  0][0/1/0][  100040080][0-0]                               ^
[...]

PR #2729, opened 2020-04-04
Fix CT_TYPE is incorrectly tokenized as CT_WORD in OC block typedef

samishchandra:
Fixes issue #2727.


samishchandra:
Fixes issue #2727.

samishchandra:
Updated by adding `cpp` tests.

PR #2748, opened 2020-04-20
Improve consistency of 'nl_(before|after)_access_spec'

geimer:
The behavior of `nl_before_access_spec` was dependent on the settings of other newline-modifying options.  By handling this option late, it now takes precedence over other options such as `nl_after_func_proto`, etc. Also, to be consistent with `nl_before_access_spec`, `nl_after_access_spec` now no longer adds extra blank lines when the access specifier is located immediately before a closing brace.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -5633,22 +5633,6 @@ void do_blank_lines(void)
          }
       }
 
-      // Control blanks before an access spec
-      if (  (options::nl_before_access_spec() > 0)
-         && (options::nl_before_access_spec() != pc->nl_count)
-         && chunk_is_token(next, CT_ACCESS))
-      {
-         log_rule_B("nl_before_access_spec");
-
-         // Don't add blanks after a open brace
-         if (  prev == nullptr
-            || (prev->type != CT_BRACE_OPEN && prev->type != CT_VBRACE_OPEN))
-         {
-            blank_line_set(pc, options::nl_before_access_spec);
-            log_rule_B("nl_before_access_spec");
-         }
-      }
-
       // Control blanks before a class
       if (  (chunk_is_token(prev, CT_SEMICOLON) || chunk_is_token(prev, CT_BRACE_CLOSE))
          && get_chunk_parent_type(prev) == CT_CLASS)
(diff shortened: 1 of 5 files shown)


src/newlines.cpp
```diff
@@ -5928,6 +5920,23 @@ void do_blank_lines(void)
...
+      if (  (options::nl_before_access_spec() > 0)
+         && (options::nl_before_access_spec() != pc->nl_count)
+         && chunk_is_token(next, CT_ACCESS))
+      {
+         log_rule_B("nl_before_access_spec");
+
+         // Don't add blanks after an open brace
+         if (  prev == nullptr
+            || (  !chunk_is_token(prev, CT_BRACE_OPEN)
+               && !chunk_is_token(prev, CT_VBRACE_OPEN)))
```
geimer, line 5933: While this is equivalent to the original code, I am wondering whether VBRACEs are actually possible before/after an access specifier. I'm quite sure that this is impossible for C/C++, but don't know enough about some of the other supported languages.

geimer: Just found `chunk_is_not_token`, which is probably better than `!chunk_is_token`. However, both also test for `pc != nullptr`, which is always true in this case and doesn't need to be tested again. The compiler may be able to optimize this test away, however, this is a bit unclear from the (huge) assembly code blob of this file. Thus the question: Shall the refactoring be reverted to not rely on the compiler or is the slightly increased readability of `chunk_is_not_token` worth it?

gmaurel: YES! Refactoring would be fine with **chunk_is_not_token**

src/newlines.cpp
```diff
@@ -5686,7 +5686,15 @@ void do_blank_lines(void)
...
          log_rule_B("nl_after_access_spec");
-         blank_line_set(pc, options::nl_after_access_spec);
+
+         // Don't add blanks before a closing brace
+         if (  next == nullptr
+            || (  !chunk_is_token(next, CT_BRACE_CLOSE)
+               && !chunk_is_token(next, CT_VBRACE_CLOSE)))
+         {
+            blank_line_set(pc, options::nl_after_access_spec);
+            log_rule_B("nl_after_access_spec");
```
gmaurel, line 5696: let the log_rule_B before blank_line_set

PR #2749, opened 2020-04-20
Avoid extra blank lines after 'case' if 'nl_before_comment' > 1

geimer:
Similar to avoiding extra blank lines before comments after braces and virtual braces, also avoid them after `case` or `default` statements of a `switch` construct.


geimer:
C++-style comments are allowed since C99, see for example https://en.cppreference.com/w/c/comment

PR #2754, opened 2020-04-23
Improve 'nl_before_return' behavior in the presence of comments

geimer:
So far, the `nl_before_return` option handling only covers a very restricted set of use cases for comments in front of the `return` statement. For example, it fails to prevent adding extra blank lines after a (virtual) brace or `case` if a comment is in between the brace/`case` and the `return`. See the enhanced test cases in the first commit for more examples of broken behavior.

This PR reworks the `newline_before_return` function to handle such comments more thoroughly. (The diff looks a bit messy, though.)

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -2501,30 +2501,26 @@ static void newline_before_return(chunk_t *start)
 {
    LOG_FUNC_ENTRY();
 
-   chunk_t *nl = chunk_get_prev(start);
+   chunk_t *pc = chunk_get_prev(start);
+   chunk_t *nl = pc;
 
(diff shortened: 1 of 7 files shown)


guy-maurel:
- If the git diff command, or the diff part of the git gui,
      don't produce accurate output, it might be necessary to add
      some lines to the ~/.gitconfig file:
```.txt
         [diff]
                 algorithm = patience
         [gui]
                 diffopts = --patience

```

geimer:
This hardly makes a difference in this case.  The `newline_before_return` function has been more or less rewritten, and the diff algorithm just picks up some "random" braces and thinks that they have been kept...

Issue #2253, opened 2019-04-09
pp_indent_case does not work as intended.


Josar:
Maybe similar to https://github.com/uncrustify/uncrustify/issues/489

I try to define a cfg for uncrustify. But now i have this issue.
I think there are some options that clashes.

```
uint16_t testing(test_t tests)
{

    uint16_t rslt = 0;

    switch (tests)
    {
        #if (TEST_1 == TEST_2)
        case test1:
        {
            rslt = 1;
            break;
        }
        #elif (TEST_1 == TEST_3)
        case test1:
        {
            rslt = 2;
            break;
        }
            #endif

    }

    return rslt;
}
```
The `#end` is not handled correctly. 
Also the indenting of the `case` is not done properly.

My config
```
# The number of columns to indent per level.
# Usually 2, 3, 4, or 8. Default=8.
indent_columns                  = 4        # unsigned number

# # How to use tabs when indenting code
# # 0=spaces only
# # 1=indent with tabs to brace level, align with spaces (default)
# # 2=indent and align with tabs, using spaces when not on a tabstop
indent_with_tabs                = 0        # unsigned number

# #
# # Preprocessor options
# #

# # Control indent of preprocessors inside #if blocks at brace level 0 (file-level).
pp_indent                       = force   # ignore/add/remove/force

# # Whether to indent #if/#else/#endif at the brace level (True) or from column 1 (False).
pp_indent_at_level              = true    # false/true

# # Specifies the number of columns to indent preprocessors per level at brace level 0 (file-level).
# # If pp_indent_at_level=False, specifies the number of columns to indent preprocessors per level at brace level > 0 (function-level).
# # Default=1.
pp_indent_count                 = 3        # unsigned number

# # Add or remove space after # based on pp_level of #if blocks.
pp_space                        = remove   # ignore/add/remove/force

# # Sets the number of spaces added with pp_space.
pp_space_count                  = 0        # unsigned number

# The indent for #region and #endregion in C# and '#pragma region' in C/C++.
pp_indent_region                = 0        # number

# Whether to indent the code between #region and #endregion.
pp_region_indent_code           = false    # false/true

# # If pp_indent_at_level=True, sets the indent for #if, #else and #endif when not at file-level.
# # 0:  indent preprocessors using output_tab_size.
# # >0: column at which all preprocessors will be indented.
pp_indent_if                    = 0        # number

# # Control whether to indent the code between #if, #else and #endif.
pp_if_indent_code               = true    # false/true

[...]

PR #2765, opened 2020-04-28
Pp/ct case

Josar:
Tries to solve https://github.com/uncrustify/uncrustify/issues/2253

Endif is not handled correctly when after a case.


guy-maurel:
Don't forget to run (from build-directory):
`ctest --output-on-failure make`

gmaurel:
To many errors

Issue #2758, opened 2020-04-27
Missing `nl_func_call_end`


nopeslide:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

Uncrustify supports manipulating newlines after the opening paren of a function call:
```ini
# Whether to add a newline after '(' in a function call,
# has preference over nl_func_call_start_multi_line.
nl_func_call_start              = remove   # ignore/add/remove/force 
```
but not before the closing one.
Something like
```ini
# Whether to add a newline before ')' in a function call,
# has preference over nl_func_call_end_multi_line.
nl_func_call_end              = remove   # ignore/add/remove/force 
```
is missing.
When a style applies newlines, they have to be manually deleted if you want them removed again.

guy-maurel:
PR #2767 is a bugfix proposal.
Please test it and report.

PR #2767, opened 2020-04-29
Introduce a new option nl_func_call_end

guy-maurel:
Ref. #2758


nopeslide:
seems to work.
I updated my config and tested it on my codebase: the diff shows only manipulation of specified newlines.

PR #2775, opened 2020-05-11
suppress the warning

guy-maurel:
for stdin reopen


CDanU:
Warning: 

> ../src/uncrustify.cpp:1038:11: warning: ignoring return value of ‘FILE* freopen(const char*, const char*, FILE*)’, declared with attribute warn_unused_result [-Wunused-result]

PR #2786, opened 2020-05-16
Test for gcc on ubuntu 14

guy-maurel:
Onky for me, to test gcc


mihaipopescu:
Probably we'd need to add all Ubuntu LTS's in the test matrix anyway. Trusty, Xenial, Bionic, to Focal should all be in there.

PR #2790, opened 2020-05-20
Use modern way of setting C++ standard in CMake

AMDmi3:


AMDmi3:
Forgot to mention that it should fix `cc1plus: error: unrecognized command line option "-std=gnu++0x"` error I've encountered before with FreeBSD port of uncrustify, but I'm not sure which environment these were encountered on.

gmaurel:
On your progile here, the link http://amdmi3.ru doesn't work.
Could you have a look?

AMDmi3:
> On your progile here, the link http://amdmi3.ru doesn't work.

That is known and intentional. Thanks for merging.

PR #2818, opened 2020-06-17
D raw string

### LINKED ISSUE #2816: [D] [dlang] Backtick strings are not recognized as raw
Version: Uncrustify_d-0.71.0_f

D string literals quoted with backticks are raw, i.e. no escapes are expanded inside. Uncrustify attempts to expand escapes inside. This code causes complaints:

```d
auto s = `\`;
```

It should work.

CDanU:
fixes #2816

### DIFF
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -363,112 +363,112 @@ static bool d_parse_string(tok_ctx &ctx, chunk_t &pc)
 {
    size_t ch = ctx.peek();
 
-   if (  (ch == '"')
-      || (ch == '\'')
-      || (ch == '`'))
+   if (ch == '"' || ch == '\'' || ch == '`')
    {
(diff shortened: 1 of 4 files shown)


src/tokenize.cpp
```diff
@@ -1109,37 +1116,40 @@ static bool parse_string(tok_ctx &ctx, chunk_t &pc, size_t quote_idx, bool allow
...
+      // if last char in prev loop was escaped the one in the current loop isn't
+      if (escaped)
+      {
          escaped = false;
          continue;
       }
 
-      if (!escaped)
+      // see if the current char is a escape char
+      if (allow_escape)
```
andralex, line 1135: Great, I think this will work properly. I messed a bit with the previous code and it still escaped `termination_character`, so it had problems with e.g. `r"\"`.

Issue #2795, opened 2020-05-28
Uncrustify-0.71.0-21-d4da2b0f-dirty never returns


Massimo-B:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

```
$ uncrustify -v
Uncrustify-0.71.0-21-d4da2b0f-dirty
```

This command never returns and no debug output is created. Files are attached here.
```
$ uncrustify -l CPP -c uncrustify.cfg -f tttest.cpp  -p debug.txt
Parsing: tttest.cpp as language CPP
^C

```
[uncrustify.cfg.tar.gz](https://github.com/uncrustify/uncrustify/files/4694448/uncrustify.cfg.tar.gz)

[tttest.tar.gz](https://github.com/uncrustify/uncrustify/files/4694435/tttest.tar.gz)

guy-maurel:
First aid is to use the option:
```.cfg
# Limit the number of loops.
# Used by uncrustify.cpp to exit from infinite loop.
# 0: no limit.
debug_max_number_of_loops       = 10
```
It doesn't solved the problem but prevent the infinit loop.

guy-maurel:
You wish at the same configuration:
[...]

PR #2820, opened 2020-06-18
Remove infinite loop if the one liner is too long

guy-maurel:
Ref. #2795 
Mark the br_open with PCF_NOT_POSSIBLE

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -2212,24 +2212,27 @@ static void newlines_brace_pair(chunk_t *br_open)
             // make a vector to save the chunk
             vector<chunk_t> saved_chunk;
 
-            saved_chunk.reserve(16);
-            chunk_t *current       = chunk_get_prev_ncnlni(br_open);
-            chunk_t *next_br_close = chunk_get_next(br_close);
-            current = chunk_get_next(current);
-
-            while (current != nullptr)
+            if (options::code_width() > 0)
             {
-               LOG_FMT(LNL1LINE, "%s(%d): zu  kopieren: current->orig_line is %zu, orig_col is %zu, text() is '%s'\n",
-                       __func__, __LINE__, current->orig_line, current->orig_col, current->text());
-               saved_chunk.push_back(*current);
-               chunk_t *the_next = chunk_get_next(current);
+               saved_chunk.reserve(16);
+               chunk_t *current       = chunk_get_prev_ncnlni(br_open);
+               chunk_t *next_br_close = chunk_get_next(br_close);
+               current = chunk_get_next(current);
 
-               if (  the_next == nullptr
-                  || the_next == next_br_close)
+               while (current != nullptr)
                {
-                  break;
+                  LOG_FMT(LNL1LINE, "%s(%d): zu  kopieren: current->orig_line is %zu, orig_col is %zu, text() is '%s'\n",
+                          __func__, __LINE__, current->orig_line, current->orig_col, current->text());
+                  saved_chunk.push_back(*current);
+                  chunk_t *the_next = chunk_get_next(current);
+
+                  if (  the_next == nullptr
+                     || the_next == next_br_close)
+                  {
+                     break;
+                  }
+                  current = the_next;
                }
-               current = the_next;
             }
             tmp = chunk_get_prev_ncnlni(br_open);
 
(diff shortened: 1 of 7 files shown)


tests/config/Issue_2795.cfg
```diff
@@ -0,0 +1 @@
+indent_columns = 4
```
CDanU, line 1: Based on the problem description I'd expected a config containing the `code_width` option. What is tested here? 

Also please keep the naming scheme: It should be named indent_columns-4.cfg so it can be reused for other tests. There is nothing gained from the issue ticket here, you have that info already in the input file name: ` cpp/Issue_2795.cpp`.

src/newlines.cpp
```diff
@@ -2205,6 +2207,30 @@ static void newlines_brace_pair(chunk_t *br_open)
 
          if (is_it_possible)
          {
+            // Issue 2795
```
CDanU, line 2210: This only is usefull when we actually have the option set, so move it into a `if (  options::code_width() > 0)` block

src/newlines.cpp
```diff
@@ -2223,6 +2249,41 @@ static void newlines_brace_pair(chunk_t *br_open)
...
+               // is too long
+               // it is not possible to make an one_liner
+               // because the line would be too long
+               chunk_flags_set(br_open, PCF_NOT_POSSIBLE);
+               // restore the code
+               size_t  count;
+               chunk_t tmp;
+               chunk_t *current = br_open;
+
+               for (count = 0; count < saved_chunk.size(); count++)
```
CDanU, line 2265: Use a range based for loop here.

src/newlines.cpp
```diff
@@ -2223,6 +2249,41 @@ static void newlines_brace_pair(chunk_t *br_open)
...
+               {
+                  tmp = saved_chunk.at(count);
+
+                  if (tmp.orig_line != current->orig_line)
```
CDanU, line 2269: Is it possible that `current` is a nullptr at this place?

src/newlines.cpp
```diff
@@ -2223,6 +2249,41 @@ static void newlines_brace_pair(chunk_t *br_open)
             }
             chunk_flags_set(br_open, PCF_ONE_LINER);         // set the one liner flag if needed
             chunk_flags_set(br_close, PCF_ONE_LINER);
+
+            if (  options::code_width() > 0
+               && br_close->column > options::code_width())
```
CDanU, line 2254: What happens with cases where the brace close is not the last element? Off the top of my head I can only think of classes/structs and lambdas where the braces are followed by a semi colon.

tests/config/Issue_2795.cfg
guy-maurel: Thanks for the comments

src/newlines.cpp
guy-maurel: NO!
Only after making some lines to a one liner, the user wanted:
_nl_create_func_def_one_liner = true_
we are able to know if the line is bigger as _code_with_
Here we have:
_code_width  = 140_
We need a copy to be able to restore the code.
We mark the opening brace with _PCF_NOT_POSSIBLE_ to do the job
only once.

src/newlines.cpp
guy-maurel: NO.
The array was fill just before.

src/newlines.cpp
CDanU: Whats the purpose of the vector content if `code_with` is not set? `code_with = 0`

guy-maurel:
Thanks to @CDanU

Biswa96:
With this commit, `chunk_t tmp` variable is defined twice, previously as a pointer. The variable name need to be changed. Here is the compiler warning:

```c
uncrustify/src/newlines.cpp: In function 'void newlines_brace_pair(chunk_t*)':
uncrustify/src/newlines.cpp:2265:24: warning: declaration of 'tmp' shadows a previous local [-Wshadow]
 2265 |                chunk_t tmp;
      |                        ^~~
uncrustify/src/newlines.cpp:2184:16: note: shadowed declaration is here
 2184 |       chunk_t *tmp      = chunk_get_prev_ncnlni(br_open); // Issue #2279
      |                ^~~
```

PR #2825, opened 2020-06-20
cmake: install license and docs

Biswa96:


CMakeLists.txt
```diff
@@ -398,6 +398,15 @@ set(uncrustify_headers
   ${PROJECT_BINARY_DIR}/uncrustify_version.h
 )
 
+set(uncrustify_docs
+  "${PROJECT_SOURCE_DIR}/AUTHORS"
+  "${PROJECT_SOURCE_DIR}/BUGS"
+  "${PROJECT_SOURCE_DIR}/ChangeLog"
+  "${PROJECT_SOURCE_DIR}/COPYING"
+  "${PROJECT_SOURCE_DIR}/HELP"
+  "${PROJECT_SOURCE_DIR}/README.md"
```
CDanU, line 407: @guy-maurel Do we keep those up to date?

guy-maurel:
The files are part of the git data.
Why should we "install" them?

CDanU:
If I remember correctly the install step will create a linux tree structure, binaries at `/usr/bin/x`, docs at`/usr/local/share/doc/x`. Depnding on how you configure it and what rights you have it goes directly into your system, or it will create the structure at your build place. Or somewhere along that way.

We already do that for the `etc/` dir, so it shouldn't hurt it those are up to date.

Biswa96:
It is good practice to place license, readme, changelog in documentation in a package. For example, see the [bison][1] package file list in Ubuntu.

[1]: https://packages.ubuntu.com/amd64/bison/filelist

PR #2829, opened 2020-06-21
keep emscripten bindings buildable

CDanU:
fake away emscripten code with defines when a normal uncrustify build is done

### DIFF
--- a/src/uncrustify_emscripten.cpp
+++ b/src/uncrustify_emscripten.cpp
@@ -44,7 +44,9 @@
  *   --parsed, -p  ( use debug() )
  */
 
-#ifdef EMSCRIPTEN
+
+#if defined (__linux__)
+
 
 #include "keywords.h"
 #include "log_levels.h"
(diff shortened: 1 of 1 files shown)


gmaurel:
@CDanU Please have a look to https://travis-ci.org/github/uncrustify/uncrustify/jobs/700662475

CDanU:
On my gcc version 10.1.0 everything worked.

I know we had this discussion a few years ago: 
Ubuntu 16.04 is going end of support soon, 18.04 is released as is 20.04 too. The current `g++ (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0` is old, starting with v8 gcc (included in 18.04) is having good c++17 support. 
What exactly is blocking us from upgrading?

gmaurel:
Right. But we should try to keep old version running.
Have a look to 
https://github.com/uncrustify/uncrustify/pull/2786#issuecomment-629759930
from @mihaipopescu

mihaipopescu:
Sorry, I was wrong. I think we can drop 14.04 Trusty completely. It's too old and discontinued.

For Xenial 16.04 LTS I agree on dropping even if it's going to be supported but if that causes big problems on our side we can just remove it and wait to see if people start complaining about it.

So let's follow the main Ubuntu dev streams these days, only 18.04 and 20.04 LTS.

PR #2832, opened 2020-06-23
chunk_t tmp variable is defined twice: change the name

guy-maurel:
Thanks to @Biswa96
Ref. #2795


Biswa96:
A meaningful name would be more readable :)

PR #2838, opened 2020-07-02
use indent_cpp_lambda_only_once lambda expressions in parameters getting doubly indented

nivekkagicom:
Multi-line lambda's provided as a parameter get doubly indented. The following code should be formatted as shown:
```
			deleteSharedLicenseFile_([=](bool const inSuccess, MVError const& inError){
				if (inSuccess || inError.isUserCancelledError()) {
					return inCompletionHandler(inSuccess, inError);
				}
				inCompletionHandler(false, inError);
			});

```
Without indent_cpp_lambda_only_once being specified (the default is false) the code is formatted as follows:
```
			deleteSharedLicenseFile_([=](bool const inSuccess, MVError const& inError){
					if (inSuccess || inError.isUserCancelledError()) {
						return inCompletionHandler(inSuccess, inError);
					}
					inCompletionHandler(false, inError);
				});

```
When the indent_cpp_lambda_only_once is specified as true then the code is formatted as shown in the first code section.

### DIFF
--- a/src/remove_extra_returns.cpp
+++ b/src/remove_extra_returns.cpp
@@ -31,6 +31,8 @@ void remove_extra_returns(void)
          // look for a closing brace
          bool    remove_it      = false;
          chunk_t *closing_brace = chunk_get_next_type(pc, CT_BRACE_CLOSE, 1);
+         LOG_FMT(LRMRETURN, "%s(%d): on orig_line %zu, level is %zu\n",
+                 __func__, __LINE__, pc->orig_line, pc->level);
 
          if (closing_brace != nullptr)
          {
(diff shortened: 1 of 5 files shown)


tests/input/cpp/lambda4.cpp
```diff
@@ -0,0 +1,25 @@
...
+inCompletionHandler(false, inError);
+});
+auto const x = [=](bool const inSuccess, MVError const& inError){
+if (inSuccess || inError.isUserCancelledError()) {
+return inCompletionHandler(inSuccess, inError);
+}
+inCompletionHandler(false, inError);
+};
+}
+}
```
CDanU, line 25: Please reformat this input file with an empty config, this way I can see what your option is doing without all the stuff that is happening by default.

PR #2840, opened 2020-07-03
Don't define away __attribute__ on GCC and Clang on Windows

mstorsjo:
GCC (and Clang acting in GCC mode, with `__GNUC__` defined) can handle `__attribute__` just fine.

This fixes building with libc++ - otherwise, the define no-ops out cruicial attributes within libc++ headers.

### DIFF
--- a/src/logger.h
+++ b/src/logger.h
@@ -74,6 +74,16 @@ void log_set_mask(const log_mask_t &mask);
 void log_get_mask(log_mask_t &mask);
 
 
+#ifdef __MINGW_PRINTF_FORMAT
+// On MinGW, the printf functions can be provided by a number of different
+// implementations, with different format string support. Annontate log_fmt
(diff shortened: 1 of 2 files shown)


CDanU:
Add a comment into the source about the 
> otherwise, the define no-ops out cruicial attributes within libc++ headers

mstorsjo:
> Add a comment into the source about the
> 
> > otherwise, the define no-ops out cruicial attributes within libc++ headers

Sure, I added a comment explaining in more detail why they're essential.

Biswa96:
This adds many format specifier warnings while compiling with gcc but fixes the build issue with clang.

mstorsjo:
> This adds many format specifier warnings while compiling with gcc but fixes the build issue with clang.

I made a PR that silences these warnings in #2846.

mstorsjo:
Appended the commit from #2846 here and closed that one.

PR #2842, opened 2020-07-04
Refactor/define vars

CDanU:
refactor defines

- most are just plain const vars
- some span multiple files/fucntions so they are moved to uncrustify_limits.h
- others that are used in a single function and are moved back as const vars to
  those from uncrustify_limits.h


gmaurel:
@CDanU can you please have a look and resolve the conflict?

PR #2847, opened 2020-07-05
output: fix spaces after * when wrapping comments around

### LINKED ISSUE #2835: Missing space when wrapping C-style comments
[//]: # 
[//]: # 
[//]: # 
[//]: # 
[//]: # "      or"

Hi, I encountered a bug in uncrustify's comment wrapping. Here are my settings and how to reproduce:

Settings:
```ini
cmt_width = 30
cmt_star_cont = true
cmt_sp_after_star_cont = 1
```
Unformatted source:
```cpp
/* Short
 * Very long statement, needs wrapping */
```
Formatted source:
```cpp
/* Short
 * Very long statement, needs
 *wrapping */
```
Note the missing space between `*` and `wrapping`.

Also, debug output of uncrustify:
```
# Uncrustify_d-0.71.0-21-d4da2b0f
cmt_width                       = 30
cmt_star_cont                   = true
cmt_sp_after_star_cont          = 1
# option(s) with 'not default' value: 3
#
# -=====-
# number of loops               = 0
# -=====-
# language                      = CPP
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>      COMMENT_MULTI|      COMMENT_WHOLE|     PARENT_NOT_SET[  1/  1/ 42/  0][0/0/0][          0][1-0] /* Short␤ * Very long statement, needs wrapping */
#   2>            NEWLINE|               NONE|     PARENT_NOT_SET[ 42/ 42/  1/  0][0/0/0][          0][1-0]
# -=====-
```

Something I noticed: the bug happens only if the first line is short but the second is long.

ammen99:
Fixes #2835

I hope I did everything right, please tell me if I should change anything.


CDanU:
Looks good so far but we need way more testing:

- make test different values for 
  - cmt_width
  - cmt_sp_after_star_cont
- the code that was changed also deals with `//` comments so add those too

ammen99:
> * the code that was changed also deals with `//` comments so add those too

You're right, this breaks :/
Now the question is: how do I detect whether we are inside `//` comments or inside `/**/` comments in the code? The check you wrote where we compare to ` * ` doesn't really work (just tried it on a `//` comment, it is still true).

CDanU:
I did write that check before realizing that the cont_text can be just `\ *` as well as `\ * \ ` try checking if the first non whitespace char in `cont_text` is a star.

ammen99:
I tried adding a few more tests, and they seem to work (I also had to do a few other changes). Now there is a single test failing: 30211. The problematic line is the following:

https://github.com/uncrustify/uncrustify/blob/master/tests/input/cpp/indent_comment_align_thresh.cpp#L103

With the expected result being:
https://github.com/uncrustify/uncrustify/blob/master/tests/expected/cpp/30211-indent_comment_align_thresh.cpp#L103

Now, I am not sure why the space exists there in the expected output. Which option is responsible for that?

CDanU:
Heres What I get with your changeset:
```diff
diff --git 1/home/dan/_Files/devel/eclipse/uncrustify/tests/expected/cpp/30211-indent_comment_align_thresh.cpp 2/home/dan/_Files/devel/eclipse/uncrustify/tests/results/cpp/30211-indent_comment_align_thresh.cpp
index bc25f6b43..e0e8c7265 100644
--- 1/home/dan/_Files/devel/eclipse/uncrustify/tests/expected/cpp/30211-indent_comment_align_thresh.cpp
+++ 2/home/dan/_Files/devel/eclipse/uncrustify/tests/results/cpp/30211-indent_comment_align_thresh.cpp
@@ -100,7 +100,7 @@ class FooClass
// bar trailing comment ...
// Baz trailing comment end
/*! Baz description begin
-         * Baz description ...
+         *Baz description ...
* Baz description end */
baz(a,
b); /* Baz trailing comment begin
MISMATCH: cpp:30211
diff --git 1/home/dan/_Files/devel/eclipse/uncrustify/tests/expected/cpp/30212-indent_comment_align_thresh.cpp 2/home/dan/_Files/devel/eclipse/uncrustify/tests/results/cpp/30212-indent_comment_align_thresh.cpp
index abddd42e7..c9594a741 100644
--- 1/home/dan/_Files/devel/eclipse/uncrustify/tests/expected/cpp/30212-indent_comment_align_thresh.cpp
+++ 2/home/dan/_Files/devel/eclipse/uncrustify/tests/results/cpp/30212-indent_comment_align_thresh.cpp
@@ -100,7 +100,7 @@ class FooClass
// bar trailing comment ...
// Baz trailing comment end
/*! Baz description begin
-         * Baz description ...
+         *Baz description ...
* Baz description end */
baz(a,
b); /* Baz trailing comment begin
MISMATCH: cpp:30212
0 / 2 tests passed
```

159:30211  space_indent_class-t_columns-4.cfg   cpp/indent_comment_align_thresh.cpp
```
```



160:30212  indent_comment_align_thresh_2.cfg    cpp/indent_comment_align_thresh.cpp
```
indent_col1_comment             = true
```

Out of those two the only relevant one is `indent_col1_comment` and maybe `indent_class`

ammen99:
@CDanU Sure, I will fix formatting, I wanted to make sure that all tests are correct first.

> It is however more likely what happens when the `options::cmt_sp_after_star_cont` option is not set, therefore `0`, note that `cmt_width` is also set to `0` for those tests. So the problems occur with the new chunks that you made for the lines below line 1792.

Sorry, I don't understand what you mean. Is the test that breaks with the PR actually wrong?

CDanU:
The tests are ok, the pr is breaking them because it removes spaces.

src/output.cpp
```diff
@@ -1787,7 +1797,8 @@ static void output_comment_multi(chunk_t *pc)
                      add_text(cmt.cont_text);
                      // The number of spaces to insert after the star on subsequent comment lines.
                      log_rule_B("cmt_sp_after_star_cont");
-                     output_to_column(ccol + options::cmt_sp_after_star_cont(), false);
+                     const int first_char_col = ccol + options::cmt_sp_after_star_cont();
+                     output_to_column(first_char_col, false);
```
CDanU, line 1801: The problems are most likely caused by this,

src/output.cpp
```diff
@@ -1822,7 +1833,10 @@ static void output_comment_multi(chunk_t *pc)
                         && (line[0] == '*')
                         && unc_isalnum(line[1]))
                      {
-                        line.insert(1, ' ');
+                        for (size_t i = 0; i < options::cmt_sp_after_star_cont(); i++)
+                        {
+                           line.insert(1, ' ');
+                        }
```
CDanU, line 1839: or this.

ammen99:
@CDanU Sorry for the late reply, I had quite a lot of things to do.

To clarify my question: I can see that my PR makes uncrustify not add the spaces to the tests. However, I do not understand: why do we need the spaces there? There seems to be no option which mandates a space after the `*`, because `sp_after_cont`  is `0`, so to me it looks like the **existing** test is just wrong.

CDanU:
No problem, the code is not running away :)

Now that you've rephrased the question it suddenly clicked here. Sorry I was a bit too dense. You are right there is nothing in place to enforce that space. We just forcefully always added one, while that is for almost all cases what the user would desire I agree that this should be configurable.

To not break peoples configs out there we should default `cmt_sp_after_star_cont` to 1.

guy-maurel:
what is the state just now?

ammen99:
Unfortunately, there are still broken tests when I change the defaults. I have tried to verify them manually and saw that indeed, this PR changes behavior in unwanted ways (unfortunately). I haven't had much time to invest though, so I cannot say when I'll be able to debug it.

ammen99:
I am not able to fix all the tests, and I have lost motivation to work further, sorry. Closing this PR. If anyone is willing to continue working on the issue, feel free to take over the changes in this PR.

PR #2858, opened 2020-07-13
Fix #2817: Allow literals to follow the D template instantiation symbol '!'

### LINKED ISSUE #2817: [D] [dlang] Space inserted before `!` when used as template parameter
Version: Uncrustify_d-0.71.0_f

For this code:

```d
auto x = abc!("def");
auto x = abc!"def";
auto x = abc!(y);
auto x = abc!y;
auto x = abc!(1);
auto x = abc!1;
```

uncrustify with all default produces:

```d
auto x = abc!("def");
auto x = abc !"def";
auto x = abc!(y);
auto x = abc!y;
auto x = abc!(1);
auto x = abc !1;
```

i.e. inserts a space in two instances. It should insert no spaces.

andralex:
D allows a template instantiation with a literal to omit parentheses.


andralex:
Do I need to format the PR title a specific way?

CDanU:
The title is ok, we don;t have specific title formatting rules. The tests are a bit sluggish. While it is easy to piggy back on an existing test, the ones you attached to are already using a way to big config, mixing and matching stuff without any goal. 

Your test code apparently does not even need any options set at all. Make a new test that uses the `empty.cfg`, see: 
https://github.com/uncrustify/uncrustify/blob/master/tests/d.test 

There we also don't have conventions for numbering / naming, except that the files need to be in each specific sub dir `cfg` | `input/<lang>/<filename>` | `expected/<lang>/<nr>-<filename>`

andralex:
OK so should I define one new input file and one new expected file?

CDanU:
Yes.

andralex:
@CDanU you got it.

andralex:
Bit confused about the status - e.g. in my local branch and in the symmetry branch the file tests/d.test has the proper vertical alignment, but not in the pull request. What gives?

andralex:
@CDanU okay, here you go. Thanks!

PR #2860, opened 2020-07-14
byref ampersand (&) may follow either a type or a qualifier

nivekkagicom:
In C++ a reference to a const type (T) may be either `const T&` or `T const&` and uncrustify should treat the latter as a CT_BYREF not a CT_ARITH.


gmaurel:
I am missing some test data.
Please, provide a test with expected output

PR #2884, opened 2020-08-03
Support { } in delctypes embedded within templates

parkerjs:
- Modified the check_template() function to account for braces that may be
  embedded within a decltype statement that itself is within a template
  argument list. Without this modification, check_template() incorrectly
  interprets the angle brackets within a template as comparisons

- Moved existing code already used to detect braced initializer statements
  into a separate function so that check_template() may leverage this
  detection code as well

- Added new flag 'PCF_IN_DECLTYPE' to indicate chunk within cpp decltype
  statement

- Made modifications to mark chunks belonging to decltype statements

- Updated cli test scripts and expected output to reflect additions to
  pcf_flag_e bitfields.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -16,6 +16,7 @@
 #include "combine_tools.h"
 #include "ChunkStack.h"
 #include "error_types.h"
+#include "flag_braced_init_list.h"
 #include "flag_parens.h"
 #include "lang_pawn.h"
 #include "language_tools.h"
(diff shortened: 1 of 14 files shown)


parkerjs:
I tried to change as little as possible to add support for the referenced subject. Hopefully I followed established protocols/precedents, but feel free to let me know if I should do things differently.

src/combine.cpp
```diff
@@ -1155,72 +1155,28 @@ void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next)
...
-         log_pcf_flags(LFCNR, pc->flags);
-         auto brace_open = chunk_get_next_ncnl(pc);
+         tmp = chunk_get_next_ncnl(brace_close);
 
-         if (  chunk_is_token(brace_open, CT_BRACE_OPEN)
-            && (  get_chunk_parent_type(brace_open) == CT_NONE
-               || get_chunk_parent_type(brace_open) == CT_ASSIGN
-               || get_chunk_parent_type(brace_open) == CT_RETURN
-               || get_chunk_parent_type(brace_open) == CT_BRACED_INIT_LIST))
+         if (tmp)
```
gmaurel, line 1169: Please use:
if (tmp != nullptr)

src/combine.cpp
```diff
@@ -3832,3 +3788,65 @@ static void handle_java_assert(chunk_t *pc)
       }
    }
 }
+
+
+/**
+ * Detect a cpp braced init list
+ */
+bool detect_cpp_braced_init_list(chunk_t *pc, chunk_t *next)
```
gmaurel, line 3796: Please put the function in extra files (.cpp, .h)
Use an #include ...
Modify the file CMakeLists.txt, to take care

src/combine.h
```diff
@@ -74,4 +74,10 @@ chunk_t *set_paren_parent(chunk_t *start, c_token_t parent);
 void do_symbol_check(chunk_t *prev, chunk_t *pc, chunk_t *next);
 
 
+/**
+ * Detect a cpp braced init list
+ */
+bool detect_cpp_braced_init_list(chunk_t *pc, chunk_t *next);
```
gmaurel, line 80: same as above

src/tokenize_cleanup.cpp
```diff
@@ -92,6 +92,12 @@ static void mark_selectors_in_property_with_open_paren(chunk_t *open_paren);
 static void mark_attributes_in_property_with_open_paren(chunk_t *open_paren);
 
 
+/**
+ * Mark all chunks within the scope of a decltype() statement
+ */
+static void mark_cpp_decltype(chunk_t *pc);
```
gmaurel, line 98: same as above

src/tokenize_cleanup.cpp
```diff
@@ -1189,8 +1204,8 @@ static void check_template(chunk_t *start, bool in_type_cast)
    {
       /*
        * We may have something like "a< ... >", which is a template where
-       * '...' may consist of anything except braces {}, a semicolon, and
-       * unbalanced parens.
+       * '...' may consist of anything except braces {} (unless perhaps
+       * embedded within a decltype), a semicolon, and unbalanced parens.
```
gmaurel, line 1208: put some more example(s) here

src/tokenize_cleanup.cpp
```diff
@@ -1219,13 +1234,37 @@ static void check_template(chunk_t *start, bool in_type_cast)
...
       {
          if (  (chunk_is_token(pc, CT_SEMICOLON) && hit_semicolon)
-            || chunk_is_token(pc, CT_BRACE_OPEN)
-            || chunk_is_token(pc, CT_BRACE_CLOSE)
             || chunk_is_token(pc, CT_SQUARE_CLOSE))
          {
             break;
          }
 
+         if (pc->type == CT_DECLTYPE)
```
gmaurel, line 1242: Use the function chunk_is_token(pc, CT_DECLTYPE)

src/tokenize_cleanup.cpp
```diff
@@ -1219,13 +1234,37 @@ static void check_template(chunk_t *start, bool in_type_cast)
...
+         {
+            mark_cpp_decltype(pc);
+         }
+
+         if (chunk_is_token(pc, CT_BRACE_OPEN))
```
gmaurel, line 1247: ... such as here

src/tokenize_cleanup.cpp
```diff
@@ -1567,3 +1623,30 @@ static void mark_attributes_in_property_with_open_paren(chunk_t *open_paren)
       tmp = tmp->next;
    }
 }
+
+
+void mark_cpp_decltype(chunk_t *pc)
```
gmaurel, line 1628: same as above

tests/cli/output/p.txt
```diff
@@ -8,72 +8,72 @@ newlines                        = crlf
 # language                      = CPP
```
gmaurel, line 8: The file p.txt should NOT be part of the PR

src/combine.cpp
parkerjs: What shall I name the new files? Any suggestions? Thanks

src/combine.h
parkerjs: Same as previous remark.

src/tokenize_cleanup.cpp
parkerjs: Same as previous remark. I ask because you probably have a better idea as to what the new files should be called.

tests/cli/output/p.txt
parkerjs: Ok, so I updated this file because I added PCF_IN_DECLTYPE to the bitfields in pcf_flag_e. If I don't update this, I fail the CLI options test. Please advise.

src/combine.cpp
parkerjs: I moved this function into a new file, 'flag_braced_init_list.cpp'

src/tokenize_cleanup.cpp
parkerjs: I moved this function into a new file, 'flag_decltype.cpp'

tests/cli/output/p.txt
guy-maurel: run uncrustify/tests/cli/test_cli_options.py
to see what has changed, and review the output/xyz.txt file

parkerjs: I've reviewed the results of the diff, and I think the reason for which I updated the p.txt output file still stands. I've inserted the PCF_IN_DECLTYPE flag at bit 16 (17th LSB), and as such the "Flag" field must be updated to reflect this. For example, if you take line 11 of the existing p.txt output file, you'll see that the value under the Flag column is 1000e0001; however, that value is before I inserted the new flag. With the addition of PCF_IN_DECLTYPE at the 16th bit position, the upper 48 bits must be left-shifted by one (the most significant bit is discarded, but it's not used anyway). This means that 1000e0001 must be updated to the value 2001c0001 for line 11.
```
           1    0    0    0    e    0    0    0    1
Before: 0001 0000 0000 0000 1110 0000 0000 0000 0001
           2    0    0    1    c    0    0    0    1
After:  0010 0000 0000 0001 1100 0000 0000 0000 0001
                               ^
                               insertion here at 16th bit position (17th LSB)
```

I've updated the Flag fields where applicable - some Flag fields did not change since no bits higher than the 16th LSB were set in the first place. I hope this makes sense.

PR #2908, opened 2020-09-10
Fix mishandling of some class one-liners

### LINKED ISSUE #2907: nl_after_func_class_proto_group(?) not working properly
It appears that `nl_after_func_class_proto_group` *may* not actually recognize constructors; at least, not if they have inline definitions. However, it's also possible this is caused by one-liners not being preserved.

Input code:
```c++
template< typename Enum > class Flags
{
  public:
    constexpr Flags() : value{ 0 } {}
    constexpr Flags( Enum f ) : value( static_cast< value_t >( f ) ) {}
    constexpr Flags( Flags const& ) = default;
}
```

Input config:
```
nl_class_leave_one_liners = true
nl_before_opening_brace_func_class_def = force
nl_after_func_class_proto_group = 2
```

No change is expected. Actual output is:
```c++
template< typename Enum > class Flags
{
  public:
    constexpr Flags() : value{ 0 }
    {}

    constexpr Flags( Enum f ) : value( static_cast< value_t >( f ) )
    {}

    constexpr Flags( Flags const& ) = default;
}
```

I can provide more information if needed, however I'm hoping to chase this down myself; this report is as much as anything to generate an issue number to use in a new test.

mwoehlke:
Teach `nl_before_opening_brace_func_class_def` to respect `nl_class_leave_one_liners`.

Also, tweak `mark_cpp_constructor` to also set the parent type on any qualifiers ('`virtual`', '`constexpr`') preceding a ctor/dtor prototype or definition. This fixes `nl_class_leave_one_liner_groups` not being applied to constexpr constructors.
    
Fixes #2907.

### DIFF
--- a/src/combine_fix_mark.cpp
+++ b/src/combine_fix_mark.cpp
@@ -1279,6 +1279,19 @@ void mark_cpp_constructor(chunk_t *pc)
       LOG_FMT(LFCN, "%s(%d):  Marked '%s' as FUNC_CLASS_PROTO on orig_line %zu, orig_col %zu\n",
               __func__, __LINE__, pc->text(), pc->orig_line, pc->orig_col);
    }
+   tmp = chunk_get_prev_ncnlni(pc); // Issue #2907
+
+   if (chunk_is_token(tmp, CT_DESTRUCTOR))
+   {
+      set_chunk_parent(tmp, pc->type);
+      tmp = chunk_get_prev_ncnlni(tmp);
+   }
+
+   while (chunk_is_token(tmp, CT_QUALIFIER))
+   {
+      set_chunk_parent(tmp, pc->type);
+      tmp = chunk_get_prev_ncnlni(tmp);
+   }
 } // mark_cpp_constructor
 
 
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -3904,11 +3904,19 @@ void newlines_cleanup_braces(bool first)
 
          case CT_FUNC_CLASS_DEF:                             // Issue #2343
          {
-            log_rule_B("nl_before_opening_brace_func_class_def");
-
-            if (options::nl_before_opening_brace_func_class_def() != IARF_IGNORE)
+            if (!one_liner_nl_ok(pc))
             {
-               newline_iarf_pair(chunk_get_prev(pc), pc, options::nl_before_opening_brace_func_class_def());
+               LOG_FMT(LNL1LINE, "a new line may NOT be added\n");
+               // no change - preserve one liner body
+            }
+            else
+            {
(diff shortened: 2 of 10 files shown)


mwoehlke:
Well, this is an interesting failure (30117):
```
class Capteur_CO2 : public Capteur {
    public:
        Capteur_CO2() :
            un_membre_en_plus ( 0 ) {}
};
```

...seems that's considered a one-liner? That... doesn't seem right?

gmaurel (approved):

gmaurel (commented):
Some problems at 31-test

mwoehlke:
> Some problems at 31-test

...as already noted in [my prior comment](https://github.com/uncrustify/uncrustify/pull/2908#issuecomment-689919272) (and in somewhat more detail [here](https://github.com/uncrustify/uncrustify/issues/2907#issuecomment-689920718)). I'm not sure what's the correct behavior here; I'd appreciate your thoughts?

guy-maurel:
I have made a test.
Merge your changes, updated the p.txt, 31.txt, 66.txt
I now get:
```.txt
Uncrustify_d-0.71.0-157-a784082a-dirty
Parsing: mwoehlke.cpp as language CPP
```
```.cpp
template< typename Enum > class Flags
{
public:
constexpr Flags() : value{ 0 } {}
constexpr Flags( Enum f ) : value( static_cast< value_t >( f ) ) {}
constexpr Flags( Flags const& ) = default;

}

```

Is it right?

mwoehlke:
Arf, I missed the CLI / sources-format issues. Those are probably easy; I'll try to take a look later. The one that's vexing me is cpp:30117. I could bash that into passing in a couple ways, but what is the *correct* behavior for that? Is the change in behavior there desired or not?

mwoehlke:
I've updated the PR so that this:
```c++
  class_ctor() :
    member{...} {}
```
...is **not** considered a one-liner for `nl_before_opening_brace_func_class_def`. (But I didn't change `one_liner_nl_ok`.)

Please consider if this seems like a reasonable approach.

I also fixed the CLI tests and formatting; hopefully everything is passing now.

mwoehlke:
> Please consider if this seems like a reasonable approach.

In fact, I am inclined (partly, after noticing that's the way my own code is formatted :stuck_out_tongue:) to *not* take that approach, and to indeed treat that as a one-liner. That would imply changing the expected output of cpp:30117, however.

mwoehlke:
> In fact, I am inclined to *not* take that approach, and to indeed treat that as a one-liner. That would imply changing the expected output of cpp:30117, however.

I've now pushed that version instead. Please take note of the changes to cpp:30117 and consider whether or not they seem reasonable.

PR #2915, opened 2020-09-16
Add pos_shift

mwoehlke:
Add `CT_SHIFT`, making shift operators a different token type from `CT_ARITH`. Add `pos_shift` allowing shift operators to be positioned differently from other arithmetic operators.


mwoehlke:
Should this be added to `forUncrustifySources.cfg`? We don't currently have `pos_arith` (or indeed, very many `pos_*` options) in there, and also AFAICT we don't have any instances of multi-line statements broken at a shift. (We also use almost no streaming operators, which is where such code usually occurs.)

gmaurel (approved):

PR #2920, opened 2020-09-19
Fix formatting of call operator after init-list

mwoehlke:
Correctly mark parentheses following a braced initializer list as a function call. (Specifically, this is invoking `operator()` of the just-constructed object.) Also, check for this when applying spaces so that we use `sp_func_call_paren[_empty]` rather than failing to match any rule and falling back to the default of always adding a space.


mwoehlke:
Without this change, the output of the added test is:
```c++
auto x = foo{0} ();
```

(Note the incorrectly added space between `}` and `()`.)

gmaurel (approved):

PR #2973, opened 2020-11-05
Define the new command option

gmaurel:
--tracking_space FILE : Prepare tracking informations for debugging.


gmaurel:
make a new one later

PR #3005, opened 2020-12-08
Fix newlines before namespaces

parkerjs:
This commit addresses an issue in which the can_increase_nl() function
fails to allow newlines to be added before a namespace, if say, the
preceding statment does not enforce a subsequent newline and the
closing brace of the namespace is the last token in the file. The
fixes imposed by this commit are further demonstrated by applying
the configuration specified by 'forUncrustifySources.cfg' to the
uncrustify source code itself. In 'uncrustify_limits.h', 'namespace
uncrustify' is preceded by a '#pragma once' statement, but there
should be a newline before the namespace according to the
configuration specified in 'forUncrustifySources.cfg'.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -394,6 +394,18 @@ static bool can_increase_nl(chunk_t *nl)
       }
    }
 
+   if (chunk_is_token(prev, CT_BRACE_CLOSE))
+   {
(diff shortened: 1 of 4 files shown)


gmaurel (approved):
Fine catch!
Thanks

PR #3006, opened 2020-12-09
Fix issues with nl_inside_empty_func

parkerjs:
This commit addresses issues that arise when 'nl_inside_empty_func'
is used in conjunction with 'eat_blanks_before_close_brace' or
'eat_blanks_after_open_brace'. The problem occurs when a line is
split according to code width constraints imposed by the
'code_width' option. It turns out that the 'nl_max_blank_in_func'
was also ignored for similar reasons, and as such an additional
call to newlines_functions_remove_extra_blank_lines() is made
after a call to do_code_width() *if* a change results from
imposing line width constraints.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -365,7 +365,7 @@ static bool can_increase_nl(chunk_t *nl)
 
    if (chunk_is_token(next, CT_BRACE_CLOSE))
    {
-      if (  options::nl_inside_namespace()
+      if (  options::nl_inside_namespace() > 0
          && get_chunk_parent_type(next) == CT_NAMESPACE)
       {
          log_rule_B("nl_inside_namespace");
@@ -408,7 +408,7 @@ static bool can_increase_nl(chunk_t *nl)
 
    if (chunk_is_token(prev, CT_BRACE_OPEN))
    {
-      if (  options::nl_inside_namespace()
+      if (  options::nl_inside_namespace() > 0
          && get_chunk_parent_type(prev) == CT_NAMESPACE)
       {
          log_rule_B("nl_inside_namespace");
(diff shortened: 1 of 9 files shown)


src/newlines.cpp
```diff
@@ -2459,9 +2459,18 @@ static void newlines_brace_pair(chunk_t *br_open)
    {
       if (chunk_is_newline(next))
       {
+         log_rule_B("nl_inside_empty_func");
          log_rule_B("nl_inside_namespace");
 
-         if (options::nl_inside_namespace() && get_chunk_parent_type(br_open) == CT_NAMESPACE)
+         if (  options::nl_inside_empty_func()
```
guy-maurel, line 2465: Please change to:
if (  options::nl_inside_empty_func() > 0

src/newlines.cpp
```diff
@@ -2459,9 +2459,18 @@ static void newlines_brace_pair(chunk_t *br_open)
...
+            && chunk_is_token(chunk_get_next_ncnl(br_open), CT_BRACE_CLOSE)
+            && (  get_chunk_parent_type(br_open) == CT_FUNC_CLASS_DEF
+               || get_chunk_parent_type(br_open) == CT_FUNC_DEF))
+         {
+            blank_line_set(next, options::nl_inside_empty_func);
+         }
+         else if (  options::nl_inside_namespace()
```
guy-maurel, line 2472: and here to:
else if (  options::nl_inside_namespace() > 0

src/newlines.cpp
```diff
@@ -4105,8 +4114,17 @@ void newlines_cleanup_braces(bool first)
             if (chunk_is_newline(prev))
             {
                log_rule_B("nl_inside_namespace");
+               log_rule_B("nl_inside_empty_func");
 
-               if (options::nl_inside_namespace() && get_chunk_parent_type(pc) == CT_NAMESPACE)
+               if (  options::nl_inside_empty_func()
```
guy-maurel, line 4119: also here

src/newlines.cpp
```diff
@@ -4105,8 +4114,17 @@ void newlines_cleanup_braces(bool first)
...
+                  && chunk_is_token(chunk_get_prev_ncnlni(pc), CT_BRACE_OPEN)
+                  && (  get_chunk_parent_type(pc) == CT_FUNC_CLASS_DEF
+                     || get_chunk_parent_type(pc) == CT_FUNC_DEF))
+               {
+                  blank_line_set(prev, options::nl_inside_empty_func);
+               }
+               else if (  options::nl_inside_namespace()
```
guy-maurel, line 4126: and here

guy-maurel (changes requested):
have a look to my comments

gmaurel (approved):

PR #3009, opened 2020-12-10
Fix indent after return within lambda expressions

parkerjs:
This commit fixes an issue in which line-wrapped return statements
are in some cases not properly aligned after the return keyword.
This happens in some lambda expressions, but not all. Below is an
example illustrating a case in which the alignment fails. It
turns out that lambdas (legitimately) embedded within a set of
parens were previously excluded from the alignment.

With 'indent_off_after_return=false', before this commit:

```
template<typename U>
U *
find(const std::string &name = "") const
{
	return find<U>([&name] (auto *pComposite)
	{
		return name.empty() ||
		pComposite->getName() == name;
	});
}
```

With 'indent_off_after_return=false', after this commit:

```
template<typename U>
U *
find(const std::string &name = "") const
{
	return find<U>([&name] (auto *pComposite)
	{
		return name.empty() ||
		       pComposite->getName() == name;
	});
}
```


gmaurel (changes requested):
why do you remove the cpp test 30015?
If it is no more necessary, please remove the files
tests/input- and tests/expected- data are missing

parkerjs:
> why do you remove the cpp test 30015?

I'm not sure how I removed this, but it was a mistake. I've added this back, and I'll try to be more careful moving forward.

gmaurel (approved):
fine

PR #3014, opened 2020-12-12
Fix eat_blanks_after_open_brace for line-wrapping

parkerjs:
This commit fixes issues where the eat_blanks_after_open_brace
option, if true, would ordinarily remove blanks following an
open brace - if not for line-wrapping imposed by code-width
constraints (non-zero code_width is specified). After this
commit, blanks are removed (if eat_blanks_after_open_brace=true)
regardless of whether or not any line-wrapping actually occurs.


gmaurel (approved):

micheleCTDE:
This introduced a regression described at issue #3061.

PR #3042, opened 2021-01-02
Add feature to align doxygen/javadoc comment tags

parkerjs:
This commit adds the ability to align doxygen javadoc-style comment
tags. To accomplish this goal, the following two options have been
added:

- cmt_align_doxygen_javadoc_tags
- cmt_sp_before_doxygen_javadoc_tags

The following example demonstrates this new feature, before
and after application of this new feature:

Before:

```
/**
 * @param fromFile file from which a piece is being moved
 * @param fromRank rank from which a piece is being moved
 * @param toFile file to which a piece is being moved
 * @param toRank rank to which a piece is being moved
 * @return true if the move is valid, otherwise false
 * @since 1.0
 */
```

After:

```
/**
 * @param  fromFile file from which a piece is being moved
 * @param  fromRank rank from which a piece is being moved
 * @param  toFile   file to which a piece is being moved
 * @param  toRank   rank to which a piece is being moved
 * @return          true if the move is valid, otherwise false
 * @since           1.0
 */
```

If comment width constraints are imposed (cmt_width > 0), it may
be necessary to enable cmt_indent_multi and set cmt_reflow_mode = 2
in order to achieve the desired alignment for line-wrapping.


gmaurel (changes requested):
The part "If comment width constraints..." should be seen in the doc to inform the next users.
Please look for a suitable place.

gmaurel (approved):

PR #3043, opened 2021-01-02
insert closing brace before the preprocessor line

### LINKED ISSUE #3040: Brace incorrectly added after a PP directive
This issue is similar to issue #3033 but I am reporting it separately since it shows up on a different config option. I wouldn't be surprised if the root cause is the same as the aforementioned issue.

uncrustify 0.72.0 + backport of commits 45681e88, 64d93e3.
Files in attachment:
- t1_in.cpp: test file (subset of a real file, it would fail compilation :wink: )

uncrustify should add braces around "case" blocks, but when PP directive are presents, it seems to get confused.

Current output:
```
void main()
{
	switch (opcode)
	{
		case LocaleCompare:
		{
			return Number(localeCompare(s, a0.toString(exec)));

#ifndef KJS_PURE_ECMA
		}

		case Big:
		{
			result = String("<big>" + s + "</big>");
			break;
		}
#endif
	}
}
```

Note how the closing brace of the "case LocaleCompare:" block is added after the PP #ifndef instead than before it.

uncrustify does not report any error, but if you try to format the file a second time with "uncrustify -c test.cfg t1_out.cpp", the following errors are reported, which at least is a pointer that somethgin is wrong.
```
Parsing: t1_out_1st_pass.cpp as language CPP
indent_text(3862): size is 2
indent_text(3864): File: t1_out_1st_pass.cpp, open_line is 2, parent is NONE: Unmatched BRACE_OPEN
indent_text(3862): size is 2
indent_text(3864): File: t1_out_1st_pass.cpp, open_line is 2, parent is NONE: Unmatched BRACE_OPEN
```

[wrong_case_brace.zip](https://github.com/uncrustify/uncrustify/files/5756752/wrong_case_brace.zip)

guy-maurel:
Fixes #3040

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -1330,7 +1330,7 @@ static chunk_t *mod_case_brace_remove(chunk_t *br_open)
 static chunk_t *mod_case_brace_add(chunk_t *cl_colon)
 {
    LOG_FUNC_ENTRY();
-   LOG_FMT(LMCB, "%s(%d): line %zu",
+   LOG_FMT(LMCB, "%s(%d): line %zu\n",
            __func__, __LINE__, cl_colon->orig_line);
 
    chunk_t *pc   = cl_colon;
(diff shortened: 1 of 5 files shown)


micheleCTDE:
See comment in issue [3040](https://github.com/uncrustify/uncrustify/issues/3040#issuecomment-753575582).
This commit introduces a regression.

gmaurel (approved):

micheleCTDE:
Great, thanks Guy!

PR #3048, opened 2021-01-06
rewrite mod_case_brace_add

guy-maurel:
fixes #3040

PR #3050, opened 2021-01-08
rewrite mod_case_brace_add

guy-maurel:
fixes #3048 
only part 1 is fixed.
The less important latter.

### DIFF
--- a/tests/expected/cpp/34327-Issue_3044.cpp
+++ b/tests/expected/cpp/34327-Issue_3044.cpp
@@ -1,0 +1,61 @@
+int main()
+{
+	int af;
+	int A;
+	int B;
+	switch (af)
+	{
+		case 1:
+		{
+			B = 2;
+		}
+
+		case 2:
+		{
+			return 1;
+		}
+
+		case 3:
+		{
+			A = 1;
+			break;
+#ifdef ALL_THE_CASE
+		}
+
+		case 4:
+		{
+			return 2;
+#endif
+#ifdef ALL_THE_CASE
+		}
+
+		case 5:
+		{
+			B = 2;
+#endif
+		}
+
+		case (6):
+		{
+			B=13;
+#ifdef PART_OF_THE_CASE_UNDER
+			A=1;
+#endif
+			break;
+		}
+
+		case (7):
+		{
+#ifdef PART_OF_THE_CASE_ABOVE
+			A=5;
+#endif
+			B=7;
+			break;
+		}
+
+		default:
+		{
+			B= 50;
+		}
+	}
+}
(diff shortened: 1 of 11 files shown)


tests/expected/cpp/34327-Issue_3044.cpp
```diff
@@ -0,0 +1,61 @@
...
+		{
+			return 2;
+#endif
+#ifdef ALL_THE_CASE
+		}
+
+		case 5:
+		{
+			B = 2;
+#endif
```
micheleCTDE, line 35: Althought technically speaking this results in correct code, it definitely looks a bit weird in the way the case is broken up.

micheleCTDE:
Hi Guy, thanks again for your work on this.
Before I go ahead and run my usual tests on my test repo, I wonder what you think of my comment above.
The results in the "tests/expected/cpp/34327-Issue_3044.cpp" file related to "#ifdef ALL_THE_CASE" looks a bit awkward to read, although technically correct. 
Do you think that could be better reworked to provide a result as below? Or is it too problematic to distinguish the different scenarios?

```
	case 3:
		{
			A = 1;
			break;
		}

#ifdef ALL_THE_CASE
		case 4:
		{
			return 2;
		}
#endif

#ifdef ALL_THE_CASE
		case 5:
		{
			B = 2;
		}
#endif
```

micheleCTDE:
Hi Guy,
apologies for the delay, yesterday I was busy and could not test. 
This version introduces a new regression.

Test 1: uncrustify 0.72.0 + backport of commits 45681e88, 64d93e3 + fix for issue 3040 (both commits in PR #3043) 
Test 2: Uncrustify-0.72.0-101-eeeb2372 (this PR)

Files in attachment:
- t1_out_old.cpp: output of test 1
- t1_out_new.cpp: output of test 2

Step to reproduce: run "uncrustify -c test.cfg -f t1_in.cpp -o t1_out_old[/new].cpp" as per test1/2.

Regression: if there is a comment before a "case", the closing brace of the previous "case" is added on the same line as the comment in the code, meaning the brace is commented out. This makes the code failing compilation and also causes uncrustify to throw errors on a second pass.
I have observed this problem in multiple files.

Original code:
```
    case CenteredAutoFit:
        [ ... omissis ]
        // fall through
    case CenteredMaxpect: {
```

Result from this PR code:
```
		case CenteredAutoFit:
		{
                        [ ... omissis ]
			// fall through}

		case CenteredMaxpect:
```

[PR3050_regression.zip](https://github.com/uncrustify/uncrustify/files/5793849/PR3050_regression.zip)

gmaurel:
It is very fine to see your produced comments. Very appreciated!
I now have to work on them. It will take lot of time. Please wait a little bit.

Please make separate issue for each bug you have found.
It is easier for me to operate.

micheleCTDE:
No problem Guy, take your time and when a new PR is ready I will test it through my usual test repo :+1: 
Will create separate reports for future issues!

guy-maurel:
see you latter with a new PR

PR #3056, opened 2021-01-14
rewrite mod_case_brace_add, add  calculate_closing_brace_position

### LINKED ISSUE #3044: Case brace incorrectly added after #endif directive instead of before it
This issue is similar to issue #3040 but relates to the end brace of the case block. The problem happens both with and without the fix proposed in PR #3043.

Test 1: uncrustify 0.72.0 + backport of commits 45681e88, 64d93e3 
Test 2: uncrustify 0.72.0 + backport of commits 45681e88, 64d93e3 + fix for issue 3040 (both commits in PR #3043) 

Files in attachment:

The end brace of the "case AF_INET6:" block should be placed before the #endif directive, but instead it is added after that. I understand uncrustify is not a compiler but as for issue #3040, perhaps we can do something to detect this kind of situation.

Current output extract:
```
#ifdef AF_INET6
		case AF_INET6:
		{
			return 2;
#endif
	}
```

uncrustify does not report any error, but if you try to format the file a second time with "uncrustify -c test.cfg t1_out.cpp", the following errors are reported:
```
Output suffix: .uncrustify
Parsing: t1_out.cpp as language CPP
parse_cleanup(456): pc->orig_line is 20, orig_col is 1, text() is '}', type is BRACE_CLOSE
parse_cleanup(458): (frm.top().type + 1) is UNKNOWN
parse_cleanup(869): Unmatched BRACE_CLOSE
   orig_line is 20, orig_col is 1
parse_cleanup(876): Try the option 'tok_split_gte = true'
```

[wrong_end_case_brace.zip](https://github.com/uncrustify/uncrustify/files/5763823/wrong_end_case_brace.zip)

guy-maurel:
fixes #3044, #3048


micheleCTDE:
Hi Guy, thanks for this PR.
It works good and I have not run into any regression in my sample repository (3200+ files). I think it is ok to merge in.

There is a couple of minor imperfections, but I will file separate issues for those and you can work on them after merging this PR I guess.

Well done!!

gmaurel (approved):

PR #3060, opened 2021-01-18
Fix cmt_insert_func_header for head chunk

parkerjs:
The cmt_insert_func_header option does not appear to properly
add function headers if the function or class occurs at the
very beginning of the source file. This commit attempts to
fix this, and the change to 34200-i1536.cpp demonstates the
fix made by this commit accordingly.


gmaurel (approved):
Fine catch

PR #3066, opened 2021-01-25
Fix bugs related to cmt_align_doxygen_javadoc_tags

parkerjs:
Fixed a potential crash condition and also fixed a few potential
overflow conditions due to signed/unsigned mixing in output.cpp.

Also modified the code to disallow alignment of javadoc/doxygen tags
that do not occur at the beginning of a line (or after the lead star
in a multi-line comment, if present). Updated
doxy-javadoc-alignment.java to demonstrate this behavior.


gmaurel (approved):
Thanks

PR #3069, opened 2021-01-27
a better message if stdin is used

### LINKED ISSUE #3064: [UX] uncrustify behaves oddly on stdin, because it doesn't know the language.
Behavior:

When using uncrustify with input from STDIN it formats the file, but it's formatted differently than when passing the file directly.

Expected:

If language (with -l) is not set, should output clear error that it doesn't know the file type.

Reasoning:

I wanted to pass in files from stdin, uncrustify didn't complain, and formatted things wrong. Instead it's better to tell the user with a clear error message that the language flag has to be set. That way the user doesn't have to guess why the formatting is off.

guy-maurel:
fixes #3064

### DIFF
--- a/src/sorting.h
+++ b/src/sorting.h
@@ -1,10 +1,11 @@
 /**
  * @file sorting.h
- * prototypes for sorting.c
+ * prototypes for sorting.cpp
  *
  * @author  Ben Gardner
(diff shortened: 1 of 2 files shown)


gmaurel (approved):
Thanks for the message

PR #3084, opened 2021-02-02
FreeBSD style has a <TAB> after #define

motminh:
$ man 9 style
...
Put a single tab character between the #define and the macro name.


gmaurel (approved):
Thanks for the PR

PR #3089, opened 2021-02-06
Fix test documentation in readme

micheleCTDE:
Expected test file names are named after the test number and the test
input file, not the config file.
For example in tests/cpp.test there is 
```
30015  ben_008.cfg                          cpp/exception.cpp
```
and the expected file is called "30015-exception.cpp", not "30015-ben_008.cfg"


gmaurel (approved):
Fine catch, thanks

Issue #3061, opened 2021-01-18
Regression in current master: missing line at the end of the file


micheleCTDE:
Files in attachment:
- test.cfg: configuration used for the test. This forces a new line at the end of the file
- dcopclient_no_problem.cpp: test file without problem
- dcopclient_no_problem_0_72.cpp: output of 0.72.0 execution
- dcopclient_no_problem_master.cpp: output of current master execution
- dcopclient_problem.cpp: test file with problem
- dcopclient_problem_0_72.cpp: output of 0.72.0 execution
- dcopclient_problem_master.cpp: output of current master execution

Step to reproduce: 
1. using version 0.72.0:
uncrustify -c test.cfg -f dcopclient_no_problem -o dcopclient_no_problem_0_72.cpp
uncrustify -c test.cfg -f dcopclient_problem -o dcopclient_problem_0_72.cpp

2. using current master version (Uncrustify-0.72.0-103-b0598a72):
uncrustify -c test.cfg -f dcopclient_no_problem.cpp -o dcopclient_no_problem_master.cpp
uncrustify -c test.cfg -f dcopclient_problem.cpp -o dcopclient_problem_master.cpp

The file dcopclient_no_problem.cpp results in the same output with both version of uncrustify.
The file dcopclient_problem.cpp results in different output (1 byte difference): the current version removes the new line at the end of the file. Note this is against the settings in use (nl_end_of_file_min = 2).

I have not checked when this was introduced, but it was quite recent, probably within the last two weeks. I didn't see this issue while testing PR #3048 but I ran into this regression when testing PR #3056.
The problem shows up in multiple files and it does not always (but mostly) happens. It depends on the contents in the file. I guess something gets set while processing the file and this triggers the issue.
Suggest to experiment also with the original file without the last new line and make sure it is added correctly with the attached config file (as it should).

[missing_end_line.zip](https://github.com/uncrustify/uncrustify/files/5828885/missing_end_line.zip)

micheleCTDE:
Issue introduced by commit 313f6eb (PR #3014).
Fix proposed in PR #3095.

micheleCTDE:
PR has been merged, so closing the issue.

PR #3095, opened 2021-02-10
Fix issue 3061

micheleCTDE:
Issue #3061 was introduced by commit 313f6eb7 (PR #3014).

'void newlines_remove_disallowed()' should not eat new line characters
in the following cases:
1. we are at the end of the file
2. there are consecutive new line characters
because that is already dealt with in a different part of the code.

Tests with different combinations of existing and required numbers of nl are included. 4 sets, each set requiring a different number of nl on 3 different files with different original numbers of nl. This tests adding nl, removing nl, leaving nl unchanged at the end of the file.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -3647,14 +3647,19 @@ void newlines_remove_disallowed()
 {
    LOG_FUNC_ENTRY();
 
-   auto *pc = chunk_get_head();
+   chunk_t *pc = chunk_get_head();
+   chunk_t *next;
 
    while ((pc = chunk_get_next_nl(pc)) != nullptr)
    {
       LOG_FMT(LBLANKD, "%s(%d): orig_line is %zu, orig_col is %zu, <Newline>, nl is %zu\n",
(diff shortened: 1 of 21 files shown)


src/newlines.cpp
```diff
@@ -3647,14 +3647,19 @@ void newlines_remove_disallowed()
...
 
    while ((pc = chunk_get_next_nl(pc)) != nullptr)
    {
       LOG_FMT(LBLANKD, "%s(%d): orig_line is %zu, orig_col is %zu, <Newline>, nl is %zu\n",
               __func__, __LINE__, pc->orig_line, pc->orig_col, pc->nl_count);
 
-      if (!can_increase_nl(pc))
+      next = chunk_get_next(pc);
+
+      if (  next
```
gmaurel, line 3660: Please use:
if ( next != nullptr
...

gmaurel (changes requested):

src/newlines.cpp
micheleCTDE: updated as required

gmaurel (approved):
Thanks

PR #3096, opened 2021-02-12
Improve code base

micheleCTDE:
Get rid of 'chunk_del' macro and replace it with a proper function call, making the code easier to read.
All tests passed successfully. In addition to that, testing on a small set of 3200 files yields no difference at all.

### DIFF
--- a/src/chunk_list.cpp
+++ b/src/chunk_list.cpp
@@ -538,10 +538,11 @@ chunk_t *chunk_add_before(const chunk_t *pc_in, chunk_t *ref)
 }
 
 
-void chunk_del_2(chunk_t *pc)
+void chunk_del(chunk_t * &pc)
 {
    g_cl.Pop(pc);
    delete pc;
+   pc = nullptr;
 }
 
 
--- a/src/chunk_list.h
+++ b/src/chunk_list.h
@@ -87,12 +87,7 @@ chunk_t *chunk_add_before(const chunk_t *pc_in, chunk_t *ref);
  *
  * @param pc  chunk to delete
  */
-#define chunk_del(pc)    do { \
-      chunk_del_2((pc));      \
-      (pc) = nullptr;         \
-} while (false)
-
-void chunk_del_2(chunk_t *pc);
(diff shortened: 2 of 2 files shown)


micheleCTDE:
The new commit is the same as before, just removed the dot at the end of the commit line, as per guidelines.

src/chunk_list.cpp
```diff
@@ -538,10 +538,11 @@ chunk_t *chunk_add_before(const chunk_t *pc_in, chunk_t *ref)
 }
 
 
-void chunk_del_2(chunk_t *pc)
+void chunk_del(chunk_t * &pc)
```
gmaurel, line 541: I don't understand _* &_
Could it be?
_void chunk_del(chunk_t pc)_

gmaurel (changes requested):

src/chunk_list.cpp
micheleCTDE: "chunk_t * &pc" means that "pc" is a reference to a pointer to chunk_t. It is the same as any other argument passed by reference, the only difference being that this time we are passing a pointer.

"chunk_del_2()" needs a pointer for its functionality, then the pointer is set to "nullptr" by the macro. By using a reference to a pointer, we can make sure the pointer is set to "nullptr" in the function "chunk_del()" and the change is propagate back to the calling environment without the need to use a macro, an additional scope block and a not-used do-while loop.

micheleCTDE: An alternative solution would be to use a pointer to pointer as in here:
```
void chunk_del(chunk_t **ppc)
{
   g_cl.Pop(*pc);
   delete *pc;
   *pc = nullptr;
}
```
This would implement exactly the same thing, but it would require changing all the points where chunk_del() is called, since we would have to explicitely pass the pointer address. Using the proposed PR does not require all those additional changes, but if you prefer a pointer to pointer I can amend the PR.

mwoehlke-kitware (approved):

gmaurel (approved):

PR #3101, opened 2021-02-18
Fix inconsistent chunk names.

micheleCTDE:
The names of some of the chuck functions is not consistent, specifically those related to non-new lines (except chunk_get_next_nnl() which is fine).

chunk_get_next_nl() is for new lines.
chunk_get_next_nnl() is for non-new lines.
chunk_get_next_nc() is for non-comments.
chunk_get_next_ncnl() is for non-comments, non-new lines. It's name is inconsistent and should be chunk_get_next_ncnnl().

Similar observation holds true for the other functions renamed in this commit.


gmaurel (approved):
Thanks for the changes

PR #3103, opened 2021-02-24
Fix building when SUPER_LOG is #defined.

micheleCTDE:
When building current uncrustify/master with SUPER_LOG defined [here](https://github.com/uncrustify/uncrustify/blob/master/src/log_rules.h#L38), you get a FTBFS due to the wrong definition of log_rule3() [here](https://github.com/uncrustify/uncrustify/blob/master/src/log_rules.cpp#L36).

This PR fixes the function definition. To verify the fix, you need to define SUPER_LOG and build. Please note that if you do so, some tests will fail.

### DIFF
--- a/src/log_rules.cpp
+++ b/src/log_rules.cpp
@@ -33,7 +33,7 @@ void log_rule2(const char *func, size_t line, const char *rule, chunk_t *first, 
 #ifdef SUPER_LOG
 
 
-void log_rule3(const char *func, size_t line, const char *rule)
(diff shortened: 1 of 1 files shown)


micheleCTDE:
Side note: I think printing the line number used in log_rule_B() is vert useful and IMO we should enable this by default. 
Once this PR is merged, I can prepare a PR for that in case you agree on the above.

gmaurel (approved):

gmaurel:
Change the default would be fine. Do it, please.

PR #3106, opened 2021-03-03
correct alignment if call with namespace

### LINKED ISSUE #3010: Incorrect indentation of initializer list when namespaces are involved (C++)
Issue: When calling a constructor that has a namespace in the initializer list, it impacts the alignment of the next element in the initializer list.

Version: uncrustify 0.71

Input:
```cpp
namespace SomeLongNamespaceName {
class Foo { };
}

class Bar : SomeLongNamespaceName::Foo {
public:
Bar()
	: SomeLongNamespaceName::Foo(),
	myNumber(3), // <-- this line
	myOtherNumber(5)
{
}
private:
int myNumber;
int myOtherNumber;
};
```

Output:
```cpp
namespace SomeLongNamespaceName {
class Foo { };
}

class Bar : SomeLongNamespaceName::Foo {
public:
Bar()
	: SomeLongNamespaceName::Foo(),
	                         myNumber(3), // <-- this line
	myOtherNumber(5)
{
}
private:
int myNumber;
int myOtherNumber;
};
```

Config: 100% default. When trying to get a minimal working example, I ended up with an empty config file, and I could still reproduce it.

(Edit: made the input match the output format and added a comment so the bug is more visible)

guy-maurel:
fixes #3010


micheleCTDE:
No regressions on my usual test sample of 3200 files.

guy-maurel:
replaced by #3111

PR #3107, opened 2021-03-04
suppress some duplicate definitions

guy-maurel:
local duplicates shadow the less local definitions


src/indent.cpp
```diff
@@ -3277,7 +3277,6 @@ void indent_text(void)
             pc->indent.ref   = frm.top().ip.ref;
             pc->indent.delta = frm.top().ip.delta;
          }
-         char copy[1000];
```
micheleCTDE, line 3280: nice. I had spot this as well but had not prepared a PR for it yet :smile:

gmaurel (approved):

PR #3121, opened 2021-03-17
the code braces are added to the case block

### LINKED ISSUE #3058: Comment added on wrong line after case brace
Uncrustify-0.72.0-103-b0598a72

Files in attachment:

The original code has a comment on the same line of a break inside a case block.
```
        break; // Just a comment
```
After uncrustifying the code braces are added to the case block but the comment is incorrectly placed after the added closing brace.
```
			break;
		}          // Just a comment
```

[comment_after_brace.zip](https://github.com/uncrustify/uncrustify/files/5824261/comment_after_brace.zip)

guy-maurel:
the comment is correctly placed
fixes #3058


micheleCTDE:
Hi Guy,
thanks for the PR. It works well for the problem reported in issue #3058 but it introduced a regression for another case. I will add a new test case in issue #3058, so we keep all in one place.

micheleCTDE:
Hi Guy,
thanks for the additional work. Tested and it fixes all the cases except one. I will upload the new test sample on #3058. We are almost there :smile:

gmaurel (approved):

PR #3125, opened 2021-03-19
add indent_comma_brace option

### LINKED ISSUE #3114: Indenting comma to opening brace possible?
I have the following snippet:
```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer, myLongExpressionYieldingASize };
}
```

I want it to break at the comma, as the line exceeds the right margin. I use the following options:

```
code_width         = 80
pos_comma          = lead_force
ls_func_split_full = true
```

I get the following result:
```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer
                        , myLongExpressionYieldingASize };
}
```

However, I'd like to have the comma align with the opening brace, similar to what `indent_comma_paren` does:
```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer
                      , myLongExpressionYieldingASize };
}
```

I experimented a bit and it seems like adding an option `indent_comma_brace` would be quite possible. But before I create a pull request, I wondered if I maybe missed some option that already achieves that?

Thanks!

gg-yb:
This works similar to indent_comma_paren.

One example for its usage is the following snippet:

```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer, myLongExpressionYieldingASize };
}
```

which uses C++11 list initialization.

If you want this snippet to break with a leading comma, you could use
the following options:

```
code_width         = 80
pos_comma          = lead_force
ls_func_split_full = true
```

which yields the following formatted snippet:

```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer
                        , myLongExpressionYieldingASize };
}
```

In order to align the comma to the brace, you can use the new option:

```
indent_comma_brace = true
```

which results in:

```cpp
std::string foo()
{
    return std::string{ myLongExpressionYieldingABuffer
                      , myLongExpressionYieldingASize };
}
```

**Merging this closes #3114**

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -3574,9 +3574,13 @@ void indent_text(void)
          else if (chunk_is_token(pc, CT_COMMA))
          {
             log_rule_B("indent_comma_paren");
+            bool comma_paren = options::indent_comma_paren()
+                               && chunk_is_paren_open(frm.top().pc);
+            log_rule_B("indent_comma_brace");
+            bool comma_brace = options::indent_comma_brace()
+                               && chunk_is_opening_brace(frm.top().pc);
 
-            if (  options::indent_comma_paren()
-               && chunk_is_paren_open(frm.top().pc))
+            if (comma_paren || comma_brace)
             {
                indent_column_set(frm.top().pc->column);
             }
--- a/src/options.h
+++ b/src/options.h
@@ -1517,6 +1517,11 @@ indent_paren_after_func_decl;
 extern Option<bool>
 indent_paren_after_func_call;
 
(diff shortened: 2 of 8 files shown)


gmaurel (changes requested):
Fine.
But complete please with some (2) tests.
Have a look to https://github.com/uncrustify/uncrustify/blob/master/README.md
A note about pull requests

gg-yb:
Thanks for your feedback, I will make the suggested changes in the following week.

gg-yb:
I will re-open this with an appropriately named branch...

Issue #3127, opened 2021-03-24
Incorrect parsing of template angle brackets.


t-bltg:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"
Forcing a space around compare operators fails template angle brackets (#1462) :

`$ cat tests/input/cpp/bug_1462.cpp | uncrustify -l cpp -c <(echo 'sp_compare=force')`

**stdout**
```c++
#include <type_traits>

template <
	typename ... Args,
	typename E = typename std::enable_if<(sizeof...(Args) >= 1), bool>::type
	>
void fun1(Args&& ... args)
{
}

template <
	typename ... Args,
	typename E = typename std::enable_if<(sizeof...(Args) > 1), bool>::type
	>
void fun2(Args&& ... args)
{
}

template <
	typename ... Args,
	typename E = typename std::enable_if < (sizeof...(Args) < 3), bool > ::type
	//                                  ^^^                ^^^        ^^^
	//                                  NOK                 OK        NOK
	>
void fun3(Args&& ... args)
{
}
```
Using latest `Uncrustify 0.72.0`.

PR #3134, opened 2021-04-01
Fix incorrect parsing of open angle token in template context

t-bltg:
Proposed fix for #3127.


gmaurel (approved):
Fine catch.
Thanks

t-bltg:
Thanks!

Issue #3126, opened 2021-03-23
Issue with lambda and list initialization.


t-bltg:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"
Consider the following config file:
**uc.cfg**
```
nl_after_brace_open             = true
indent_with_tabs                = 0
nl_cpp_lambda_leave_one_liners  = true
sp_inside_type_brace_init_lst   = remove
```
and the following input file:
**a.cc**
```c++
int main()
{
        auto f1 = [&]() { return 1; };
        auto f2 = [&]() ->decltype(auto) { return 2; };
        string s1{ 'a', 'b' };
}
```
Running `$ cat a.cc | uncrustify -l cpp -c uc.cfg -o b.cc` yields:
**b.cc**
```c++
int main()
{
        auto f1 = [&]() { return 1; };
        auto f2 = [&]() ->decltype(auto) {return 2;};
        //                               ^^^ [sp_inside_type_brace_init_lst = remove] NOK
        string s1{'a', 'b'};
        //       ^^^ [sp_inside_type_brace_init_lst = remove] OK
}
```

Using latest `uncrustify-0.72.0`.

[...]

PR #3135, opened 2021-04-01
Fix incorrect parsing BRACED_INIT_LIST in lambda & decltype context.

t-bltg:
Proposed fix for https://github.com/uncrustify/uncrustify/issues/3126.


gmaurel (approved):
Fine catch.
Thanks.

PR #3136, opened 2021-04-02
Fix for issue #3116 [C++] lambdas are not aligned within compound statements

CharlieYJH:
This is my proposed fix for https://github.com/uncrustify/uncrustify/issues/3116. It fixes the cases raised in the issue and tries to fix other common cases as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ xcuserdata
 # CLion
 .idea/
 cmake-build-*/
+
+# Vim
+*.swp
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1578,6 +1578,10 @@ void indent_text(void)
          {
             log_rule_B("indent_cpp_lambda_body");
             frm.top().brace_indent = frm.prev().indent;
+            if (are_chunks_in_same_line(frm.prev().pc, chunk_get_prev_ncnnlnp(frm.top().pc)))
+            {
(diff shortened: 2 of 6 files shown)


gmaurel (approved):
Thank for the PR

mwoehlke:
So... this caused a regression:
```c++
// What you want
const auto compare = [] (const auto i, const auto j)
{
    return i >= j;
};

// What neither of us wants
const auto compare = [] (const auto i, const auto j)
    {
        return i >= j;
    };

// What I want and used to get
const auto compare = [] (const auto i, const auto j)
                     {
                         return i >= j;
                     };

// But now it is this which nobody wants
const auto compare = [] (const auto i, const auto j)
                 {
                     return i >= j;
                 };
```

CharlieYJH:
Could you share your config?

mwoehlke:
> Could you share your config?

https://github.com/Kitware/kwiver/blob/master/.uncrustify.cfg

mwoehlke:
...and here's a test case:
```c++
void func()
{
  auto deleter = []( void* ptr )
                 {
                   free( ptr );
                 };
}
```

Prior to this PR, the above was unchanged. With this PR, it is malformatted.

CharlieYJH:
Opened a PR at https://github.com/uncrustify/uncrustify/pull/3200 which tries to fix this regression.

PR #3140, opened 2021-04-09
Nested lambda inside function fix

CharlieYJH:
This fixes a regression introduced by https://github.com/uncrustify/uncrustify/pull/3136 which messes up the formatting for lambdas inside a function that is nested in another function/lambda, as such:

```
obj.AddObject(Object::UniqueName(), 10, [this] {
    holder.Access([this](const auto &info) {
    if (IsGood(info)) {
        Add(info);
    }
});
});
```

Expected:

```
obj.AddObject(Object::UniqueName(), 10, [this] {
    holder.Access([this](const auto &info) {
        if (IsGood(info)) {
            Add(info);
        }
    });
});
```

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1615,14 +1615,30 @@ void indent_text(void)
                }
             }
 
+            bool toplevel = true;
+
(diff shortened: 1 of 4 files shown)


gmaurel (approved):
Fine
Thanks

PR #3149, opened 2021-04-18
Sp paren comma

guy-maurel:
Add a test for sp_paren_comma


guy-maurel:
have a look to
30944  sp_paren_comma.cfg                   cpp/sp_paren_comma.cpp

PR #3161, opened 2021-05-05
Dont break macro definitions

### LINKED ISSUE #3156: Operator leaking from #define statement
This code:
```c
#define X(a, b, ...) 1 +
static c d[e] =
#undef X
0;
```
becomes:
```c
#define X(a, b, ...) 1
+ static c d[e] =
#undef X
    0;
```
when running with our config file and version `Uncrustify_d-0.72.0`. It is the same result with 0.68 and 0.69, and with other operators like `-`, `*`, `|`, `&` or `/`. ` >>` and `<<` works fine though.

I expect the `+` to stay in the macro's definition so the code keeps working.

Here's the debug file: 
[debug.txt](https://github.com/uncrustify/uncrustify/files/6350824/debug.txt)

A workaround is to not write such horrible macros in the first place.

bofh69:
This fixes #3156

With  pos_arith = lead, newlines_chunk_pos(CT_ARITH, ...) called chunk_move_after with "+" as pc when formatting "#define X 1 +\n int a;". chunk_move_after just moves the newline before the plus, without adding a backslash before, causing the program to become invalid.

This fix just avoids the call, like it is done a few lines further down in the same function.

A better solution would perhaps be to do something like chunk_move_after here, but also add a backslash if needed. I don't know the API enough to pull that off.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -5552,8 +5552,12 @@ void newlines_chunk_pos(c_token_t chunk_type, token_pos_e mode)
 
             if (next->nl_count == 1)
             {
-               // move the CT_BOOL to after the newline
-               chunk_move_after(pc, next);
+               if (  prev != nullptr
+                  && !prev->flags.test(PCF_IN_PREPROC))
+               {
+                  // move the CT_BOOL to after the newline
+                  chunk_move_after(pc, next);
+               }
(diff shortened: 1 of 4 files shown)


gmaurel (approved):
fine, thanks

PR #3185, opened 2021-05-19
Fixed FTBFS when using RelWithDebInfo config in debian-like distros.

micheleCTDE:
When building with RelWithDebInfo, I have a FTBFS caused by failing 'tracking_space' test in test_cli_options.py due to wrong config used during the test ('Debug').

```
The following tests FAILED:
        1436 - cli_options (Failed)
Errors while running CTest
```


gmaurel (approved):
fine

PR #3186, opened 2021-05-21
Fixed detection of python on distros with older cmake (< 3.12)

micheleCTDE:
This was causing FTBFS on those distros.

Backported from this commit:
https://mirror.git.trinitydesktop.org/gitea/TDE/extra-dependencies/commit/766688b5179933bc1772c4572efcd884c4d8159f


gmaurel (approved):
Thanks to Slávek Banko

PR #3189, opened 2021-05-22
Minor fix to location of log message for oneliners

micheleCTDE:
The log message for keeping one liners should be printed only if the new line is not added, not all the times in this part of the code, IMO.


gmaurel (approved):
fine, the actions are working

PR #3193, opened 2021-05-28
add nl_before_struct option

gg-yb:
This works similar to nl_before_class, but the option to modify newlines
before structs is somehow missing, as far as I can see.

Currently, we can do the following:

```cpp
int foo = 1;
class Bar {};
int baz = 2;
```

and

```
nl_before_class = 2
```

and we obtain

```cpp
int foo = 1;

class Bar {};
int baz = 2;
```

Alas, when we have

```cpp
int foo = 1;
struct Bar {};
int baz = 2;
```

this is currently not possible.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -6024,15 +6024,17 @@ void do_blank_lines(void)
          }
       }
 
-      // Control blanks before a class
+      // Control blanks before a class/struct
       if (  (  chunk_is_token(prev, CT_SEMICOLON)
             || chunk_is_token(prev, CT_BRACE_CLOSE))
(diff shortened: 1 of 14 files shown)


gmaurel (changes requested):
Please go to:
  https://github.com/gg-yb/uncrustify/actions
and activate the workflow(s)

gg-yb:
Thanks for your feedback, I enabled the workflows.

gg-yb:
I have pushed a change, see the action here https://github.com/gg-yb/uncrustify/actions/runs/891858037

gmaurel (approved):

PR #3212, opened 2021-06-22
Add new option nl_before_ignore_after_case

clint-lawrence:
This option is useful when the following are set to add or force:
- nl_before_if
- nl_before_for
- nl_before_while
- nl_before_do
- nl_before_switch
- nl_before_synchronized

Often, the block following a case will be indented from the case. In
that circumstance, the indent provides sufficent visual separation
and forcing a newline before isn't necessary.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -1080,6 +1080,12 @@ static void newlines_if_for_while_switch_pre_blank_lines(chunk_t *start, iarf_e 
       }
       else
       {
+         if (  chunk_is_token(pc, CT_CASE_COLON)
+            && options::nl_before_ignore_after_case())
+         {
+            return;
+         }
+
          if (do_add) // we found something previously besides a comment or a new line
          {
             // if we have run across a newline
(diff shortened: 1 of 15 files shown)


gmaurel (changes requested):
uncrustify has already the option _nl_after_label_colon_
Isn't the same?

clint-lawrence:
No, I don't think so. That is only for goto label. I notice now that there is also nl_after_case, but it doesn't do what I'm after either.

The difference is how it works when using options nl_before_for/if/while etc set to add or force.

nl_after_case is applied in newlines_cleanup_braces. This is called before newlines_insert_blanklines.
https://github.com/uncrustify/uncrustify/blob/799d36bbda9db8cde9639224e4b1280992ca197d/src/newlines.cpp#L4489-L4505

If the original source had
```
case LABEL: if(cond) do_stuff();
```
nl_after_case will add a newline
```
LABEL:
    if(cond) do_stuff();
```

However, I have nl_before_if=force, which inserts a blank line before the if. Because it runs after nl_after_case , it takes precedence. The result is:
```
case LABEL:

    if(cond) do_stuff();
```
But I would like
```
case LABEL:
    if(cond) do_stuff();
```

gmaurel (approved):
Fine, thanks

PR #3220, opened 2021-06-28
introduce the options sp_after/before_ptr_star_trailing

guy-maurel:
Enable the value _ignore_ for the *ptr_star* options
Update some test configurations and test data.


guy-maurel:
Some tests are not successfull, #3222 
The PR will be replaced later.

PR #3236, opened 2021-07-13
Add failing test for C++ if-init-statement with `mod_full_paren_if_bool=true`

bertwesarg:
`mod_full_paren_if_bool` wrongly includes the init statement, when adding
parenthese. Like in:

```c++
if ((auto val = getValue(); val < 0)|| (val > 0))
```

Instead of

```c++
if (auto val = getValue(); (val < 0) || (val > 0))
```

### DIFF
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -993,3 +993,5 @@
 60055  issue_3116.cfg                       cpp/issue_3116.cpp
 60056  issue_3116.cfg                       cpp/issue_3116-2.cpp
 60057  issue_3116-2.cfg                     cpp/issue_3116.cpp
+
+60100  mod_full_paren_if_bool.cfg           cpp/if_init_statement.cpp
--- a/tests/expected/cpp/60100-if_init_statement.cpp
+++ b/tests/expected/cpp/60100-if_init_statement.cpp
@@ -1,0 +1,19 @@
+/* multi boolean works, but removes ' ' before '||' */
+if ((val < 0) || (val > 0))
+{
+}
+
+/* single condition works */
+if (auto val = getValue(); condition(val))
+{
+}
+
+/* multi boolean expression with existing parentheses will not change */
+if (auto val = getValue(); (val < 0) || (val > 0))
+{
+}
+
+/* multi boolean expression without existing parentheses produces invalid C++ code */
+if (auto val = getValue(); (val < 0) || (val > 0))
+{
+}
--- a/tests/input/cpp/if_init_statement.cpp
+++ b/tests/input/cpp/if_init_statement.cpp
@@ -1,0 +1,19 @@
+/* multi boolean works, but removes ' ' before '||' */
+if (val < 0 || val > 0)
+{
+}
+
+/* single condition works */
+if (auto val = getValue(); condition(val))
+{
+}
+
+/* multi boolean expression with existing parentheses will not change */
+if (auto val = getValue(); (val < 0) || (val > 0))
+{
+}
+
+/* multi boolean expression without existing parentheses produces invalid C++ code */
+if (auto val = getValue(); val < 0 || val > 0)
+{
+}


bertwesarg:
Please also note, the removed space before `||`

bertwesarg:
The whole point of this PR is to show you, that uncrustify produces broken code, if `mod_full_paren_if_bool=true`. Someone should fix it and then this PR will also succeed.

gmaurel:
I cannot reproduce. I get:
```.cpp
Uncrustify_d-0.73.0-76-a6df3c5c
false
Parsing: bertwesarg.cpp as language CPP
if (auto val = getValue(); (val < 0) || (val > 0))
true
Parsing: bertwesarg.cpp as language CPP
if (auto val = getValue(); (val < 0) || (val > 0))
```
Could you push the two output files, p.txt and A.txt using:
`uncrustify -c bertwesarg-t.cfg -f bertwesarg.cpp -p p.txt -L A 2>A.txt`

bertwesarg:
But you changed the input file! If there are already parentheses around the binary expression, `uncrustify` does not add more parentheses and thus does not invalid C++ code.

gmaurel:
No.
I have changed the configuration file.
With YOUR source and YOUR configuration, could you please push the files I wish to get?
Could you push the two output files, p.txt and A.txt?

bertwesarg:
I may be able to _attach_ these two files, but where should I _push_ them? In any case, you should be able to produce these two rather simple by yourself (again, this is the whole point of this PR, that the developers have a test case at hand and the CI can already reproduce the failure):

    $ ./uncrustify -c ../tests/config/mod_full_paren_if_bool.cfg -f ../tests/input/cpp/if_init_statement.cpp -p p.txt -L A 2>A.txt

Not sure what _my_ result can contribute to this at all, but here we go:

- [p.txt](https://github.com/uncrustify/uncrustify/files/6836455/p.txt)
- [A.txt](https://github.com/uncrustify/uncrustify/files/6836458/A.txt)

And here is stdout:

```
/* multi boolean works, but removes ' ' before '||' */
if ((val < 0)|| (val > 0))
{
}

/* single condition works */
if (auto val = getValue(); condition(val))
{
}

/* multi boolean expression with existing parentheses will not change */
if (auto val = getValue(); (val < 0) || (val > 0))
{
}

/* multi boolean expression without existing parentheses produces invalid C++ code */
if ((auto val = getValue(); val < 0)|| (val > 0))
{
}
```

gmaurel:
You are using version 0.58. This is too old for me. I begann with 0.62
Current is 0.73.
Please update your uncrustify, run again and attach the new output files. Thanks.
The "... removes ' ' before '||' ..." is solved.
You need to modify your configuration file with:
**sp_bool              = force**
The last case is a real bug. Bugfix is comming soon.

bertwesarg:
I have no idea why `./uncrustify --version` advertised itself as version 0.58. My build and this PR are based on master (087209872ff4caa37488623a6ae8b4c8da4e8101). Here is my ./uncrustify_version.h` from the build tree:

```c++
/**
 * @file uncrustify_version.h
 * Simply defines UNCRUSTIFY_VERSION.
 * I don't particularly like how autoconf and friends handle the version...
 *
 * @author  Ben Gardner
 * @license GPL v2+
 */
#ifndef UNCRUSTIFY_VERSION_H_INCLUDED
#define UNCRUSTIFY_VERSION_H_INCLUDED

#ifdef DEBUG
#define UNCRUSTIFY_VERSION    "Uncrustify_d-0.73.0-83-83132f4d"
#else
#define UNCRUSTIFY_VERSION    "Uncrustify-0.73.0-83-83132f4d"
#endif

#endif /* UNCRUSTIFY_VERSION_H_INCLUDED */
```

bertwesarg:
I solved the version mystery. This was a autoconf tree before and this had left its `src/uncrustify_version.h`, which was still 0.58. The build system should prioritize the build tree in its include order and not the source tree.

bertwesarg:
> The "... removes ' ' before '||' ..." is solved.
> You need to modify your configuration file with:
> **sp_bool = force**

With all due respect, but if `uncrustify` **removes** anything with **sp_bool = ignore**, then its a bug and I should not be forced to set **sp_bool = force** to workaround it.

bertwesarg:
> attach the new output files. Thanks.

Here is the diff to the new files:

```diff
-# 0.58
+# Uncrustify_d-0.73.0-83-83132f4d
```

guy-maurel:
PR #4099 is a bug fix proposal.
Please test it and report.
Sorry for the delay.

guy-maurel:
PR #4110 is a bug fix proposal.
Please test it and report.

PR #3244, opened 2021-07-17
Add sp_sparen_paren option

### LINKED ISSUE #3209: sp_inside_sparen doesn't add any spaces if the next character is also a parenthesis
Input code:

```cpp
#include <stdio.h>
int main() {
	FILE *f;
	if ((f = fopen("/dev/null", "r")))
		puts("file is open");
	return 0;
}
```

(The double parentheses in the `if` statement are to avoid a GCC warning when `-Wall` is enabled.)

Config file:

```
sp_inside_sparen add
```

I would expect this configuration to produce one of the following two results (either is acceptable):

```
	if ( (f = fopen("/dev/null", "r")) )
```

```
	if (( f = fopen("/dev/null", "r") ))
```

Instead, the source code is returned unmodified.

```
$ uncrustify --version
Uncrustify_d-0.73.0_f
```

alexhenrie:
This option adds or removes spaces between double parentheses in control
statements. If set to 'ignore', the previous behavior controlled by
sp_paren_paren is preserved.

Fixes #3209

### DIFF
--- a/tests/c.test
+++ b/tests/c.test
@@ -394,3 +394,5 @@
 10014  indent_continue-8.cfg                c/sparen-indent.c
 10015  empty.cfg                            c/Issue_2845.h
 10016  Issue_3233.cfg                       c/Issue_3233.c
+10017  sp_sparen_paren-a.cfg                c/double-sparen.c
+10018  sp_sparen_paren-i.cfg                c/double-sparen.c
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -316,6 +316,9 @@ sp_inside_sparen_open           = ignore   # ignore/add/remove/force/not_defined
 # Overrides sp_inside_sparen.
 sp_inside_sparen_close          = ignore   # ignore/add/remove/force/not_defined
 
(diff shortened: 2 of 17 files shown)


tests/c.test
```diff
@@ -394,3 +394,4 @@
 10014  indent_continue-8.cfg                c/sparen-indent.c
 10015  empty.cfg                            c/Issue_2845.h
 10016  Issue_3233.cfg                       c/Issue_3233.c
+10017  sp_sparen_paren.cfg                  c/double-sparen.c
```
gmaurel, line 397: Please add another example with the value ignore, to test it.

gmaurel (changes requested):

gmaurel:
Have a look to:
https://github.com/alexhenrie/uncrustify/actions
and activate the workflow run.
Thanks

tests/c.test
alexhenrie: I added a test that shows that when `sp_sparen_paren=ignore`, `sp_paren_paren=add` continues to apply, which I assume is what you meant?

gmaurel (approved):

alexhenrie:
Thank you!

PR #3247, opened 2021-07-28
Add pp_indent_in_guard option

### LINKED ISSUE #3246: Uncrustify never indents code inside of a header guard
Uncrustify currently treats header guards specially: It does not ever indent their contents. For example:

Config file:

```
pp_if_indent_code = true
pp_indent = add
pp_indent_count = 8
```

Before:

```
#ifndef FOO
#define FOO

int foo();

#ifdef BAR
#define BAZ
int bar();
#endif
#endif
```

After:

```
#ifndef FOO
#define FOO

int foo();

#ifdef BAR
	#define BAZ

	int bar();
#endif
#endif
```

I need an option to disable this special treatment and indent the body of the `#ifndef` anyway, like so:

```
#ifndef FOO
	#define FOO

	int foo();

	#ifdef BAR
		#define BAZ

		int bar();
	#endif
#endif
```

alexhenrie:
This option allows treating an #if that guards the entire file the same
as any other #if.

Fixes #3246

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -3027,6 +3027,9 @@ pp_indent_if                    = 0        # number
 # Whether to indent the code between #if, #else and #endif.
 pp_if_indent_code               = false    # true/false
 
+# Whether to indent the body of an #if that encompasses all the code in the file.
+pp_indent_in_guard              = false    # true/false
+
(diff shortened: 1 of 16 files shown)


gmaurel (approved):
Fine, thanks.

alexhenrie:
Thank you!

PR #3250, opened 2021-07-29
Add indent_single_line_comments_after option

### LINKED ISSUE #3249: Need option like indent_sing_line_comments but for comments that follow code rather than preceding it
The `indent_sing_line_comments` option is great for indenting comments that come before a line of code:

```
    // Greet the user.
printf("Hello world!\n");
```

However, to implement my company's style guide, I actually need an option to indent comments that come _after_ a line of code:

```
printf("Hello world!\n");
    // Greet the user.
```

alexhenrie:
And rename indent_sing_line_comments to
indent_single_line_comments_before. The two options do the same thing,
except one applies to comments that precede code and the other applies
to comments that follow code. If a comment is in between two lines of
code, the single_line_comments_before option takes precedence.

Fixes #3249


gmaurel (changes requested):
fine, but: 
you cannot change the name of an option without adding a `deprecated` information, such as:
option.cpp:330

```.cpp
bool process_option_line_compat_0_70(const std::string &cmd,
                                     const char        *filename)
{
   if (cmd == "sp_word_brace")                     // Issue #2428
...
```
Please align the assign characters in the config files

alexhenrie:
> you cannot change the name of an option without adding a `deprecated` information

Done.

> Please align the assign characters in the config files

I tried aligning the equals signs, but it caused `make test` to fail.

gmaurel (approved):
Fine, thanks

alexhenrie:
Thank you!

PR #3253, opened 2021-07-29
Add $(year) keyword for use in headers and footers

### LINKED ISSUE #3251: Option to output current year in header and footer
I'd like to use the `cmt_insert_file_header` option to insert the following at the top of every file, without having to manually increment the year every January 1:

```
// Copyright 2021 VPI Engineering
```

A new `$(year)` keyword (similar to `$(filename)` etc.) would be perfect.

alexhenrie:
Fixes #3251

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -17,6 +17,7 @@
 #include "unc_ctype.h"
 #include "unicode.h"
 
+#include <ctime>
 #include <map>
 #include <regex>
 #include <set>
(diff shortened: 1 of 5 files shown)


gmaurel (changes requested):
Can you provide a test example?

alexhenrie:
The expected output would have to be dynamically generated to have the current year. Is there an example of a test case that dynamically generates its output?

guy-maurel:
Have a look to tests/cli/test_cli_options.py at line 775:
```
    for L in Ls_A:
...
                err_result_manip=[reg_replace(r'\([0-9]+\)', ' '),
...
```
The function **reg_replace** is what you need.

alexhenrie:
OK, I have added a test to test_cli_options.py that finds and replaces the current year with a static string. Does it look all right to you?

gmaurel (approved):
Fine, thanks

alexhenrie:
Thank you!

Issue #3252, opened 2021-07-29
sp_after_tparen_close doesn't work if function pointer is inside a "typedef struct"


alexhenrie:
Config file:

```
sp_after_tparen_close = add
```

Test code:

```
void (*x)(void);

typedef struct
{
    void (*y)(void);
} z;
```

Expected output:

```
void (*x) (void);

typedef struct
{
    void (*y) (void);
} z;
```

Actual output:

```
void (*x) (void);

typedef struct
{
    void (*y)(void);
} z;
```

Debug output:

```
# Uncrustify_d-0.73.0_f
sp_after_tparen_close           = add
# option(s) with 'not default' value: 1
#
# -=====-
# number of loops               = 0
# -=====-
# language                      = C
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>               TYPE|           FUNC_VAR|     PARENT_NOT_SET[  1/  1/  5/  0][0/0/0][      c0000][0-0] void
#   1>        TPAREN_OPEN|           FUNC_VAR|     PARENT_NOT_SET[  6/  6/  7/  1][0/0/0][  203000000][0-0]      (
#   1>           PTR_TYPE|               NONE|     PARENT_NOT_SET[  7/  7/  8/  0][0/1/0][  200080000][0-0]       *
#   1>           FUNC_VAR|               NONE|     PARENT_NOT_SET[  8/  8/  9/  0][0/1/0][    3080000][0-0]        x
#   1>       TPAREN_CLOSE|           FUNC_VAR|     PARENT_NOT_SET[  9/  9/ 10/  0][0/0/0][  200000000][0-0]         )
#   1>        FPAREN_OPEN|         FUNC_PROTO|     PARENT_NOT_SET[ 11/ 10/ 11/  0][0/0/0][  200000000][0-0]           (
#   1>               TYPE|               NONE|     PARENT_NOT_SET[ 12/ 11/ 15/  0][0/1/0][      80000][0-0]            void
#   1>       FPAREN_CLOSE|         FUNC_PROTO|     PARENT_NOT_SET[ 16/ 15/ 16/  0][0/0/0][  200000000][0-0]                )
#   1>          SEMICOLON|           FUNC_VAR|     PARENT_NOT_SET[ 17/ 16/ 17/  0][0/0/0][  200000000][0-0]                 ;
[...]

Issue #1234, opened 2017-07-11
Investigate opening space after brace in certain scenarios


mihaipopescu:
*Single line functions*

    public string filepath {get; set; }

    bool HasPending() {return m_HasPendingDeletes != 0; }

    public AnimationCurve curveToSaveAsPreset {get {return m_Curve; } set {m_Curve = value; }}

This seems to happen with no spaces on the interior. Opening brace doesn't get one, closing brace does.

Turning on sp_inside_braces=add fixes it, but also changes a lot of initializer code we don't want to touch (like x = {1}). May need special support, or perhaps there's a bug..

*Initializers*

Not sure if this is what we want..

    session.CustomEvent("customEvent1", new Dictionary<string, object>(){ { "k1", "v1" }, { "k2", "v2" } });
    ... --> ...
    session.CustomEvent("customEvent1", new Dictionary<string, object>() { { "k1", "v1" }, { "k2", "v2" } });

Second line adds a space before the initializer {. Is that what we want for C#?

Expected versus output diff:
---
```diff
diff --git "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\10051-UNI-1338.cs" "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\10051-UNI-1338.cs"
index 3bdb6f0..e6f6642 100644
--- "a/C:\\work\\external\\uncrustify\\tests\\output\\staging\\10051-UNI-1338.cs"
+++ "b/C:\\work\\external\\uncrustify\\tests\\results\\staging\\10051-UNI-1338.cs"
@@ -1,27 +1,27 @@
 // *Single line functions*
 
 public class Class
 {
     public string foo {get; set;}
 
     bool HasBar() {return m_HasBar != 0;}
 
     public Bar prop {get {return m_bar;} set {m_bar = value;}}
 
 // This seems to happen with no spaces on the interior. Opening brace doesn't get one, closing brace does.
 
 // Turning on sp_inside_braces=add fixes it, but also changes a lot of initializer code we don't want to touch (like x = {1}). May need special support, or perhaps there's a bug..
 
 // *Initializers*
 
 // Not sure if this is what we want..
 
     public void foo()
     {
-        sas.Foo("bar", new Dictionary<string, object>(){ { "k1", "v1" }, { "k2", "v2" } });
+        sas.Foo("bar", new Dictionary<string, object>() { { "k1", "v1" }, { "k2", "v2" } });
         // ... --> ...
[...]

PR #3257, opened 2021-08-03
Apply sp_cparen_oparen to parenthesized indirect function calls

alexhenrie:
`sp_cparen_oparen` is the right option to apply to `(get_hello())()`. Furthermore, this patch provides a workaround for issue #3252 because setting `sp_cparen_oparen = ignore` will avoid unwanted space removal in cases where Uncrustify does not parse function pointers correctly.

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -1602,9 +1602,9 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
          log_rule("sp_after_cast");
          return(options::sp_after_cast());
       }
-      // Must be an indirect/chained function call?
-      log_rule("REMOVE");
-      return(IARF_REMOVE);  // TODO: make this configurable?
+      // Probably a parenthesized indirect function call or similar (issue #3260)
+      log_rule("sp_cparen_oparen");
+      return(options::sp_cparen_oparen());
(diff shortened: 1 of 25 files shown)


guy-maurel:
I am not sure about your diagnostic.
I found a bug at the function do_symbol_check, which make a difference between the lines
#3252:
`void (*x)(void);`
and
`void (*y)(void);`
The code:
`   if (  !pc->flags.test_any(PCF_IN_TYPEDEF | PCF_IN_TEMPLATE)`
makes it impossible to call
`mark_function_type(pc);`
which changes the **x** from **CT_TYPE** to **CT_FUNC_VAR**
Please have a look to my proposal #3258 and comment.

gmaurel (changes requested):
Please add something such as:
`// Issue #1234`
Should the comment/question: 
 `// Must be an indirect/chained function call?`
stay or can be removed?
Thanks

alexhenrie:
I have rewritten the comment to be more clear and include the issue number:

```
// Probably a parenthesized indirect function call or similar (issue #3260)
```

gmaurel (approved):

alexhenrie:
Thank you!

PR #3281, opened 2021-08-16
Add sp_between_semi_for_empty option

### LINKED ISSUE #3280: Need an option to add spaces around "forever" loops but not between the semicolons
Config file:

```
sp_before_semi_for_empty = add
sp_after_semi_for_empty = add
```

Example code:

```
int main(void)
{
	for (; ;)
		continue;
}
```

Output:

```
int main(void)
{
	for ( ; ; )
		continue;
}
```

What I'm looking for:

```
int main(void)
{
	for ( ;; )
		continue;
}
```

So, Uncrustify has the options to add a space before the first semicolon and after the second semicolon, but it has no option to remove the space in the middle.

alexhenrie:
Fixes #3280

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -412,18 +412,24 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
 
       if (get_chunk_parent_type(second) == CT_FOR)
       {
-         if (  chunk_is_token(first, CT_SPAREN_OPEN) // a
-            || chunk_is_token(first, CT_SEMICOLON))  // b
+         if (chunk_is_token(first, CT_SPAREN_OPEN))
          {
-            // empty, ie for (;;)
-            //               ^ is first    // a
-            //                ^ is second  // a
-            // or
-            //                ^ is first   // b
-            //                 ^ is second // b
-            // Add or remove space before a semicolon of an empty part of a for statement.
+            // empty, e.g. for (;;)
+            //                 ^ is first
+            //                  ^ is second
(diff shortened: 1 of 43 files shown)


gmaurel (commented):
have a look to
https://github.com/uncrustify/uncrustify/issues/3280#issuecomment-900392451

gmaurel (changes requested):
Fine, but we need to test the case if 
 ` options::sp_between_semi_for_empty() == IARF_IGNORE`

alexhenrie:
Done. Please let me know if there's anything else I should change.

src/space.cpp
```diff
@@ -412,6 +412,14 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
 
       if (get_chunk_parent_type(second) == CT_FOR)
       {
+         if (  chunk_is_token(first, CT_SEMICOLON)
+            && options::sp_between_semi_for_empty() != IARF_IGNORE)
+         {
+            // Add or remove space between semicolons of an empty middle part of a for statement.
+            log_rule("sp_between_semi_for_empty");
+            return(options::sp_between_semi_for_empty());
+         }
```
gmaurel, line 421: we need here some code to consider
`options::sp_between_semi_for_empty() == IARF_IGNORE`

alexhenrie: OK, I have modified the patch so that sp_before_semi_for_empty and sp_between_semi_for_empty function independently.

gmaurel (approved):

gmaurel:
Fine, thanks

PR #3286, opened 2021-08-17
Make sure pointers are valid before dereferencing them

micheleCTDE:
As per PR title.
Passed all tests and also tested on about 3200 source files successfully.

### DIFF
--- a/src/chunk_list.h
+++ b/src/chunk_list.h
@@ -470,21 +470,23 @@ bool are_chunks_in_same_line(chunk_t *start, chunk_t *end);
  */
 static inline bool is_expected_type_and_level(chunk_t *pc, c_token_t type, int level)
 {
-   // we don't care about the level (if it is negative) or it is as expected
-   // and the type is as expected
-   return(  (  level < 0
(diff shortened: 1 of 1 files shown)


gmaurel (changes requested):
if the next line is 
`  chunk_is_token(...`
it is not necessary to check.
OK?

micheleCTDE:
> if the next line is ` chunk_is_token(...` it is not necessary to check.

Hi Guy,
I agree with you that a call to 'chunk_is_token()' will basically check that the pointer is valid, and in fact I had also considered that myself. On the other hand the code is relying on it to check a pointer before dereferencing it. If for any reason in future we decide to change the behavior of 'chunk_is_token()' for nullptr, we would leave the door open to potential SEGV when dereferencing the pointer. Having a direct check on the pointer validity will prevent this situation.
What do you think?

If you prefer to keep it as it is now, let me know and I will amend the PR removing the changes where 'chunk_is_token()' is used.

micheleCTDE:
Btw, rebased on top of current master.

gmaurel (approved):

PR #3296, opened 2021-08-24
Fix wrong joining operation by pos_assign followed by '{'.

micheleCTDE:
According to documentation in defaults.cfg, pos_assign should not
influence '=' when followed by '{'.
When pos_assign = join, an '=' followed by a '{' on the next line are joined together, contrary to the documentation.
The PR fixes such behavior and avoid the join if the next token is a '{'


micheleCTDE:
One test case for pawn was affected, but it did not relate to the problem for which the test was put in place.

Issue #3196, opened 2021-05-31
cli_options test fails for 0.73.0


mjbogusz:
When building the newest version of Uncrustify, the build is failing on the test stage (`make -j1 test` -> `/usr/bin/ctest --force-new-ctest-process -j1`). I'm building in Docker (for a clean environment) with an up-to-date Ubuntu 20.04.

I've tried both the 0.73.0 as in the release and the latest master (e3d4054802c777252fd0ee967f4a75ce003d3341), the behavior is the same.

<details>
<summary>Build/test log for the latest master (the interesting part)</summary>

```
1446/1447 Test #1446: cli_options ......................***Failed    0.12 sec
Debug
Uncrustify binary found: /uncrustify-e3d4054802c777252fd0ee967f4a75ce003d3341/obj-x86_64-linux-gnu/uncrustify

************************************
Problem (1) with results/tracking_space.html
************************************
'  <html>'
'  <head>'
'     <meta http-equiv="content-type" content="text/html; charset=utf-8"/>'
'     <title>Uncrustify: where do the Spaces options work</title>'
'  </head>'
'  <body lang="en-US">'
'  <p>'
'  </p>'
'  <pre>'
'- {'
'-    a = b + 111 - 55;'
'- }'
'- template<typename ... A, int... B>'
'- struct foo1 : foo1<A..., (sizeof...(A) + B)...>'
'- {'
'-    foo1() {'
'-       int x = sizeof...(A);'
'-       bool b = x > 1;'
'-    }'
'- };'
'+ {<a title="1,REMOVE"><font color="red">M</font></a>'
('+    a<a title="2,sp_assign"><font color="red">M</font></a>=<a title="3,86,sp_assign"><font color="red">M</font></a>b<a title="4,87,93,sp_arith"><font color="red">M</font></a>+<a title="5,88,94,99,sp_arith"><font color="red">M</font></a>111<a '
 'title="6,89,95,100,104,sp_arith"><font color="red">M</font></a>-<a title="7,90,96,101,105,108,sp_arith"><font color="red">M</font></a>55<a title="8,91,97,102,106,109,111,sp_before_semi"><font color="red">M</font></a>;<a title="9,92,98,103,107,110,112,113,REMOVE"><font '
 'color="red">M</font></a>')
'+ }<a title="10,REMOVE"><font color="red">M</font></a>'
[...]

PR #3306, opened 2021-09-02
Fix tracking_space tests for non-debug builds

mjbogusz:
This should fix https://github.com/uncrustify/uncrustify/issues/3196.

Of course, packagers should use `Release` or `RelWithDebInfo` build types anyway, but with this PR other types (e.g. `MinSizeRel` or `undefined`) won't fail as well.


gmaurel:
Please, have a look to:
https://github.com/mjbogusz/uncrustify/actions
and activate the workflow

mjbogusz:
Activated and rebased onto master: https://github.com/mjbogusz/uncrustify/actions/runs/1198918001

gmaurel (approved):
Fine. Thanks.

PR #3328, opened 2021-09-28
Indent #line and #error to the same level as #pragma

### LINKED ISSUE #3327: #error and #line can't be indented at code level
The pp_define_at_level option causes both `#define` and `#pragma` to be indented at code level, but `#error` and `#line` are left out.

Config file:

```
pp_define_at_level = true
pp_if_indent_code = true
pp_indent_at_level = true
```

Test code:

```
int main(void)
{
#if 1
#pragma GCC warning "This code won't compile"
#define FOO 1
#line 7
#error
#endif
    return 0;
}
```

Expected output:

```
int main(void)
{
	#if 1
		#pragma GCC warning "This code won't compile"
		#define FOO 1
		#line 7
		#error
	#endif
	return 0;
}
```

Actual output:

```
int main(void)
{
	#if 1
		#pragma GCC warning "This code won't compile"
		#define FOO 1
	#line 7
	#error
	#endif
	return 0;
}
```

alexhenrie:
Fixes #3327

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -937,7 +937,8 @@ void indent_text(void)
             frm.top().indent_tab = frm.top().indent;
             log_indent_tmp();
          }
-         else if (  get_chunk_parent_type(pc) == CT_PP_PRAGMA
+         else if (  (  get_chunk_parent_type(pc) == CT_PP_PRAGMA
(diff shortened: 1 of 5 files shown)


gmaurel (approved):
Thanks

alexhenrie:
Thank you!

PR #3333, opened 2021-09-30
Drop cases covered by sp_*_constr_colon from add_space_table

### LINKED ISSUE #3330: Spaces are added around constructor colon even when sp_*_constr_colon = ignore
Config file:

```
indent_class = true
sp_before_constr_colon = ignore
sp_after_constr_colon = ignore
```

Test code:

```
class Spaceship
{
public:
	Spaceship():shields(100)
	{
	}

	int shields;
};
```

Expected results: No change

Actual results:

```
class Spaceship
{
public:
	Spaceship() : shields(100)
	{
	}

	int shields;
};
```

alexhenrie:
And create a separate indent_ctor_init_following option so that it's
still possible to indent the first initializer line less than subsequent
initializer lines.

Fixes #3330


src/options.h
```diff
@@ -601,11 +601,11 @@ sp_before_class_colon;
 
 // Add or remove space after class constructor ':'.
 extern Option<iarf_e>
-sp_after_constr_colon;
+sp_after_constr_colon; // = IARF_ADD
```
gmaurel, line 604: NO!
we cannot change a default value, which is already in use by many users.

alexhenrie: The default behavior doesn't change. Before this patch, sp_after_constr_colon=ignore and sp_after_constr_colon=add did the same thing (unless indent_ctor_init_leading is changed from its default value of 2). After this patch, setting sp_after_constr_colon=ignore works as expected and does not add any spaces. Therefore the default value of sp_after_constr_colon has to change to preserve the default behavior that users expect.

gmaurel (approved):

alexhenrie:
Thank you!

PR #3349, opened 2021-10-11
Add sp_return option

### LINKED ISSUE #3348: No option to remove all spaces or preserve extra spaces after return keyword
Test code:

```
{
    return    /*success*/0;
}
```

I need a way to change this code to say `return/*success*/0` (no spaces) and also a way to leave alone with all of the spaces. Currently Uncrustify always puts exactly one space after the `return` keyword no matter what.

alexhenrie:
Fixes #3348


alexhenrie:
Thank you!

PR #3372, opened 2021-11-02
Add indent_switch_body option

### LINKED ISSUE #3370: Need option to control switch indentation before "case"
I'm looking for a set of options that produces the following output:

```
int main(int argc, char** argv)
{
	switch (argc)
	{
	// If FOO or BAR is defined, treat it specially
	#ifdef FOO
		case FOO:
			return 1;
	#endif
	#ifdef BAR
		case BAR:
			return 2;
	#endif
	default:
		return 100;
	}
}
```

Note that both the comment and the #ifdefs are indented to the same level as the `default` case label.

With a default Uncrustify configuration, the comment has too much indentation:

```
int main(int argc, char** argv)
{
	switch (argc)
	{
		// If FOO or BAR is defined, treat it specially
	#ifdef FOO
	case FOO:
		return 1;
	#endif
	#ifdef BAR
	case BAR:
		return 2;
	#endif
	default:
		return 100;
	}
}
```

And when `pp_indent_at_level = true`, both the comment and the preprocessor directives are indented too much:

```
int main(int argc, char** argv)
{
	switch (argc)
	{
		// If FOO or BAR is defined, treat it specially
		#ifdef FOO
	case FOO:
		return 1;
		#endif
		#ifdef BAR
	case BAR:
		return 2;
		#endif
	default:
		return 100;
	}
}
```

alexhenrie:
Fixes #3370

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1206,6 +1206,10 @@ indent_relative_single_line_comments = false    # true/false
 # It might wise to choose the same value for the option indent_case_brace.
 indent_switch_case              = 0        # unsigned number
 
+# Spaces to indent the body of a 'switch' before any 'case'.
+# Usually the same as indent_columns or indent_switch_case.
+indent_switch_body              = 0        # unsigned number
+
 # Spaces to indent '{' from 'case'. By default, the brace will appear under
 # the 'c' in case. Usually set to 0 or indent_columns. Negative values are OK.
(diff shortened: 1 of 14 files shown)


gmaurel (approved):

gmaurel (changes requested):
have a look to the confilts

alexhenrie:
The conflict is resolved.

gmaurel (approved):

alexhenrie:
Thank you!

PR #3379, opened 2021-11-06
Add sp_inside_for, sp_inside_for_before, and sp_inside_for_after options

### LINKED ISSUE #3377: Need option like sp_inside_sparen but specifically for 'for' statements
To implement my company's style guide, I need Uncrustify to add spaces around control statements in general, like changing `if (x)` to `if ( x )`, but to ignore them in for statements: Both `for ( x = 0; x < 10; x++ )` and `for (x = 0; x < 10; x++)` are OK. `for` statements get an exception because they contain multiple individual statements.

alexhenrie:
Fixes #3377


gmaurel (approved):
Fine

alexhenrie:
Thank you!

PR #3381, opened 2021-11-08
Add a special indent_label value that keeps the original indentation

### LINKED ISSUE #3378: Need option to not reindent goto labels
I need an option to not change any of the following:

```
void foo(void)
{
    void *x = malloc(10);
    if (!x) return;

    void *y = malloc(10);
    if (!y) goto cleanup;

    void *z = malloc(10);
    if (!z) goto cleanup;

    bar(x, y, z);

cleanup:
    free(x);
    free(y);
    free(z);
}
```

```
void foo(void)
{
    void *x = malloc(10);
    if (!x) return;

    void *y = malloc(10);
    if (!y) goto cleanup;

    void *z = malloc(10);
    if (!z) goto cleanup;

    bar(x, y, z);

    cleanup:
    free(x);
    free(y);
    free(z);
}
```

```
void foo(void)
{
    void *x = malloc(10);
    if (!x) return;

    void *y = malloc(10);
    if (!y) goto cleanup;

    void *z = malloc(10);
    if (!z) goto cleanup;

    bar(x, y, z);

        cleanup:
    free(x);
    free(y);
    free(z);
}
```

alexhenrie:
Fixes #3378

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1252,6 +1252,7 @@ indent_comment_align_thresh     = 3        # unsigned number
 #
 #  >0: Absolute column where 1 is the leftmost column
 # <=0: Subtract from brace indent
+# -17: Keep original indentation
 #
 # Default: 1
 indent_label                    = 1        # number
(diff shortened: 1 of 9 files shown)


micheleCTDE:
Hi Alex,
I am not the mainteiner, but I don't think this is the right way. If you want to preserve the local of goto label, we should probably have a boolean option like "preserve_goto_label_indentation". If true, indentation is not affected. If false, the value of indent_label is used.
What do you think?

gmaurel:
-17 is not a fine value!
Try the proposal at https://github.com/uncrustify/uncrustify/pull/3381#issuecomment-963733506

alexhenrie:
See #3385

Issue #3383, opened 2021-11-10
the direct use doesn't work


guy-maurel:
I make:
```.txt
cd <uncrustify>/tests/cli/
./test_cli_options.py 
Uncrustify binary found: ../../build/uncrustify
Traceback (most recent call last):
  File "./test_cli_options.py", line 853, in <module>
    main(argv[1:])
  File "./test_cli_options.py", line 516, in main
    clear_dir(s_path_join(test_dir, 'results'))
  File "./test_cli_options.py", line 462, in s_path_join
    p_splits = list(path_split(path))
  File "/usr/lib/python3.8/posixpath.py", line 103, in split
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
It was some weeks ago possible.
It might be caused by:
eb8b6af4d2508dcfb9a46a8539f4e65e8e02917e
Please help!

micheleCTDE:
I will have a look!

micheleCTDE:
PR #3390 is a proposed fix.

guy-maurel:
Thanks

PR #3390, opened 2021-11-12
Fix issue #3383.

micheleCTDE:
Default value for 'test' folder was not set correctly. Now it is set to the same folder passed when calling the script from CMakeLists.txt


gmaurel (approved):
thanks

PR #3392, opened 2021-11-15
Remove unused test result files.

micheleCTDE:
This files are in the source code but are not generated nor compared during the build and test phase


gmaurel (approved):
Thanks for it!

PR #3394, opened 2021-11-20
Fix behavior of "chunk_get_next_ncnnlnp" and "chunk_get_prev_ncnnlnp" 

micheleCTDE:
From their names, "chunk_get_next_ncnnlnp" and "chunk_get_prev_ncnnlnp" 
are expected to return the next/prev chuck which is
not a comment, not a newline and not a preprocessor.
This was true only if the start chunk "cur" was a non preprocessor one.
Instead if "cur" was a preprocessor chunk, they would return either:
- the next non-preprocessor chuck
- the next preprocessor chuck that was both not a comment nor a new line.
The new version returns the expected chunk.

2. Add "chunk_get_next_ncnnl_in_pp" and "chunk_get_prev_ncnnl_in_pp"
functions, which return the next/prev preprocessor chuck that is
both not a comment nor a new line.

3. Some other code clean up and rearrangement

4. Prevent possible SEGV caused by dereferecing null pointer (this actually came up while testing the rest of this PR).

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -1413,9 +1413,9 @@ static void mod_case_brace(void)
    chunk_t *pc = chunk_get_head();
 
    // Make sure to start outside of a preprocessor line (see issue #3366)
-   while (chunk_is_preproc(pc))
+   if (chunk_is_preproc(pc))
    {
-      pc = chunk_get_next(pc);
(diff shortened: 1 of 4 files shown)


gmaurel (approved):
Fine

PR #3397, opened 2021-11-22
prepare a solution for 3353

gmaurel:
only a proposal to be test


micheleCTDE:
Tested on my usual test repo. No regressions whatssoever.

gmaurel:
Thanks for testing

PR #3408, opened 2021-11-29
Apply sp_ptr_star_func_var to function pointer types

### LINKED ISSUE #3402: sp_ptr_star_func_var doesn't work in typedefs
Config file:

```
sp_ptr_star_func_var = remove
sp_after_ptr_star = force
```

Test code:

```
int (*    func)(void);
typedef int (*    func_t)(void);
```

Expected result:

```
int (*func)(void);
typedef int (*func_t)(void);
```

Actual result:

```
int (*func)(void);
typedef int (* func_t)(void);
```

alexhenrie:
And mark function pointer types as CT_FUNC_TYPE in mark_function_type.

I'm pretty sure that this is the right approach because token_enum.h uses a typedef in its example for CT_FUNC_TYPE:

```
   CT_FUNC_TYPE,          // function type - foo in "typedef void (*foo)(void)"
```

If anyone wants to have a different rule for function pointer typedefs than for ordinary function pointers, let me know and I will happily make a separate `sp_ptr_star_func_type` option.

Fixes #3402

### DIFF
--- a/src/combine_fix_mark.cpp
+++ b/src/combine_fix_mark.cpp
@@ -2211,7 +2211,7 @@ bool mark_function_type(chunk_t *pc)
    {
       if (pc->flags.test(PCF_IN_TYPEDEF))
       {
-         set_chunk_type(varcnk, CT_TYPE);
+         set_chunk_type(varcnk, CT_FUNC_TYPE);   // Issue #3402
       }
       else
       {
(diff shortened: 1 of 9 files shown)


gmaurel (changes requested):
Yes, make it. 
It make sence to separate.

alexhenrie:
See pull request #3417

Issue #3422, opened 2021-12-11
Alignment on assignment not working with multiple assignments.


micheleCTDE:
uncrustify b9dc602c.
If multiple assignments are used in the same line, alignment of the following lines doesn't work correctly.

Input
```
void main()
{
        unsigned long uncompressed_size = 0, compressed_size = 0;
        char *data_buffer1 = NULL;

        unsigned long uncompressed_size = 0;
        char *data_buffer2 = NULL;
}
```

Output:
```
void main()
{
        unsigned long uncompressed_size = 0, compressed_size = 0;
        char         *data_buffer1 = NULL;

        unsigned long uncompressed_size = 0;
        char         *data_buffer2      = NULL;
}
```

Expected: same alignment of assignment for both data_buffer1 and data_buffer2 lines.

This issue is mostly for tracking purpose. I will work on a solution and propose a PR when ready.

gmaurel:
uncrustify doesn't support multiple declarations.
It is **evertime** better to spend some more lines for declarations.
Let us never begin with support multiple declarations.

micheleCTDE:
Hi Guy,
I agree on the fact that is better to split declarations on multiple lines, but in case a huge code base has multiple declarations on a line, wouldn't it be nicer to consider the first assignment for alignment with other lines? The example above is minimal, but the definition block could actually several lines longer.
I am not talking about aligning multiple assignments on multiple rows, just the most left side one.
That is I am talking about:
```
int aaaa = 1, bbbb = 2, cccc = 3;
int a       = 1, b = 2, c = 3;
```
Not 
```
int aaaa = 1, bbbb = 2, cccc = 3;
int a    = 1, b    = 2, c    = 3;
```
What do you think?

guy-maurel:
I had the experience of very surprising error with multiple declaration.
Would it be wise to propose a new option to first split such a declaration?

micheleCTDE:
> I had the experience of very surprising error with multiple declaration. 

Interesting, what kind of issue did you run into just for my info. You can drop me an email if you prefer.

>Would it be wise to propose a new option to first split such a declaration?

That's a good idea. Perhaps we can do two things:
1) add option to enable/disable alignment on first assignment (this is quick to do). Default will be false, so no chances to current behavior.
2) add option to split multiple definitions into single definitions over multiple lines (this will probably take some extra work but I am willing to look into it). Default will also be false.

Does it sound good?

micheleCTDE:
[...]

PR #3425, opened 2021-12-12
Fix issue #3422

micheleCTDE:
One test was affected, referring to bug 612, but I am not able to find any reference to what that bug was. I propose we update the test, which is what I have done.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -2579,6 +2579,11 @@ align_assign_func_proto_span    = 0        # unsigned number
 # 0: No limit (default).
 align_assign_thresh             = 0        # number
 
+# Whether to align on the left most assignment when multiple
+# definitions are found on the same line.
+# Depends on 'align_assign_span' and 'align_assign_thresh' settings.
+align_assign_on_multi_var_defs  = false    # true/false
+
 # How to apply align_assign_span to function declaration "assignments", i.e.
 # 'virtual void foo() = 0' or '~foo() = {default|delete}'.
 #
--- a/documentation/htdocs/default.cfg
+++ b/documentation/htdocs/default.cfg
@@ -2579,6 +2579,11 @@ align_assign_func_proto_span    = 0        # unsigned number
 # 0: No limit (default).
 align_assign_thresh             = 0        # number
 
+# Whether to align on the left most assignment when multiple
+# definitions are found on the same line.
+# Depends on 'align_assign_span' and 'align_assign_thresh' settings.
+align_assign_on_multi_var_defs  = false    # true/false
+
 # How to apply align_assign_span to function declaration "assignments", i.e.
 # 'virtual void foo() = 0' or '~foo() = {default|delete}'.
 #
--- a/etc/defaults.cfg
+++ b/etc/defaults.cfg
@@ -2579,6 +2579,11 @@ align_assign_func_proto_span    = 0        # unsigned number
 # 0: No limit (default).
 align_assign_thresh             = 0        # number
 
+# Whether to align on the left most assignment when multiple
+# definitions are found on the same line.
+# Depends on 'align_assign_span' and 'align_assign_thresh' settings.
+align_assign_on_multi_var_defs  = false    # true/false
+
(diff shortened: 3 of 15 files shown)


micheleCTDE:
As an alternative solution, we could add an option to whether align/not align when multiple variable definitions are found on the same line. This should not require a change of the test for bug 612.
@guy-maurel What do you think? Add option or keep PR as is?

gmaurel (changes requested):
NO support here

micheleCTDE:
I tested on my ~3200 files test repo. Mostly good, but I see a regression with virtual functions declarations in 3 files. Will need more work, together with an option to enable/disable this.

micheleCTDE:
I analysed the "egression in more detail.
The difference in those 3 files (out of ~3200 files) is with pointers to function, not with virtual functions.
In fact I would say it is not even a regression but a feature, because before this PR pointers to functions were aligned on return type but not on assignment, as in this example:

Original code:
```
int (*F_recv)(int, void*, unsigned long int, int) = 0L;
int (*F_send)(int, const void*, unsigned long int, int) = 0L;
int (*F_getsockname)(int, struct sockaddr*, ksocklen_t*) = 0L;
int (*F_getpeername)(int, struct sockaddr*, ksocklen_t*) = 0L;
int (*F_accept)(int, struct sockaddr*, ksocklen_t*) = 0L;
int (*F_select)(int, fd_set*, fd_set*, fd_set*, struct timeval*) = 0L;
int (*F_listen)(int, int) = 0L;
int (*F_bind)(int, const struct sockaddr*, ksocklen_t) = 0L;
int a = 0;
double bbbbb = 5.0;
double ccccccccccc = 5.0;
static double mylongclassname::dddddddddddddddd = 5.0;
```

Without this PR:
```
int           (*F_recv)(int, void*, unsigned long int, int) = 0L;
int           (*F_send)(int, const void*, unsigned long int, int) = 0L;
int           (*F_getsockname)(int, struct sockaddr*, ksocklen_t*) = 0L;
int           (*F_getpeername)(int, struct sockaddr*, ksocklen_t*) = 0L;
int           (*F_accept)(int, struct sockaddr*, ksocklen_t*) = 0L;
int           (*F_select)(int, fd_set*, fd_set*, fd_set*, struct timeval*) = 0L;
int           (*F_listen)(int, int) = 0L;
int           (*F_bind)(int, const struct sockaddr*, ksocklen_t) = 0L;
int           a                                 = 0;
double        bbbbb                             = 5.0;
double        ccccccccccc                       = 5.0;
static double mylongclassname::dddddddddddddddd = 5.0;
```

With this PR:
```
int           (*F_recv)(int, void*, unsigned long int, int)                = 0L;
int           (*F_send)(int, const void*, unsigned long int, int)          = 0L;
int           (*F_getsockname)(int, struct sockaddr*, ksocklen_t*)         = 0L;
int           (*F_getpeername)(int, struct sockaddr*, ksocklen_t*)         = 0L;
int           (*F_accept)(int, struct sockaddr*, ksocklen_t*)              = 0L;
int           (*F_select)(int, fd_set*, fd_set*, fd_set*, struct timeval*) = 0L;
int           (*F_listen)(int, int)                                        = 0L;
int           (*F_bind)(int, const struct sockaddr*, ksocklen_t)           = 0L;
int           a                                 = 0;
double        bbbbb                             = 5.0;
double        ccccccccccc                       = 5.0;
static double mylongclassname::dddddddddddddddd = 5.0;
```

You can see the side of the return type gets aligned, but the side of the assignment is not.
I would be inclined on merging this PR as is (or perhaps with an additional enable/disable option) and later work on adding more options for alignment of "pointers to function" enable/disable, alignment on return type side and alignment on assignment side.

What do you think @gmaurel?

gmaurel:
yes

micheleCTDE:
I have added the option 'align_assign_on_multi_var_defs' to allow the user to enable/disable this behavior. 
Default value is false, so no changes in behavior from the original uncrustify code. Changes on the test file for bug 612 are no longer required.
When the option is set to true, assignment will be aligned on the left most assign of a multi variable definition.

gmaurel (approved):

PR #3432, opened 2021-12-15
Add pp_indent_at_level0 option

### LINKED ISSUE #3431: pp_indent_at_level doesn't apply if the brace level is 0
Config file:

```
indent_paren_close = 1
pp_if_indent_code = true
pp_indent_at_level = true
```

Test code:

```
int add(
int a,
int b
#ifdef HAVE_C
, int c
#endif
)
{
int sum = a + b;
#ifdef HAVE_C
sum += c;
#endif
return sum;
}
```

Expected results:

```
int add(
	int a,
	int b
	#ifdef HAVE_C
		, int c
	#endif
       )
{
	int sum = a + b;
	#ifdef HAVE_C
		sum += c;
	#endif
	return sum;
}
```

Actual results:

```
int add(
	int a,
	int b
#ifdef HAVE_C
		, int c
#endif
       )
{
	int sum = a + b;
	#ifdef HAVE_C
		sum += c;
	#endif
	return sum;
}
```

Interestingly, pp_indent_at_level works as expected if the function is wrapped in a namespace.

alexhenrie:
Fixes #3431

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -3152,6 +3152,10 @@ pp_indent                       = ignore   # ignore/add/remove/force/not_defined
 # indented from column 1.
 pp_indent_at_level              = false    # true/false
 
+# Whether to indent #if/#else/#endif at the parenthesis level if the brace
+# level is 0. If false, these are indented from column 1.
+pp_indent_at_level0             = false    # true/false
+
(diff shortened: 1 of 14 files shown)


gmaurel (approved):

alexhenrie:
Thank you!

PR #3437, opened 2021-12-17
Add indent_paren_close=-1 option to preserve original indentation

### LINKED ISSUE #3436: Need indent_paren_close option to preserve original indentation
I need an option to _not_ harmonize the indentation of closing parentheses. In other words, I need Uncrustify to be able to take the following and return it unchanged:

```
#include <stdio.h>

int main(void)
{
	puts("");
	puts(""
	       );
	puts(""
	     );
	puts(""
	    );
	puts(""
	);
	return 0;
}
```

alexhenrie:
Fixes #3436

### DIFF
--- a/src/options.h
+++ b/src/options.h
@@ -1620,10 +1620,11 @@ indent_paren_nl;
 
 // How to indent a close parenthesis after a newline.
 //
-// 0: Indent to body level (default)
-// 1: Align under the open parenthesis
-// 2: Indent to the brace level
-extern BoundedOption<unsigned, 0, 2>
+//  0: Indent to body level (default)
+//  1: Align under the open parenthesis
+//  2: Indent to the brace level
+// -1: Preserve original indentation
+extern BoundedOption<signed, -1, 2>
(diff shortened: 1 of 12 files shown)


src/options.h
```diff
@@ -1623,7 +1623,8 @@ indent_paren_nl;
 // 0: Indent to body level (default)
 // 1: Align under the open parenthesis
 // 2: Indent to the brace level
-extern BoundedOption<unsigned, 0, 2>
+// -1: Preserve original indentation
+extern BoundedOption<signed, -1, 2>
```
gmaurel, line 1627: It would be fine to align the colons

gmaurel (changes requested):

src/options.h
alexhenrie: I have added spaces before the positive numbers so that the colons are aligned.

gmaurel (approved):

alexhenrie:
Thank you!

PR #3447, opened 2021-12-28
Provide space option for consecutive !!

jodyhagins:
The current sp_not option will add space after each !, even when they
are consecutive.  When it is add or force, the following code is
generated:

if (! something) { }
if (! ! something) { }

The new sp_not_not option provides management over consecutive !,
allowing space after any number of ! - without any space in between.

if (! something) { }
if (!! something) { }

### DIFF
--- a/src/options.h
+++ b/src/options.h
@@ -997,6 +997,11 @@ sp_d_array_colon;
 extern Option<iarf_e>
 sp_not; // = IARF_REMOVE
 
+// Add or remove space between two '!' (not) unary operators.
+// If set to ignore, sp_not will be used.
+extern Option<iarf_e>
+sp_not_not; // = IARF_IGNORE
+
 // Add or remove space after the '~' (invert) unary operator.
 extern Option<iarf_e>
 sp_inv; // = IARF_REMOVE
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -2951,6 +2951,12 @@ static iarf_e do_space(chunk_t *first, chunk_t *second, int &min_sp)
 
    if (chunk_is_token(first, CT_NOT))
    {
+      if (  chunk_is_token(second, CT_NOT)
+         && (options::sp_not_not() != IARF_IGNORE))
+      {
+         log_rule("sp_not_not");
+         return(options::sp_not_not());
+      }
       // Add or remove space after the '!' (not) unary operator.
       log_rule("sp_not");
       return(options::sp_not());
--- a/tests/cli/output/universalindent.cfg
+++ b/tests/cli/output/universalindent.cfg
@@ -1855,6 +1855,15 @@ Choices=sp_not=ignore|sp_not=add|sp_not=remove|sp_not=force|sp_not=not_defined
 ChoicesReadable="Ignore Sp Not|Add Sp Not|Remove Sp Not|Force Sp Not"
 ValueDefault=remove
 
+[Sp Not Not]
+Category=1
+Description="<html>Add or remove space between two '!' (not) unary operators.<br/>If set to ignore, sp_not will be used.</html>"
+Enabled=false
+EditorType=multiple
+Choices=sp_not_not=ignore|sp_not_not=add|sp_not_not=remove|sp_not_not=force|sp_not_not=not_defined
+ChoicesReadable="Ignore Sp Not Not|Add Sp Not Not|Remove Sp Not Not|Force Sp Not Not"
+ValueDefault=ignore
+
 [Sp Inv]
 Category=1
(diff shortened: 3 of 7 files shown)


jodyhagins:
The cli_options test is filing in CI - I had hoped it would be passing in CI since it failed on master when I checked it out - before I made any changes.

I had hoped it was failing because of some strange working on my box - and would work under the CI environment.

jodyhagins:
So, it appears that the cli_options test always fails on my checkout of master - but fails in a different way than in CI.  The cli_options output is very painful and time consuming to look through, but it apparently looks for help text for the new option in this file

tests/cli/output/universalindent.cfg

I modified it to match the new option, so hopefully things will pass now.

micheleCTDE:
Hi Jody,
I am not the official maintainer, but I see no point to have an option for double (or multiple) not operators. If there are two not in sequence, it would be better to remove them from the source code in first place!

jodyhagins:
@micheleCTDE  The utilization of consecutive '!' operators has a long history in both C and C++ - for several reasons.

First, a single '!' by itself can some times be difficult to see, especially when smashed next to a word.  The existing sp_not option allows a space after the `!` in that case, to make it `! `

Ideally, one would use 'not' instead of `!` - but microsoft compiler requires a special option - even though this has been part of the C++ standard since the very first one in 1998.

To make it more obvious, some code uses `!!!` to make it more obvious, since that is the same as `!`

Finally, the pattern of `!!` is quite old in both C and C++ to explicitly make something 1 if non-zero or 0 if already zero.

So, there exist patterns of `!!` and `!!!` for many years.

If sp_not is set to add or force, then `!!` ends up `! ! ` and '!!!' ands up `! ! ! `

micheleCTDE:
Personally I disagree, to me it is just a sign of bad coding practice. But I have no intention to start a discussion on who is right and who is wrong.
If Guy decides to merge this PR, fair enough :smile:

jodyhagins:
I'm not saying it is good or bad either - just that the pattern originated in C before C++ was a language - especially `!!` - and there is a good bit of code that uses the pattern.

In this case, there is an existing `sp_not` to add a space after a not operator - but it also adds space between consecutive operators - and that is not desired.  This change does not impact any other existing feature, and I'm quite surprised that it is controversial.

I'm brand new to this particular tool, but the reason I'm playing with it is because it seems to offer a lot of customization choices.

I thought the goal of a tool like uncrustify was to help format existing code to the user's convention/choice - not take a stance on what was or was not a good coding pattern.

micheleCTDE:
>I thought the goal of a tool like uncrustify was to help format existing code to the user's convention/choice - not take a stance on what was or was not a good coding pattern.

Sure, very well said. I didn't mean to start a controversy at all, in fact. Maybe my original post should have been written differently.

jodyhagins:
Well, likely the example could have been better - I doubt I've ever seen it like that inside a conditional - but something a bit more common is like this...

```
    count += !! something;
```

gmaurel:
Please have a look to:
https://github.com/jodyhagins/uncrustify/actions
and activate the workflow, make a little change on your PR, so all could see the results.
Thanks

jodyhagins:
@gmaurel I'm sorry, but I haven't used github a lot, and I don't exactly understand what you are asking.

I went to that link, and clicked on the button to enable workflows - but I'm not sure what "little change on [my] PR" you are asking me to make.

guy-maurel:
**Please help me!**
I am consuting the reference
https://en.cppreference.com
and don't find anything about consecutive !!

gmaurel:
To get a new run one need a change, a new space is enought to get it run again.
The log of the workflow is very usefull for me, before I accept your PR.
Thanks.

jodyhagins:
@gmaurel Consecutive `!` is not a symbol by itself - it is two consecutive separate `!` operators - `! foo` is essentially the same as `! (! foo)`.  It will result in 0 if the value of the operand is 0, and 1 for any other operand value.

Consider...

```
    int i = 42;
    assert((! i) == 0);
    assert((!! i) == 1);
```

It is functionally equivalent to `(foo ? 1 : 0)` and usually appears in code that predates the appearance of a true bool type, though there are still some strong adherents to the methodology. 

For example...

```
    array_of_two_function_pointers[!! some_integer]();
```

will call the function at index 0 is `some_integer` is 0, and the function at index 1 for any other value.

It is also equivalent to `static_cast<bool>(some_value)` in C++ and `(bool)some_value` in C99 - but C++ didn't always have a true bool type, and lots of C code still requires pre-C99.

gmaurel (approved):
I am not happy with that, but it might be usefull for some developpers!

Issue #3448, opened 2021-12-28
indent_class_on_colon overrides indent_ctor_init


jodyhagins:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

When `indent_class_on_colon` is `true` the value in `indent_ctor_init` is ignored.  The indention for the initializer list should prevail here.

For example, given this configuration...

```
indent_with_tabs                = 0
indent_columns                  = 4
indent_continue                 = 4
indent_class                    = true
indent_class_colon              = true
indent_class_on_colon           = true
indent_constr_colon             = true
indent_ctor_init_leading        = 0
indent_ctor_init_following      = 0
indent_ctor_init                = -4
indent_access_spec              = -4
indent_before_class_colon       = -4
nl_class_brace                  = true
nl_class_init_args              = force
nl_class_colon                  = force
nl_constr_init_args             = force
nl_constr_colon                 = force
pos_class_colon                 = lead_force
pos_class_comma                 = lead_force
pos_constr_colon                = lead_force
pos_constr_comma                = lead_force
nl_collapse_empty_body          = true
```

and this input...
```
class Foo : public Bar, private Blarg, private Baz {
[...]

PR #3449, opened 2021-12-28
Allow indent_ctor_init with indent_class_on_colon

jodyhagins:
Previously, indent_ctor_init was ignored when indent_class_on_colon was
true.  This patch will prefer to use indent_ctor_init, and fall back to
indent_class_on_colon.


jodyhagins:
This patch addresses issue #3448

gmaurel (changes requested):
Fine catch.
Please activate the workflow.

gmaurel (approved):

Issue #3446, opened 2021-12-28
nl_func_scope_name is incorrectly applied inside leading decltype


jodyhagins:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

Here is the input file

```
Module::Module() {
int i = foo;
}

namespace detail {
template <typename T>
decltype(std::declval<T &>().put(converts_to_anything, converts_to_anything),
std::true_type())
has_module_api_(PriorityTag<2>);

template <typename T>
auto has_module_api_(PriorityTag<2>)
-> decltype(std::declval<T &>().put(converts_to_anything, converts_to_anything),
std::true_type());
}
```

Here is the generated output

```
Module::
Module()
{
    int i = foo;
}

namespace detail {
template <typename T>
decltype(std::
    declval<T &>().put(converts_to_anything, converts_to_anything),
    std::
    true_type())
has_module_api_(PriorityTag<2>);

template <typename T>
auto has_module_api_(PriorityTag<2>)
-> decltype(std::declval<T &>().put(converts_to_anything, converts_to_anything),
    std::true_type());
} // namespace detail
```

[...]

PR #3451, opened 2021-12-30
Resolves Issue #3446

jodyhagins:
The code for tagging chunks as being IN_DECLTYPE stopped short in
cases where all chunks had the same level.

This change does two things.

First, it modifies the code that tags chunks with the IN_DECLTYPE
flag, by manually searching for the matching closing parenthesis.

Second, it does not honor nl_func_scope_name if the chunk has
the IN_DECLTYPE flag set.


gmaurel (approved):
Fine!

gmaurel:
Add a line such as:
_fixes #3446_ 
to automatically close the issue

Issue #3414, opened 2021-12-07
Crash with one-liner struct


chris-mich:
Config: 
```
# Uncrustify-0.74 (win64 official version)
align_func_proto_span    = 1
align_single_line_func   = true
align_single_line_brace  = true
nl_func_leave_one_liners = true
```
Input: 
```
namespace
{
struct S { void f() {} };
}
```

Uncrustify will crash trying to process this code with this specific sets of parameters.

jodyhagins:
The same change suggested for #3413 fixes this crash as well.  Here is the generated output

```
namespace
{
struct S { void f() {} };
}
```

If you think this resolves both of these issues, I'll submit a pull request - though it would be nice to verify more than just that it no longer crashes.

However, all the existing tests still pass with this change, so I assume it will be ok - at least it is better than crashing...

Issue #3413, opened 2021-12-07
Crash when specifying = {} as default argument in CPP file


chris-mich:
Config:
```
# Uncrustify-0.74 (win64 official version)
align_func_proto_span   = 1
align_single_line_func  = true
align_single_line_brace = true
```

Input:
```
void FuncCrash( int a = {} )
{
}
```

"= {}"  makes Uncrustify crash with this specific set of parameters in the config file.

jodyhagins:
Hi Chris.  I took a peek at this.  Preventing the crash is quite simple, but I'm not sure if the generated output is what would be expected with this input.

Even better would be a much richer set of input to test with to make sure that not only does it no longer crash, but it actually generates the correct output.

With this diff...

```
diff --git a/src/align_func_proto.cpp b/src/align_func_proto.cpp
index b3aca3b8..d08c03d3 100644
--- a/src/align_func_proto.cpp
+++ b/src/align_func_proto.cpp
@@ -185,6 +185,14 @@ void align_func_proto(size_t span)
               && pc->flags.test(PCF_ONE_LINER))
       {
          AlignStack *stack_at_l_bl_brace = many_as_brace.at(pc->level).at(pc->brace_level);
+         if (stack_at_l_bl_brace == nullptr)
+         {
+            stack_at_l_bl_brace = new AlignStack();
+            stack_at_l_bl_brace->Start(myspan, mythresh);
+            stack_at_l_bl_brace->m_gap = mybr_gap;
+            many_as_brace.at(pc->level).at(pc->brace_level) = stack_at_l_bl_brace;
+         }
+         stack_at_l_bl_brace->Debug();
          stack_at_l_bl_brace->Add(pc);
          look_bro = false;
       }
```

And this input...

```
void FuncCrash(int a = {}) { }
void FuncCrash(int b = int{}) { }
void FuncCrash(int b = int(0)) { }
void FuncCrash(int b = double{0}) { }
void FuncCrash(int b = 0) { }
```

I see this output

```
void FuncCrash(int a =       {}) {
}
void FuncCrash(int b = int   {}) {
}
void FuncCrash(int b = int(0)) {
}
void FuncCrash(int b = double{0}) {
}
void FuncCrash(int b = 0) {
}
```


At least it no longer crashes, but I'm not familiar at all with the alignment code, so I'm not sure this is correct.

PR #3452, opened 2021-12-30
Stops crash reported in Issues #3413 and #3414

jodyhagins:
A stack object was assumed to exist in a vector of stack pointers.

The fix is to create a new stack object when needed.

Note that this does prevent the crash, but I'm not 100% sure that
the generated output is what would be expected.


gmaurel (changes requested):
fine again
Add the line
fixes #3413
fixes #3414 to close the issues

jodyhagins:
Apologies again for not knowing how these github things interact.

Where, exactly, do I put those "fixes" lines - in the commit message for this ticket, or some place else?

jodyhagins:
OK, I added 

fixes #3413
fixes #3414 

to the commit message - I hope that is what you were talking about.

gmaurel (approved):

PR #3461, opened 2022-01-06
chunk_t code improvement - part 1

micheleCTDE:
This PR is intended as the first step in a code improvement effort around chunk_t. The idea is to make use of C++ object oriented capabilities to change chunk_t into a full fledged object called Chunk. This would allow to write code like
```
               if (  ((next = pc->getNext()) != nullptr)
                  && next->isComment())
...
      if (  pc->is(CT_BOOL)
         || pc->is(CT_QUESTION)
         || pc->is(CT_COND_COLON)
         || pc->is(CT_COMMA))
```
instead of
```
               if (  ((next = chunk_get_next(pc)) != nullptr)
                  && chunk_is_comment(next))
...
      if (  chunk_is_token(pc, CT_BOOL)
         || chunk_is_token(pc, CT_QUESTION)
         || chunk_is_token(pc, CT_COND_COLON)
         || chunk_is_token(pc, CT_COMMA))
```
and so on.

This is going to require a lot of changes, so rather than doing a massive single PR that would be difficult to check and keep updated as other changes are made, I prefer to break it down in smaller bites and get it merged progressively.

What do you think?


gmaurel:
Fine.

gmaurel (approved):
Fine work!

PR #3463, opened 2022-01-09
Feat/chunk improvements 3

micheleCTDE:
1. Some improvement on the list manager
2. Added NullChunk and NullChunkPtr, together with supporting calls.
3. Replaced chunk_get_prev() with Chunk::get_prev()

NOTES:
1. while the code improvement process is on going, there will be a mix of nullptr and Chunk::NullChunkPtr in the code as well as a mix of checking for nullptr and whether a chunk is null or not. As we progress, there should be more and more checks for null chunk and less for nullptr. Ultimately when the effort is completed there should not be any nullptr chunk pointers, only NullChunkPtr.
2. chunk_get_prev() has been fully replaced by Chunk::get_prev() except for the call in select_search_fct(), where for the time being it is used a modified version called __chunk_get_prev(). This this now an internal function (removed from chunk.h) and will be ultimately removed in full when we reach the stage where select_search_fct() definition can be changed too.

### DIFF
--- a/src/ListManager.h
+++ b/src/ListManager.h
@@ -29,15 +29,15 @@ private:
    // Hide copy constructor
    ListManager(const ListManager &ref)
    {
-      first = NULL;
-      last  = NULL;
+      first = nullptr;
+      last  = nullptr;
    }
 
 public:
    ListManager()
    {
-      first = NULL;
-      last  = NULL;
+      first = nullptr;
+      last  = nullptr;
    }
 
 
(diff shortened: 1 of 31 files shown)


micheleCTDE:
Forgot to mention that this PR has been tested on my usual ~3200 files test repo and not a single byte has changed in the resulting output :smile:

gmaurel (approved):
Fine!

PR #3468, opened 2022-01-11
some more return

guy-maurel:
The function _do_symbol_check_ have many IF-blocks.
If a IF is _true_, some flags and/or type are set.
After that, it is not necessary to go on untill the end of the function, one can _return_ and go on.
@micheleCTDE : could you test it?


micheleCTDE:
Hi  @guy-maurel,
sorry for the late reply. Tested on my usual repo and no changes whatsoever. Looks good.

PR #3477, opened 2022-01-17
Add indent_ignore_arith option

### LINKED ISSUE #3476: Need option to accept any arithmetic operator indentation
Similar to issue #3472, I need an option to ignore indentation of arithmetic operators.

alexhenrie:
Fixes #3476


gmaurel (changes requested):
Do not declare more than one variable per line of code.

Look my comment.

alexhenrie:
What existing option would we add an "ignore" value to? There is no `indent_arith` option.

I am also going to need an option to ignore indentation on lines that start with a shift operator `<<` or `>>` (unless you are OK with lumping them in with the arithmetic operators), as well as an option to ignore indentation on lines that start with a semicolon.

An option to ignore the indentation of all continuation lines would be ideal and would remove most or all of the need for options to ignore indentation on lines that start with an operator. However, I don't see any way to do that without making deep and major changes to Uncrustify's architecture.

micheleCTDE:
> No. We are on, trying to change the code and options to get a NOCHANGE.cfg configuration file. The way you are going on is to add some options, to modify options. We have just now 237 bool options. This would need so many new options.
> 
> It is not the way I preffer.
> 
> Instead, I would like to keep the option, but add a possible value (ignore) I am just working on that for the spaces.
> 
> One could change the bool-options into a three values options with the values: 0: false (already exist) 1: true 2: ignore
> 
> Please, comment!

Hi @guy-maurel,
I meant to comment this morning but got busy with other stuff. 
I think adding 'ignore' as possible value to bool options if the best way, so the user has the chance to decide to use or ignore that option. This should also help to move towards a neutral config, as we discussed in the past.
It is not a small task, because currently space/tab are ignored when uncrustify is tokenizing and classifying the various chunks (unless I missed something). We will need to keep existing intentation/spacing into consideration in the chunk list and then use it or ignore it based on the various options settings. One option could work as global enable/disable, so if for example we choose to globally ignore original spacing, there is no need to remember that in the chunk list and uncrustify would work like now.

@alexhenrie 
I think the above task will take quite some time, so perhaps if you need a specific ignore option and it is not approve, you can add it locally to your build as a pre-patch. Or you can add it to your local fork.

Regardless of what is the final choice, the work I am currently doing on code improvements will not affect it, since it is purely targeted at reworking code without changing any functionality.

gmaurel (approved):

gmaurel:
I now anderstand what you have proposed.
Thanks

alexhenrie:
Thank you!

PR #3479, opened 2022-01-19
Code improvements/get_next_nl-get_prev_nl

micheleCTDE:
Currently we need an additional check for nullptr after the search and return NullChunkPtr if necessary.
When all functions will have been converted and the search function can be modified, then we can remove this extra check.

Also note the two functions are not const yet. To do that we need to make a lot of other functions const, but again I am proceeding step by step and not trying to do too many things at once, which would result in a big mess.

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -1270,12 +1270,12 @@ static void move_case_return(void)
             {
                // This may indicate a semicolon was missing in the code to format.
                // Avoid moving the return statement to prevent potential unwanted erros.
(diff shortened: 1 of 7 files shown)


gmaurel:
@micheleCTDE : Please, have a look again at the conflict.

micheleCTDE:
Conflict was caused by merge of other PRs. Solved and branch rebased on current master.

gmaurel (approved):

PR #3492, opened 2022-01-31
Add indent_shift=-1 option to preserve original indentation

### LINKED ISSUE #3480: Need option to accept any bit-shift operator indentation
Similar to issues #3472 and #3476, I need an option to ignore indentation of the shift operators `<<` and `>>`.

alexhenrie:
Fixes #3480


gmaurel (changes requested):
If there is a not correct line in the configuration file, i.e.
_indent_shift = xxx_

we don't get any error message.

gmaurel (approved):
Fine, we are on the way to get a zero-config, thanks

Issue #3525, opened 2022-02-12
Files with .sql extension are detected as C rather than SQL


alexhenrie:
If I have a file named test.sql, I get:

```
do_source_file(1507): Parsing: test.sql as language C
```

The `.sql` extension is for pure SQL. Uncrustify already recognizes a separate `.sqc` extension for SQL embedded in C code.

Issue #3524, opened 2022-02-12
Files with .js extension are detected as C rather than ECMA


alexhenrie:
If I have a file named test.js, I get:

```
do_source_file(1507): Parsing: test.js as language C
```

Since JavaScript is a superset of ECMAScript, selecting ECMA by default would be much more appropriate than selecting C.

Issue #3523, opened 2022-02-12
Files with uppercase .C extension are detected as C++


alexhenrie:
If I have a file named TEST.C, I expect:

```
do_source_file(1507): Parsing: TEST.C as language C
```

But I actually get:

```
do_source_file(1507): Parsing: TEST.C as language CPP
```

PR #3526, opened 2022-02-12
Clean up language_exts array 

### LINKED ISSUE #3522: Files with .cc extension are detected as C++, but files with .hh extension are detected as C
Config file: Empty

Original code (save to file named `test.hh`):

```
class Foo : public Bar
{
};
```

Expected result: No change

Actual result:

```
class Foo: public Bar
{
};
```

(Look at the space before the colon.)

If the file is named `test.hpp` instead, there is no problem.

The `.hh` extension has come up before in issue #2762, but the bug reporter was content to specify `-l cpp` on the command line. However, since Uncrustify already recognizes `.cc` as a C++ extension, it ought to default to C++ for `.hh` as well.

alexhenrie:
Fixes #3522, #3523, #3524, and #3525


gmaurel (approved):
fine

PR #3528, opened 2022-02-13
Update appveyor build to Visual Studio 2019

micheleCTDE:
Update appveyor to VS 2019.
The builds with VS 2015+2017 are approaching 1h 20m, so probably wiser to keep a single version rather than 2.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,9 +3,9 @@ platform: x64
 environment:
   matrix:
   # MSVC
-  - GENERATOR: Visual Studio 15 2017 Win64
-    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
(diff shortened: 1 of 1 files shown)


micheleCTDE:
The build times with VS 2019 seems huuuge. Debug mode is twice the time taken for VS 2017. Not sure it is an unlucky case (slow builder) but I wonder if the update to VS 2019 makes sense at this stage.

@guy-maurel what do you think? Perhaps it would be better to simply drop VS 2015 and still with VS2017 to half the build time.

micheleCTDE:
> yes, VS2017 wouls be enought.

Ok, I will close this PR without merging and create a separate PR to drop VS 2015 and keep only VS 2017.
Thanks for the feedback.

PR #3542, opened 2022-02-17
Capitalize and rename c_token_e to T_Token

micheleCTDE:
I was unsure whether we should rename it T_Type rather than T_Token since it is mostly used in compination with a chunk type.
What do you think @guy-maurel?


gmaurel (approved):
yes, it is fine

PR #3547, opened 2022-02-25
Add indent_ignore_first_continue option

### LINKED ISSUE #3546: Option to take indendation from first continuation line instead of from indent_continue
I'm looking for an option to ignore `indent_continue` (or `indent_continue_class_head`) and take the indentation from the first continuation line.

For example, the following input:

```
#include <stdio.h>

class
      alignas(32)
                  Foobar;

int main()
{
    typedef
            int
                int32;
    int foo,
             bar,
                  baz;
    foo = 5
            + 6
                + 7;
    if (printf("%d %d",
                        5
                          + 6,
                               7)
                                  < 0)
    {
        return 1;
    }

    return 0;
}
```

Would become the following output:

```
#include <stdio.h>

class
      alignas(32)
      Foobar;

int main()
{
    typedef
            int
            int32;
    int foo,
             bar,
             baz;
    foo = 5
            + 6
            + 7;
    if (printf("%d %d",
                        5
                        + 6,
                        7)
                                  < 0)
    {
        return 1;
    }

    return 0;
}
```

In other words, I'd like an option to allow any continuation indent, as long as it's consistent.

alexhenrie:
Fixes #3546

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1057,15 +1057,21 @@ force_tab_after_define          = false    # true/false
 # Default: 8
 indent_columns                  = 8        # unsigned number
 
+# Whether to ignore indent for the first continuation line. Subsequent
+# continuation lines will still be indented to match the first.
+indent_ignore_first_continue    = false    # true/false
+
 # The continuation indent. If non-zero, this overrides the indent of '(', '['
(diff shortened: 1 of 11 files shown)


gmaurel (approved):

alexhenrie:
Thank you!

PR #3549, opened 2022-02-25
Add indent_ignore_bool option

### LINKED ISSUE #3548: Option to preserve indentation of boolean operators outside of parentheses
Uncrustify has `indent_bool_paren=-1` to ignore indentation of `&&` and `||` when they are inside of parentheses. However, I also need an option to ignore their indentation when they are not in parentheses, for example:

```
{
	return argc < 3
	    || argc > 10;
}
```

Uncrustify unconditionally changes that to:

```
{
	return argc < 3
	       || argc > 10;
}
```

alexhenrie:
Fixes #3548


gmaurel (approved):

alexhenrie:
Thank you!

PR #3553, opened 2022-02-28
Don't try to pop the class header frame twice

### LINKED ISSUE #3552: indent_continue_class_head causes incorrect indentation of classes inside namespaces
Config file:

```
indent_with_tabs           = 0
indent_columns             = 4
indent_namespace           = true
indent_class               = true
indent_access_spec         = -4
indent_continue_class_head = 2
```

Original code:

```
namespace Salads
{
class
Waldorf
{
public:
int size;
};
}
```

Expected results:

```
namespace Salads
{
    class
      Waldorf
    {
    public:
        int size;
    };
}
```

Actual results:

```
namespace Salads
{
    class
  Waldorf
    {
public:
    int size;
    };
}
```

alexhenrie:
The frame is already popped on [line 1364](https://github.com/uncrustify/uncrustify/blob/639aed0ae591afee620688e4951815a01eb03d9e/src/indent.cpp#L1364).

Fixes #3552


gmaurel (approved):
fine catch

PR #3555, opened 2022-03-01
Optimize include/import sorting speed

kwongius:
The current include/import sorting is a bit slow for files with a large number of includes (100+). There are two primary factors causing this slowness. The first is that the comparison is a bit expensive in particular due to repeated regex matching. The second is the algorithm. The current implementation makes use of a selection sort in order to minimize the number of swaps but means a best case runtime performance of *O(n)*.

This change implements memoization of the regex lookups to improve comparison performance. This change also migrates the sorting implementation to use `std::sort()` which offers best case runtime performance of *O(n log n)*. Upon sorting values, the actual swaps are performed so the worst case number of swaps is *O(n)*.

With a large number of includes and imports (tested with 300+) I was observing execution time moving from about 100 seconds down to about 2 seconds.


src/sorting.cpp
```diff
@@ -49,13 +52,11 @@ include_category *include_categories[kIncludeCategoriesCount];
  * @retval  > 0
  * @retval  < 0
  */
-static int compare_chunks(Chunk *pc1, Chunk *pc2, bool tcare = false);
+static int compare_chunks(Chunk *pc1, Chunk *pc2, bool tcare);
```
gmaurel, line 55: why do you suppress the default value?

gmaurel (changes requested):
see the comment

src/sorting.cpp
kwongius: `compare_chunks()` is only ever called once and the parameter is always passed so the default value is never used.

gmaurel (approved):

gmaurel (changes requested):
OK for the default value,
BUT, I am missing an/many example(s) with cpp and cfg files to check your proposal.
An example could be uncrustify.cpp (the beginning)

What is about the option _mod_sort_incl_import_grouping_enabled_ ?

kwongius:
https://github.com/uncrustify/uncrustify/commit/27b33d779e7aa33c1ebc39a52e54c3a95bc153d4 introduced the `mod_sort_incl_import_grouping_enabled` option which is used in several places. As I understand it, the previous behavior prior to the flag being added didn't care about the order within the `chunks` array so it was a small optimization to avoid doing the swap. Always doing the swap is the logically correct behavior anyway. The previous behavior already swapped a maximum of O(N) times so this is a micro-optimization for large number of imports. However, with this specific PR, we need to swap order of the `chunks` held within the array in order to keep track of the swaps we make within the document itself using `chunk_swap_lines()`.

This change is purely about optimizing the behavior so the existing tests already cover this (02325, 39001, 50030-50035 are good examples). I don't see any kind of performance or runtime test infrastructure which is what would be best for this specific PR. Let me know if that's incorrect or if the existing tests are otherwise insufficient.

gmaurel:
Please, have a look to <uncrustify>/tests/cli/test_cli_options.py
It is a good place to add a test.

guy-maurel:
What is about the option _mod_sort_incl_import_grouping_enabled_ ?

gmaurel (approved):

PR #3564, opened 2022-03-02
doc: add --prefix option to git archive command in release-process.rst

### LINKED ISSUE #3563: tar file should contain the version name as the root folder
When creating the TAR file for a release, the "git archive" command needs the "--prefix" option to properly add the package name and version as the root folder.  That seems to be the standard these days.

Example command:
```
git archive -o uncrustify-0.75.0.tar.gz --prefix=uncrustify-0.75.0/ uncrustify-0.75.0
```

As near as I can tell, this is a manual step, mentioned in release-process.rst.

bengardner:
This puts the code under uncrustify-$(VERSION).
Example:
  uncrustify-0.74.0/README.md

Currently, README.md would be in the root of the tar.

Fixes #3563

### DIFF
--- a/release-process.rst
+++ b/release-process.rst
@@ -234,7 +234,7 @@ Create a tarball:
 .. code::
 
    $ cd /path/to/uncrustify
-   $ git archive -o uncrustify-0.1.2.tar.gz uncrustify-0.1.2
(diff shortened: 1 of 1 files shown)


gmaurel (approved):
Fine to see you again

PR #3571, opened 2022-03-03
Apply indent_paren_after_func_def to class constructors

### LINKED ISSUE #3570: indent_paren_after_func_def doesn't work on class constructors
Config file:

```
indent_class                 = true
indent_paren_after_func_def  = true
indent_paren_after_func_decl = true
```

Original code:

```
class Foo
{
public:
Foo
(
)
{
}

Foo
(
int x
);

void init
(
)
{
}

void init
(
int x
);
};
```

Expected results:

```
class Foo
{
public:
	Foo
	        (
	        )
	{
	}

	Foo
	        (
	        int x
	        );

	void init
	        (
	        )
	{
	}

	void init
	        (
	        int x
	        );
};
```

Actual results:

```
class Foo
{
public:
	Foo
	(
	)
	{
	}

	Foo
	        (
	        int x
	        );

	void init
	        (
	        )
	{
	}

	void init
	        (
	        int x
	        );
};
```

This example shows that `indent_paren_after_func_decl` already works on class functions and constructors, and `indent_paren_after_func_def` also works on class functions, but it does not work on constructors.

alexhenrie:
Fixes #3570

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -2862,7 +2862,8 @@ void indent_text(void)
             && (  (  (  get_chunk_parent_type(pc) == CT_FUNC_PROTO
                      || get_chunk_parent_type(pc) == CT_FUNC_CLASS_PROTO)
                   && options::indent_paren_after_func_decl())
-               || (  get_chunk_parent_type(pc) == CT_FUNC_DEF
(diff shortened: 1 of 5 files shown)


alexhenrie:
Thank you!

PR #3577, opened 2022-03-07
Reset vardefcol after the class header

### LINKED ISSUE #3576: use_indent_continue_only_once breaks indent_paren_after_func_def in C++ subclasses
Config file:

```
indent_paren_after_func_def   = true
use_indent_continue_only_once = true
```

Original code:

```
class Foo : public Bar
{
public:
void doIt
(
)
{
}
};
```

Expected results:

```
class Foo : public Bar
{
public:
void doIt
        (
        )
{
}
};
```

Actual results: No change

However, it works as expected if you remove `: public Bar` from the class.

This bug is similar to #3567 and #3570.

alexhenrie:
Fixes #3576

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -3551,7 +3551,8 @@ void indent_text(void)
 
       if (  chunk_is_semicolon(pc)
          || (  chunk_is_token(pc, CT_BRACE_OPEN)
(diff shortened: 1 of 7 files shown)


alexhenrie:
Thank you!

PR #3581, opened 2022-03-09
Add indent_ignore_case_brace option

### LINKED ISSUE #3580: Option to preserve indentation of braced switch case
I'd like an option that doesn't change the indentation of the braces in the following example:

```
#include <stdio.h>

{
    switch (argc)
    {
    case -1:
  {
      printf("Negative args shouldn't be possible\n");
      break;
  }
    case 0:
    {
      printf("Zero args can happen but shouldn't\n");
      break;
    }
    case 1:
      {
          printf("One arg\n");
          break;
      }
    case 2:
        {
            printf("Two args\n");
            break;
        }
    default:
          {
              printf("%i args\n", argc);
              break;
          }
    }

    return 0;
}
```

alexhenrie:
Fixes #3580

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1294,6 +1294,9 @@ indent_switch_case              = 0        # unsigned number
 # Usually the same as indent_columns or indent_switch_case.
 indent_switch_body              = 0        # unsigned number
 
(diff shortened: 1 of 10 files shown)


gmaurel (approved):

alexhenrie:
Thank you!

PR #3583, opened 2022-03-10
Add pp_indent_brace=-1 option to preserve original indentation

### LINKED ISSUE #3582: Option to preserve indentation of braced preprocessor block
I could use an option similar to `indent_ignore_case_brace` but for preprocessor blocks. The general idea is that some people want `pp_indent_brace=true` and others want `pp_indent_brace=false`, so we want Uncrustify to accept either.

alexhenrie:
Fixes #3582


gmaurel (approved):

alexhenrie:
Thank you!

PR #3590, opened 2022-03-14
Add mod_full_brace_if_chain=2 option

### LINKED ISSUE #3589: Option to add braces to all if/else branches if any branch already has them
Consider the following:

```
#include <stdio.h>

void is_big(int x)
{
	if (x >= 100) {
		printf("x is big\n");
	} else {
		printf("x is not big\n");
	}
}

void is_small(int x)
{
	if (x <= 100)
		printf("x is small\n");
	else
		printf("x is not small\n");
}

void is_medium(int x)
{
	if (x > 100 && x < 100) {
		printf("x is medium\n");
	} else
		printf("x is not medium\n");
}
```

I want to ensure that each if/else chain either consistently has braces or consistently does not have braces. Uncrustify has a few options for that already:

* `mod_full_brace_if=add` adds braces to all branches
* `mod_full_brace_if=remove` removes braces from all branches
* `mod_full_brace_if_chain=true` removes braces from all branches if no branch needs them and adds them to all branches if any branch needs them

`mod_full_brace_if_chain` is close to what I'm looking for, but I want something that's based on whether the branch already _has_ braces, not whether it strictly _needs_ them. So rather than removing the braces from both `is_big` and `is_medium`, I want Uncrustify to just add more braces to `is_medium`:

```
#include <stdio.h>

void is_big(int x)
{
	if (x >= 100) {
		printf("x is big\n");
	} else {
		printf("x is not big\n");
	}
}

void is_small(int x)
{
	if (x <= 100)
		printf("x is small\n");
	else
		printf("x is not small\n");
}

void is_medium(int x)
{
	if (x > 100 && x < 100) {
		printf("x is medium\n");
	} else {
		printf("x is not medium\n");
	}
}
```

alexhenrie:
Fixes #3589


src/braces.cpp
```diff
@@ -1552,7 +1552,7 @@ static void process_if_chain(Chunk *br_start)
                  __func__, __LINE__, braces.size(), pc->orig_line, tmp ? "" : "not",
                  get_token_name(pc->type));
 
-         if (!tmp)
+         if (!tmp || options::mod_full_brace_if_chain() == 2)
```
gmaurel, line 1555: please, put the second part of the line on the next.
```.cpp
if (  !tmp
   || options::mod_full_brace_if_chain() == 2)
```

gmaurel (changes requested):
have a look on my proposal

src/braces.cpp
alexhenrie: Done

gmaurel (approved):
Thanks

PR #3593, opened 2022-03-16
Add mod_full_brace_if_chain=3 option

alexhenrie:
`mod_full_brace_if_chain=3` is just like `mod_full_brace_if_chain=1` except that it only applies if the original bracing was inconsistent.


micheleCTDEAdmin (approved):
Looks good, thanks Alex.

alexhenrie:
Thank you!

PR #3610, opened 2022-04-11
Add indent_namespace_inner_only

dbartolini:
Adds the ability to indent inner namespaces only (namespaces nested inside other namespaces).

The following:
```
namespace out {
int i;
namespace in {
int i;
}
}
```

with ``indent_namespace_single_indent=2``, becomes:
```
namespace out {
int i;
namespace in {
  int i;
}
}
```
AFAICT there was no option to format namespaces in this way.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1147,6 +1147,10 @@ indent_namespace_level          = 0        # unsigned number
 # indented. Requires indent_namespace=true. 0 means no limit.
 indent_namespace_limit          = 0        # unsigned number
 
+# Whether to indent only in inner namespaces (nested in other namespaces).
+# Requires indent_namespace=true.
+indent_namespace_inner_only     = false    # true/false
+
 # Whether the 'extern "C"' body is indented.
 indent_extern                   = false    # true/false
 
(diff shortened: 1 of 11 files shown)


tests/expected/cpp/60087-indent_namespace_single_indent.cpp
```diff
@@ -0,0 +1,8 @@
+namespace out
+{
+	int i;
+namespace in
```
micheleCTDE, line 4: This does not seem right. The existing 'indent_namespace_single_indent' is used to control the indentation when multiple consecutive namespaces are found. If the parameter was true, the first namespace would be indented as per current indent level, while subsequent namespaces would use the same indent level as the first. 
You can look at the test file '30914-indent_namespace_single_indent.h' (and relative input and config) to see how that parameter works.
If you run this example with the config file 'indent_namespace_single_indent.cfg', you will get this output: 
```
namespace out
{
	int i;
        namespace in
        {
	        int i;
        }
}
```

I guess what you wanted to do is a bit different, that is controlling whether to indent the code inside a namespace or not. Problably wiser to use a new separate option for that.

micheleCTDEAdmin (approved):

micheleCTDE (approved):

PR #3613, opened 2022-04-14
Fix cast detection in C# and Vala

dbartolini:
Fixes cast detection in C# and Vala when non-trivial type identifiers are involved.


micheleCTDEAdmin (approved):

micheleCTDE:
Thanks Daniele!

PR #3617, opened 2022-04-15
Fix C++ c-style cast detection with template types

dbartolini:
Fixes c-style cast detection with template types in C++.


micheleCTDEAdmin (approved):
Looks good!

PR #3626, opened 2022-04-22
Replace chunk_is_comment with its more modern variant IsComment

dundargoc:
Also remove chunk_is_comment function as specified in the TODO.

### DIFF
--- a/src/align_func_params.cpp
+++ b/src/align_func_params.cpp
@@ -124,7 +124,7 @@ Chunk *align_func_param(Chunk *start)
       }
       else if (comma_count > 0)
       {
-         if (!chunk_is_comment(pc))
+         if (!pc->IsComment())
          {
             comma_count = 2;
             break;
--- a/src/align_tools.cpp
+++ b/src/align_tools.cpp
@@ -72,7 +72,7 @@ Chunk *scan_ib_line(Chunk *start, bool first_pass)
       Chunk *next = pc->GetNext();
 
       if (  next->IsNullChunk()
-         || chunk_is_comment(next))
+         || next->IsComment())
       {
          // do nothing
       }
--- a/src/align_var_def_brace.cpp
+++ b/src/align_var_def_brace.cpp
@@ -130,7 +130,7 @@ Chunk *align_var_def_brace(Chunk *start, size_t span, size_t *p_nl_count)
                  __func__, __LINE__, pc->orig_line, pc->orig_col, pc->Text(), get_token_name(pc->type));
       }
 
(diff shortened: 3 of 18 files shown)


micheleCTDE:
@dundargoc 
thanks for the PR. I am a bit surprised not to see any nullptr related replacement, because nullptr and null chuck are not the same and usually during the replacement process it is often necessary to replace some other code.
Just wondering if you have checked that and all is fine or if you simply were not aware of it.

dundargoc:
I see. I was not aware of this, my bad.

Do you have a quick example on what type of change might be necessary? I'll take a look at previous changes as well.

dundargoc:
Oh, you mean instead of replacing `if(chunk_is_comment(pc))` to `if(pc->IsComment())` I should change it to `if(pc != nullptr && pc->IsComment())`? 

Sure, I can do that. Wonder if there isn't a less verbose/repetitive way of solving this though.

micheleCTDE:
You can look for example at this [commit](https://github.com/uncrustify/uncrustify/commit/c5e83197c8f33ab9244355ad4a53cea73a0f94bd).
You will see in several places there are replacements of nullptr with checks for null chunk or not null chunk. Depending on where this is in the code, there may be further changes required in while/for loops or in other functions that get called on the way.

Although all the tests have passed, when doing the replacement I usually also run a comparison test on about 3200 source files to make sure the changes don't introduce regressions or different behavior. You can see for example that once I forgot to do that and I had to fix up some broken stuff after that in this other [commit](https://github.com/uncrustify/uncrustify/commit/c05ccdbd7e213c10f7b7f68763ee8dfa3c22fb3c) (see PR #3608).

micheleCTDE:
> f(pc != nullptr && pc->IsComment())

This depends on where it is in the code. Sometimes it may be necessary (if it is not known whether pc could be a nullptr or not), sometimes it is not necessary (if pc is known to be either a valid pointer or a null chunk). Again, context is the key when that is needed or not.

The whole thing behind the code improvement effort and the null chuck is to simplify the code and to avoid having to directly checking for nullptr each time. During the transition it may still be necessary to check for nullptr, but in the end this won't be required. So far we went from >1300 nullptr instances to around 900 as of today. Still quite a fair bit of work to do :-)

micheleCTDEAdmin (approved):

micheleCTDE:
I have decided to merge the PR as is, since in any case it is useful and something that sooner or later should have been done. I will double check if any nullptr replacement is required and push that as a separate commit.

PR #3635, opened 2022-04-29
Make chunk_is_str accept a string instead of a char array.

dundargoc:
This will prevent any length-counting errors since it won't be done
manually anymore.


dundargoc:
Ooops, created PR from master.

PR #3636, opened 2022-04-29
Calculate string length in chunk_is_str instead of manually specifying it.

dundargoc:
This will prevent any length-counting errors since it won't be done
manually anymore.

### DIFF
--- a/src/chunk.h
+++ b/src/chunk.h
@@ -985,8 +985,10 @@ static inline bool chunk_is_type(Chunk *pc)
 }
 
 
-static inline bool chunk_is_str(Chunk *pc, const char *str, size_t len)
+static inline bool chunk_is_str(Chunk *pc, const char *str)
 {
+   size_t len = strlen(str);
+
    return(  pc != nullptr                         // valid pc pointer
          && (pc->Len() == len)                    // token size equals size parameter
          && (memcmp(pc->Text(), str, len) == 0)); // token name is the same as str parameter
--- a/src/align_eigen_comma_init.cpp
+++ b/src/align_eigen_comma_init.cpp
@@ -73,7 +73,7 @@ void align_eigen_comma_init(void)
       }
       else if (  !pc->flags.test(PCF_IN_ENUM)
               && !pc->flags.test(PCF_IN_TYPEDEF)
-              && chunk_is_str(pc, "<<", 2))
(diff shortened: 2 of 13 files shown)


src/chunk.h
```diff
@@ -985,11 +985,13 @@ static inline bool chunk_is_type(Chunk *pc)
 }
 
 
-static inline bool chunk_is_str(Chunk *pc, const char *str, size_t len)
+static inline bool chunk_is_str(Chunk *pc, const string &str)
```
micheleCTDE, line 988: We can keep using "const char *" for the string, so we avoid unnecessary conversion to string objects.
We can determine the length on the string with strlen.

src/chunk.h
```diff
@@ -985,11 +985,13 @@ static inline bool chunk_is_type(Chunk *pc)
...
 {
-   return(  pc != nullptr                         // valid pc pointer
-         && (pc->Len() == len)                    // token size equals size parameter
-         && (memcmp(pc->Text(), str, len) == 0)); // token name is the same as str parameter
+   auto len = str.size();
```
micheleCTDE, line 990: Can use strlen on a const char pointer.
Possibly avoid using "auto": although it is easier, it makes the code less readable (one has to think about the type...)

micheleCTDE (changes requested):
The idea is good, please amend as per comments below.

src/chunk.h
dundargoc: Good point.

src/chunk.h
dundargoc: Sure. I'll need to adjust the commit message.

dundargoc: In hindsight, maybe we should just use a string and get rid of the `memcmp` which is just a goofy C-ism at this point?

micheleCTDE: There is nothing wrong in using C functions.... definitely nothing goofy in it :-) We could perhaps replace ```memcmp``` with strncmp, I missed that on the frist review, but either way would work fine.

micheleCTDEAdmin (approved):
Looks good now, thanks for updating as requested

micheleCTDE (approved):

src/chunk.h
micheleCTDE: neovim rocks btw :wink:

PR #3637, opened 2022-04-29
Move Objective-C tests to the correct 'oc' folder

dbartolini:
Moves all Objective-C tests from `cs` to the correct `oc` folder. While at it, removes duplicated tests that were already in `oc` folder and fixes a couple of expected test files.


micheleCTDEAdmin (approved):
Looks good, thanks

PR #3638, opened 2022-04-29
Vala: properly support nullable types

dbartolini:
Marks Vala nullable types as PTR_TYPE (like C# already does).

Previously, nullable types in Vala were not detected at all:
```
# -=====-
# language                      = VALA
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp         Flag   Nl  Text
#   1>               TYPE|               NONE|     PARENT_NOT_SET[  1/  1/  4/  0][0/0/0][     200c 0000][0-0] int
#   1>           QUESTION|               NONE|     PARENT_NOT_SET[  4/  4/  5/  0][0/0/0][   2 2000 0000][0-0]    ?
#   1>               WORD|               NONE|     PARENT_NOT_SET[  6/  6/  9/  1][0/0/0][     2008 0000][0-0]      foo
#   1>             ASSIGN|               NONE|     PARENT_NOT_SET[ 10/ 10/ 11/  1][0/0/0][   2 0000 0000][0-0]          =
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 12/ 12/ 16/  1][0/0/0][        8 0000][0-0]            null
#   1>          SEMICOLON|               NONE|     PARENT_NOT_SET[ 16/ 16/ 17/  0][0/0/0][   2 0000 0000][0-0]                ;
#   1>            NEWLINE|               NONE|     PARENT_NOT_SET[ 17/ 17/  1/  0][0/0/0][             0][1-0]
# -=====-
```

This PR fixes it:
```
# -=====-
# language                      = VALA
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp         Flag   Nl  Text
#   1>               TYPE|               NONE|     PARENT_NOT_SET[  1/  1/  4/  0][0/0/0][     208c 0000][0-0] int
#   1>           PTR_TYPE|               NONE|     PARENT_NOT_SET[  4/  4/  5/  0][0/0/0][   2 2080 0000][0-0]    ?
#   1>               WORD|               NONE|     PARENT_NOT_SET[  6/  6/  9/  1][0/0/0][     2300 0000][0-0]      foo
#   1>             ASSIGN|               NONE|     PARENT_NOT_SET[ 10/ 10/ 11/  1][0/0/0][   2 0000 0000][0-0]          =
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 12/ 12/ 16/  1][0/0/0][        8 0000][0-0]            null
#   1>          SEMICOLON|               NONE|     PARENT_NOT_SET[ 16/ 16/ 17/  0][0/0/0][   2 0000 0000][0-0]                ;
#   1>            NEWLINE|               NONE|     PARENT_NOT_SET[ 17/ 17/  1/  0][0/0/0][             0][1-0]
# -=====-
```

As a consequence, the option ``sp_type_question`` should now be removed, since ``sp_before_ptr_star`` can be used instead. 
I didn't bother to actually delete it, I will, should this PR be accepted.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -481,9 +481,6 @@ sp_type_ellipsis                = ignore   # ignore/add/remove/force/not_defined
 # Add or remove space between a '*' and '...'.
 sp_ptr_type_ellipsis            = ignore   # ignore/add/remove/force/not_defined
 
-# (D) Add or remove space between a type and '?'.
-sp_type_question                = ignore   # ignore/add/remove/force/not_defined
-
 # Add or remove space between ')' and '...'.
 sp_paren_ellipsis               = ignore   # ignore/add/remove/force/not_defined
 
--- a/documentation/htdocs/default.cfg
+++ b/documentation/htdocs/default.cfg
@@ -481,9 +481,6 @@ sp_type_ellipsis                = ignore   # ignore/add/remove/force/not_defined
 # Add or remove space between a '*' and '...'.
 sp_ptr_type_ellipsis            = ignore   # ignore/add/remove/force/not_defined
 
-# (D) Add or remove space between a type and '?'.
-sp_type_question                = ignore   # ignore/add/remove/force/not_defined
-
 # Add or remove space between ')' and '...'.
 sp_paren_ellipsis               = ignore   # ignore/add/remove/force/not_defined
 
--- a/etc/defaults.cfg
+++ b/etc/defaults.cfg
@@ -481,9 +481,6 @@ sp_type_ellipsis                = ignore   # ignore/add/remove/force/not_defined
 # Add or remove space between a '*' and '...'.
 sp_ptr_type_ellipsis            = ignore   # ignore/add/remove/force/not_defined
 
-# (D) Add or remove space between a type and '?'.
-sp_type_question                = ignore   # ignore/add/remove/force/not_defined
-
 # Add or remove space between ')' and '...'.
 sp_paren_ellipsis               = ignore   # ignore/add/remove/force/not_defined
 
--- a/etc/uigui_uncrustify.ini
+++ b/etc/uigui_uncrustify.ini
@@ -1072,15 +1072,6 @@ Choices=sp_ptr_type_ellipsis=ignore|sp_ptr_type_ellipsis=add|sp_ptr_type_ellipsi
 ChoicesReadable="Ignore Sp Ptr Type Ellipsis|Add Sp Ptr Type Ellipsis|Remove Sp Ptr Type Ellipsis|Force Sp Ptr Type Ellipsis"
 ValueDefault=ignore
 
-[Sp Type Question]
-Category=1
-Description="<html>(D) Add or remove space between a type and '?'.</html>"
-Enabled=false
-EditorType=multiple
-Choices=sp_type_question=ignore|sp_type_question=add|sp_type_question=remove|sp_type_question=force|sp_type_question=not_defined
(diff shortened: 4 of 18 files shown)


micheleCTDEAdmin (changes requested):
Ciao Daniele,
this PR also includes the changes of PR #3637 (which has been merged btw).
Can you please rework the code and include only the changes that relates to this specific point?
Thanks.

dbartolini:
Ciao Michele,
I rebased to latest master and split changes better.

Should we forget about ``sp_type_question`` and just nuke it from the codebase?
A more sensible approach would be to show a warning message when the option is used, while copying its value to ``sp_before_ptr_star`` to keep backward compatibility.

micheleCTDE:
```sp_type_question``` is meant to be used to control spacing in a ternary operator like ```cond ? true-value : false-value```.
```sp_before_ptr_star``` is one of the options available to control spacing around pointers.

I think it is correct to identify a nullable type as a pointer star, like in C#. But ```sp_type_question``` is used for a different purpose and should come into play in this case, therefore it should not be removed at all, IMO. Instead options like ```sp_before_ptr_star``` and ```sp_after_ptr_star``` could be used to control the spacing around it.

With reference to this example:
```
int? x = true ? null : 2;
```
the first ? should be a (nullable) pointer star token, the second should be a question token. Does the removal of the code in src/space.cpp affect this?

**Comment by dbartolini** [comment]
> `sp_type_question` is meant to be used to control spacing in a ternary operator like `cond ? true-value : false-value`.
> `sp_before_ptr_star` is one of the options available to control spacing around pointers.

``sp_type_question`` docs say otherwise:
```
// (vala) Add or remove space between a type and '?'.
extern Option<iarf_e>
sp_type_question;
```

In fact, options to control spacing in ternary operator (``sp_cond_question*``) existed and did work OK in Vala before ``sp_type_question`` was added.
At that time, though, an option to control spacing around nullable types was missing *and* ``sp_cond_question*`` were mistakenly affecting nullable types as well as ternary operators (see #2090).

Before this PR, using ``sp_cond_question_after = remove`` on your example would produce wrong output:
```
int?x = true ?null : 2;
```
Similarly wrong output when ``sp_type_question = remove`` is used instead:
```
int? x = true? null : 2;
```
Here, both the ternary op and the nullable type are affected, although the latter may go unnoticed because the space between ``int`` and ``?`` was missing already.


> the first ? should be a (nullable) pointer star token, the second should be a question token. Does the removal of the code in src/space.cpp affect this?

After this PR, you can use ``sp_cond_question*`` and ``sp_before_ptr_star`` etc. on your example and always get the right answer, because nullable types are correctly detected now:
```
// sp_cond_question_before = remove
// sp_before_ptr_star = add
// Would give correct answer:
int ? x = true? null : 2;
```

micheleCTDE:
Ciao Daniele,
my apologies, I got too deep into the role of ```CT_QUESTION``` when I checked the code yesterday and I missed that ```sp_type_question``` is only used in one place in the code. After a new check, I agree with you that ```sp_type_question``` is no longer needed since a nullable type is now identified as a "pointer star" and there are other options to set that.

> A more sensible approach would be to show a warning message when the option is used, while copying its value to sp_before_ptr_star to keep backward compatibility.

We could comment the option out from default.cfg and leave a comment in place that explains that the option is no longer avaiable and ```sp_before_ptr_star``` could be used for it.
@guy-maurel, as project maintainer, what do you think?

dbartolini:
Added ``process_option_line_compat_0_74()`` and removed any reference to ``sp_type_question``.

micheleCTDE (approved):
Looks good, thanks!

micheleCTDEAdmin (approved):

PR #3639, opened 2022-04-30
Introduce new option nl_collapse_empty_body_ignore_functions.

dundargoc:
Collapsing empty braces is nice except for in funcitons where it looks
out of place. This is a way to remedy this.


dundargoc:
I'm not really sure why the cli_options is failing. Looking into it.

dundargoc:
lol, help me out here please. I don't know what I'm supposed to fix. It says 

```
Problem (1) with /home/dgoc/programs/uncrustify/build/tests/cli/results/universalindent.cfg
```

but I'm not sure what the problem is.

micheleCTDE:
> lol, help me out here please. I don't know what I'm supposed to fix. It says
> 
> ```
> Problem (1) with /home/dgoc/programs/uncrustify/build/tests/cli/results/universalindent.cfg
> ```
> 
> but I'm not sure what the problem is.

Hi Dundar,
there is a test target ```update-cli-options``` (see [here](https://github.com/uncrustify/uncrustify/blob/master/tests/CMakeLists.txt#L82) which can be used to update the cli option output. This should possibly fix your issue.
If you build with ninja, you can go to your build folder and simply run ```ninja update-cli-options```, it will take care of it.

src/newlines.cpp
```diff
@@ -2234,6 +2234,13 @@ static bool collapse_empty_body(Chunk *br_open)
    {
       return(false);
    }
+   log_rule_B("nl_collapse_empty_body_ignore_functions");
+
+   if (  options::nl_collapse_empty_body_ignore_functions()
+      && br_open->GetPrevNcNnlNet()->Is(CT_FPAREN_CLOSE))
```
micheleCTDE, line 2240: I am confused here. Why are we looking back from the open brace? Shouldn't we look forward? What am I missing?

dundargoc: Maybe it's possibly to do it in a simpler/more straightforward way, if so please tell me :)

Basically I want to check whether the brace belongs to a function definition since this rule is only meant to be triggered for function braces. This checks if the previous chunk is a closing parenthesis that belongs to a function (I think so at least).

micheleCTDE: You can use the chunk parent to check if it s a function, like this:
```
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp         Flag   Nl  Text
#   1>               TYPE|           FUNC_DEF|     PARENT_NOT_SET[  1/  1/  5/  0][0/0/0][        c 0000][0-0] void
#   1>           FUNC_DEF|               NONE|     PARENT_NOT_SET[  6/  6/  9/  1][0/0/0][             0][0-0]      bar
#   1>        FPAREN_OPEN|           FUNC_DEF|     PARENT_NOT_SET[  9/  9/ 10/  0][0/0/0][   2 0000 0000][0-0]         (
#   1>               TYPE|               NONE|     PARENT_NOT_SET[ 10/ 10/ 14/  0][0/1/0][        8 0008][0-0]          void
#   1>       FPAREN_CLOSE|           FUNC_DEF|     PARENT_NOT_SET[ 14/ 14/ 15/  0][0/0/0][   2 0000 0008][0-0]              )
#   1>            NEWLINE|               NONE|     PARENT_NOT_SET[ 15/ 15/  1/  0][0/0/0][             0][1-0]
#   2>         BRACE_OPEN|           FUNC_DEF|     PARENT_NOT_SET[  1/  1/  2/  0][0/0/0][   2 0000 0000][0-0] {
#   2>            NEWLINE|               NONE|     PARENT_NOT_SET[  2/  2/  8/  0][1/1/0][             0][1-0]
#   3>                 IF|               NONE|     PARENT_NOT_SET[  8/  8/ 10/  0][1/1/0][        c 0000][0-0]        if
```
You can see the brace open parent is ```FUNC_DEF```.
You can use the ```-ds dump``` option on a test file to get a list of files showing you all the chunks during the process, this will help you finding more info.

Btw, could you provide an example of what the option should and should not do? I see one expected positive test but not a negative one, so I don't get the difference yet

dundargoc: Oh yeah, checking the parent is way easier. Thanks :)

Good point about the tests. I've added a few more examples. I guess the "ignore_functions" isn't technically correct as it expands/uncollapses(?) the braces, but I couldn't come up with a more correct name that wasn't obnoxiously long.

dundargoc:
[Done :)](https://github.com/dundargoc/uncrustify/actions)

micheleCTDE (changes requested):
When a new option is added, there are a number of files that need to be edited (cfg, doc). Please compare with some of the existing options and you will see what I mean.

micheleCTDE:
Unrelated to the comment above, I think it may be a better idea to structure this a bit differently.
There is already an option ```nl_collapse_empty_body``` for collapsing empty bodies.
It may be good to have an option ```nl_collapse_empty_body_function``` to override that option specifically for functions.
Later if needed we may have options like ```nl_collapse_empty_body_class```, ```nl_collapse_empty_body_namespace``` and so on, to be able to specify different behavior for different types of empty blocks if needed.
@guy-maurel what do you think?

micheleCTDE:
> When a new option is added, there are a number of files that need to be edited (cfg, doc). Please compare with some of the existing options and you will see what I mean.

Thanks, this point is ok now.

micheleCTDE:
> It may be good to have an option nl_collapse_empty_body_function to override that option specifically for functions.

As suggested, we could make this options work the other way round, as an override of ```nl_collapse_empty_body```.
For example:
```
nl_collapse_empty_body = true
nl_collapse_empty_body_functions = false
```
would collapse empty bodies but excluding functions. Later ```nl_collapse_empty_body_classes``` could be used to override ```nl_collapse_empty_body``` for empty classes and so on.

Finally: it would be good to add also some negative tests, that is a test where an empty non-function body is collapsed, while a function body is not. It's to test/prove the option works fine in both cases.

dundargoc:
Hmm, yeah, that's more intuitive for sure. Will need to check the code on how to best do that.

tests/c.test
```diff
@@ -103,6 +103,7 @@
 00154  common/sp_after_cast.cfg                   c/casts.c
 00155  c/ben_078.cfg                              c/cast_brace.c
 
+00200  c/nl_collapse_empty_body_ignore_functions.cfg  c/empty_functions.c
```
gmaurel, line 106: 00200 is not very lucky, as we try to keep the numbers in order.
156 untill 159 are free, or 200 at another place in the file

gmaurel (changes requested):
see comment

tests/c.test
dundargoc: Done

micheleCTDE:
@dundargoc 

> Finally: it would be good to add also some negative tests, that is a test where an empty non-function body is collapsed, while a function body is not. It's to test/prove the option works fine in both cases.

Nice, the new additional negative tests look very good. Thanks!

To me the only point left is to rework the option name and logic as described in earlier posts. Instead of using an "exclude" logic (```nl_collapse_empty_body_ignore_functions```), it would be better to use an "override" logic (```nl_collapse_empty_body_functions``` overriding ```nl_collapse_empty_body``` for functions only). Once this is done, I think we can merge it.

dundargoc:
Hmm, should `nl_collapse_empty_body_functions` override `nl_collapse_empty_body` for both `true` and `false`? That is, should

```
nl_collapse_empty_body = true
nl_collapse_empty_body_functions = false 
```

collapse all empty braces except for functions and

```
nl_collapse_empty_body = false
nl_collapse_empty_body_functions = true
```

collapse all empty function braces but skip all other empty braces?

The second sounds a bit dubious but just want to make sure.

micheleCTDE:
> Hmm, should `nl_collapse_empty_body_functions` override `nl_collapse_empty_body` for both `true` and `false`? That is, should

Hi @dundargoc, good question! I had not thought about the second possibility initially.
We can go two ways, so I would like to hear your opinion too.

Option 1
```nl_collapse_empty_body_functions``` overrides ```nl_collapse_empty_body``` only as per your first example, that is:
```
nl_collapse_empty_body = true
nl_collapse_empty_body_functions = false 
```
So ```nl_collapse_empty_body_functions``` works based on an overriding exclusion logic.

Option 2
```nl_collapse_empty_body_functions``` overrides ```nl_collapse_empty_body``` all the times.
In this case, ```nl_collapse_empty_body_functions``` is a standalone option for the function empty body, while ```nl_collapse_empty_body``` applies to all other types of empty bodies. In such case we need to update the description of ```nl_collapse_empty_body``` to make clear what is included and what not.

After some thinking, I think option 2 may be better, but I am interested to hear your thoughts as well. 
Question also open to @guy-maurel

**Comment by dundargoc** [comment]
Hmm, second option sounds fine by me.

In the future we might want to rework this into a single option that takes a list of strings maybe? Something like

```
collapse_these_blocks = [function, for, if]
```

Just an idea.

**Comment by micheleCTDE** [comment]
> Hmm, second option sounds fine by me.
> 
> In the future we might want to rework this into a single option that takes a list of strings maybe? Something like
> 
> ```
> collapse_empty_blocks = [function, for, if]
> ```
> 
> Just an idea.

This is indeed an interesting idea. Worth discussing this with @guy-maurel when he is back. Will need to find a way to specify both inclusions and exclusions, for example something like
 ```
collapse_empty_blocks = [all, !function]
 ```
to specify all blocks excluding the function blocks.

**Comment by micheleCTDE** [comment]
> Hmm, second option sounds fine by me.

Btw, is this PR ready for a new review or are you still working on option 2?

**Comment by dundargoc** [comment]
I thought I'd wait for Guy to respond before working more on this. I can move forward with this now as well?

**Comment by micheleCTDE** [comment]
> I thought I'd wait for Guy to respond before working more on this. I can move forward with this now as well?

@guy-maurel may not be around for a short while. I think you can move forward with option 2 :-)

**Comment by gmaurel** [comment]
the idea with collapse_empty_blocks = [all, !
is good.
I prefer more options, one for each use.

File: src/newlines.cpp [hunk 3]
```diff
@@ -2234,6 +2234,13 @@ static bool collapse_empty_body(Chunk *br_open)
    {
       return(false);
    }
+   log_rule_B("nl_collapse_empty_body_functions");
+
+   if (  options::nl_collapse_empty_body_functions()
```
micheleCTDE, line 2239: Shouldn't this be 
```
!options::nl_collapse_empty_body_functions()
```
?

tests/config/c/nl_collapse_empty_body_functions.cfg
```diff
@@ -0,0 +1,2 @@
+nl_collapse_empty_body = true
+nl_collapse_empty_body_functions = true
```
micheleCTDE, line 2: Based on the file tests/expected/c/00156-empty-braces.c, shouldn't this be 'false'? This would also match with the missing ! in the previous comment.

Second note: we should have two tests, one with:
```
nl_collapse_empty_body = true
nl_collapse_empty_body_functions = true
```
and one with 
```
nl_collapse_empty_body = true
nl_collapse_empty_body_functions = false
```
so we test the new option in both positive and negative case. One case should also collapse the function bodies, the other shouldn't.

micheleCTDE (changes requested):
Please see comment and amend.

src/newlines.cpp
dundargoc: Yes :). I wasn't quite done.

micheleCTDE:
Looks like some test is failing

dundargoc:
I've been a bit busy, so I've drafted this for now. I'll enable this for review once I'm done :)

dundargoc:
@micheleCTDE ok I think I'm done. Please have a look :)

src/options.h
```diff
@@ -1884,11 +1884,18 @@ donot_indent_func_def_close_paren;
 ///////////////////////////////////////////////////////////////////////////////
 //BEGIN Newline adding and removing options
 
-// Whether to collapse empty blocks between '{' and '}'.
-// If true, overrides nl_inside_empty_func
+// Whether to collapse empty blocks between '{' and '}' except for functions.
```
micheleCTDE, line 1887: The description here is fine, but it needs to be updated in various other files so that it is aligned everywhere (see .cfg and doc files)

src/options.h
```diff
@@ -1884,11 +1884,18 @@ donot_indent_func_def_close_paren;
...
+// Use nl_collapse_empty_body_functions to specify how empty function braces
+// should be formatted.
 extern Option<bool>
 nl_collapse_empty_body;
 
+// Whether to skip collapsing empty blocks between '{' and '}' for
```
micheleCTDE, line 1893: The description and the functionaliy of the option are opposite. 
"skip collapsing" implies that if the option is 'true', then empty function bodies will not be collapsed. But if the option is 'true' it does in fact collapse the body.
The description should be reworded in a way similar to the one of `nl_collapse_empty_body`, that is "Whether to collapse empty function blocks ...."

src/options.h
```diff
@@ -2731,7 +2738,7 @@ nl_max_blank_in_func;
 // The number of newlines inside an empty function body.
 // This option overrides eat_blanks_after_open_brace and
 // eat_blanks_before_close_brace, but is ignored when
-// nl_collapse_empty_body=true
+// nl_collapse_empty_body_functions=true
```
micheleCTDE, line 2741: Nice catch. The same update needs to be done in all option description files.

tests/config/c/nl_collapse_empty_body_functions.cfg
micheleCTDE: Nice, thanks to your new code I realized we were missing 2 test cases more.
We should have all 4 combinations of ```nl_collapse_empty_body``` and ```nl_collapse_empty_body_functions```, that is:
- true, true
- true, false
- false, true
- false, false
This will cover all the possible scenarios of the two options.

tests/expected/c/00911-pascal_ptr.c
```diff
@@ -9,4 +9,5 @@ char* main()
 }
 
 void foo(int* a, int* b, int* c)
-{}
+{
```
micheleCTDE, line 12: We can probably amend the config used for this test to avoid this unneeded change

micheleCTDE (changes requested):
We are getting ther :-)
Need to fix up a bit of text in the various option descriptions and add two missing tests. See comments for the details.

src/options.h
dundargoc: Ah, forgot. Thanks for the reminder.

Is there a way to automate this perhaps? The number of files that need updating are quite a lot and it's easy to forget to update all of them.

dundargoc: Better yet, is there a way to not need multiple identical files that are scattered around the codebase?

dundargoc:
Unrelated: there's a test in `cli_options` that fails on linux but passes on windows. It's easy to fix it for linux, but then the CI for windows fails. Any idea what's up with that?

src/options.h
micheleCTDE: That is a good point, it is something to look at on my TODO list as well.
Basically config.txt and defaults.cfg in the htdocs folder are the same as defaults.cfg in etc folder. 
universalindent.cfg in the test folder should be (IMO) the same as uigui_uncrustify.ini in etc, but currently it isn't. It is autogenerated, so there may be something to fix in the generation process..
The question is whether we need all those files or if some could be removed. Then if some duplicates are needed, we could look at automate the process, so we only update etc/defaults.cfg and the other files are autogenerated from it.
I need to check with @guy-maurel about the past history of these files to know more.

micheleCTDE:
> Unrelated: there's a test in `cli_options` that fails on linux but passes on windows. It's easy to fix it for linux, but then the CI for windows fails. Any idea what's up with that?
> 
> Edit: hmm, on closer inspection the cli_options are tested on the linux CI as well. Maybe I just misconfigured something locally?

Run 'ninja update-cli-options', this will update the cli output files accordingly. In linux the test usually pass, not sure on Windows since I don't test on that env. Maybe line ending differences affect the test if you did some manual editing?

etc/uigui_uncrustify.ini
```diff
@@ -2831,7 +2831,7 @@ ValueDefault=0
 
 [Indent Macro Brace]
 Category=2
-Description="<html>How to indent within a macro followed by a brace on the same line<br/>This allows reducing the indent in macros that have (for example)<br/>`do { ... } while (0)` blocks bracketing them.<br/><br/>true:  add an indent for the brace on the same line as the macro<br/>false: do not add an indent for the brace on the same line as the macro<br/><br/>Default: true</html>"
+Description="<html>How to indent within a macro followed by a brace on the same line<br/>This allows reducing the indent in macros that have (for example)<br/>`do { ... } while ` blocks bracketing them.<br/><br/>true:  add an indent for the brace on the same line as the macro<br/>false: do not add an indent for the brace on the same line as the macro<br/><br/>Default: true</html>"
```
micheleCTDE, line 2834: ```do { ... } while (0)``` should remain (the while condition is part of the do-while construct). I know that the output for the autogenereated cli test (see may other comment on PR conversation) is missing this ```(0)```: I think we should fix the autogeneration to make sure that ```(0)``` is there as well. This can be done separately from this PR.

micheleCTDE (changes requested):
Excellent work, one minor thing to rectify and then we are ready for merging!

dundargoc:
> > Unrelated: there's a test in `cli_options` that fails on linux but passes on windows. It's easy to fix it for linux, but then the CI for windows fails. Any idea what's up with that?
> > Edit: hmm, on closer inspection the cli_options are tested on the linux CI as well. Maybe I just misconfigured something locally?
> 
> Run 'ninja update-cli-options', this will update the cli output files accordingly. In linux the test usually pass, not sure on Windows since I don't test on that env. Maybe line ending differences affect the test if you did some manual editing?

In the `build` directory? Running `ninja update-cli-options` gived the following warning:

```
test_cli_options.py: error: argument --config: expected one argument
```

etc/uigui_uncrustify.ini
dundargoc: Done

micheleCTDE (approved):

micheleCTDEAdmin (approved):

gmaurel (approved):
fine work. thanks.

micheleCTDEAdmin:
And finally merged. Thanks @dundargoc for the good work and fixing up all the corrections that were required :-)

dundargoc:
Thanks for all the help :)

PR #3642, opened 2022-05-02
Run CI test for pull requests.

dundargoc:
The benefit is that the all the test results will be available on the PR
page. The uncrustify developers will also no longer need to ask to the
pull requester to activate their workflow.


micheleCTDEAdmin (approved):
This is nice! Thanks.

PR #3643, opened 2022-05-02
Add ability to control spacing between type and ? in C#

dbartolini:
Adds the ability to control spacing between ``type`` and ``?`` in C# via ``sp_before_ptr_star`` and ``sp_after_ptr_star``.

Previously, spaces before ``?``, if any, were always removed.

One thing to note is that the default value for ``sp_{before,after}_ptr_star`` is set to ``ignore``, which may produce different results than before if spaces between ``type`` and ``?`` are found (see bug_600.cs).


micheleCTDEAdmin (approved):
Looks good, thanks Daniele.

PR #3671, opened 2022-05-28
a tool to generate the p files from the sources

guy-maurel:
only for developers

### DIFF
--- a/tests/generate_p_files.sh
+++ b/tests/generate_p_files.sh
@@ -1,0 +1,21 @@
+#! /bin/sh
+#
+# Author:  guy maurel
+# May 2022
+#
+# generate p-files from the sources
+#
+# the test is only usefull for the developpers
+#
+LIST=`ls -1 ../src/*.cpp`
+LIST_H=`ls -1 ../src/*.h`
+P_FILES="../build/P-files"
+#
+mkdir -p ${P_FILES}
+#
+for each in ${LIST} ${LIST_H}
+do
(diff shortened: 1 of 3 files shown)


tests/generate_p_files.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/sh
+#
+# Author:  guy maurel
+# May 2022
+#
+# generate p-files from the sources
+#
+# the test is only usefull for the developpers
```
micheleCTDEAdmin, line 8: usefull --> useful
developpers --> developers

tests/generate_p_files.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+LIST=`ls -1 ../src/*.cpp`
+LIST_H=`ls -1 ../src/*.h`
+P_FILES="../build/P-files"
```
micheleCTDEAdmin, line 12: The paths are hard-coded. It would be more beneficial to get the paths from cmake during the configuration stage, so that the script can run in other folders too. For example I don't build in the ```build``` folder.

micheleCTDEAdmin (commented):
1) typos to fix
2) see comment on hard coded paths. I will leave it to you on whether you want to merge this as is or adjust it as I mentioned.

gmaurel (approved):

tests/generate_p_files.sh
guy-maurel: yes, you are right

PR #3677, opened 2022-05-29
Use auto when declaring iterators.

dundargoc:
More readable in my opinion. Obviously subjective, don't hesitate to close if you disagree.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -3689,7 +3689,7 @@ static void handle_oc_property_decl(Chunk *os)
 
          Chunk *curr_chunk = open_paren;
 
-         for (multimap<int, std::vector<ChunkGroup> >::reverse_iterator it = sorted_chunk_map.rbegin(); it != sorted_chunk_map.rend(); ++it)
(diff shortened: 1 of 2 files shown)


micheleCTDE:
Personally I don't like the use of 'auto' and if it was for me, I would actually remove it rather than adding it. It makes the code less readable because you have no clue what type the variable is, unless you go and dig further.
So I defer to @guy-maurel whether he wants to merge this PR or not.

dundargoc:
Eh, was mostly a suggestion. I'll close this :+1:

micheleCTDE:
> Eh, was mostly a suggestion. I'll close this +1

I guess we can wait for @guy-maurel 's opinion. This is an open source project and Guy is the maintainer of it.
I expressed my opinion and reasons for it, but it does not mean it is "the law" :-)

gmaurel:
I don't like "auto"
It is difficult to read it after some weeks.
Please, suppress "auto"

dundargoc:
Yes, Fair enough :)

Should we close this then?

dundargoc:
Your thoughts about the `auto` keyword makes sense. Thanks for the explanation :+1:

PR #3679, opened 2022-05-29
new option to insert the number of the line

guy-maurel:
at the beginning of each line
for developer only

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1,4 +1,4 @@
-# Uncrustify_d-0.75.0-60-a7a9d652-dirty
+# Uncrustify_d-0.76.0-development
 
 #
 # General options
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1,4 +1,4 @@
-# Uncrustify-0.75.0
+# Uncrustify_d-0.75.0-60-a7a9d652-dirty
 
 #
 # General options
@@ -3441,6 +3441,9 @@ debug_timeout                   = 0        # number
 # Set the number of characters to be printed if the text is too long,
 # 0: do not truncate.
 debug_truncate                  = 0        # unsigned number
+
+# insert the number of the line at the beginning of each line
+set_numbering_for_html_output   = false    # true/false
 
 # Meaning of the settings:
 #   Ignore - do not do any changes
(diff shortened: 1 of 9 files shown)


micheleCTDEAdmin (changes requested):
New options needs to be added to default.cfg and related files too.

documentation/htdocs/config.txt
```diff
@@ -1,4 +1,4 @@
-# Uncrustify-0.75.0
+# Uncrustify_d-0.75.0-60-a7a9d652-dirty
```
micheleCTDEAdmin, line 1: Probably this change was unintentional...

documentation/htdocs/default.cfg
```diff
@@ -1,4 +1,4 @@
-# Uncrustify-0.75.0
+# Uncrustify_d-0.75.0-60-a7a9d652-dirty
```
micheleCTDEAdmin, line 1: Probably this change was unintentional...

etc/defaults.cfg
```diff
@@ -1,4 +1,4 @@
-# Uncrustify-0.75.0
+# Uncrustify_d-0.75.0-60-a7a9d652-dirty
```
micheleCTDEAdmin, line 1: Probably this change was unintentional...

etc/uigui_uncrustify.ini
```diff
@@ -15,20 +15,20 @@ parameterOrder=ipo
 showHelpParameter=-h
 stringparaminquotes=false
 useCfgFileParameter="-c "
-version=Uncrustify-0.75.0
+version=Uncrustify_d-0.75.0-60-a7a9d652-dirty
```
micheleCTDEAdmin, line 18: Probably this change was unintentional...

micheleCTDEAdmin (changes requested):
Minor touch up required related to version number in cfg file.

micheleCTDEAdmin:
> reflect different things, exactly what we have (a new version)

Yes, I figured you used the /release_tool.py script.
I agree the current version is newer that 0.75, but I don't think it is good to add a hash reference in the config files, especially a temporary one like -dirty. Maybe we can use something like '# Uncrustify_d-0.76.0-development' to indicate that it is a version that will end up in 0.76_f but it is still in development? What do you think?

guy-maurel:
Yes, I used release_tool.py.
But, using uncrustify  --show-config produce the same line.
Your proposal is fine.
Could you integrate it in the script?
I'll make the change per editor, for just now.

micheleCTDEAdmin (approved):
Looks good.

micheleCTDEAdmin:
> Could you integrate it in the script?

I have added this in my TODO list. Will do at some point in coming weeks.

PR #3681, opened 2022-05-29
Replace chunk_is_str with IsString().

dundargoc:


micheleCTDEAdmin (changes requested):
The idea is good, but we need to refine it.
1) Chunk::Is() and Chunk::IsNot() methods relates to the concept of chunk type. I think it would be better to call the new method Chunk::IsString() to make sure the method Chunk::IsNot() does not become ambiguous (not a type? not a string? not both? if you get what I mean)
2) there is already a method Chunk::IsStringAndLevel() which does pretty much the same thing. We can leverage a call to it in Chunk::IsString(), we just need to pass the string length and ANY_LEVEL as level (which will basically ignore the level check).
3) you should not use "str" as an argument name in a Chunk method, it would shadow the internal member with the same name. Although not a problem, it gives linting warnings and is not a good practice either. You can see that in "Chunk::IsStringAndLevel()" I had to rename that into cStr instead.

Other than that, it is a good idea so go ahead with it.

micheleCTDEAdmin (changes requested):
Please amend the mentioned points

PR #3684, opened 2022-05-30
Fix todo by making function argument const.

dundargoc:


src/chunk.cpp
```diff
@@ -646,7 +646,7 @@ Chunk *chunk_first_on_line(Chunk *pc)
 }
 
 
-bool chunk_is_last_on_line(Chunk *pc)  //TODO: pc should be const here
+bool chunk_is_last_on_line(const Chunk *pc)  //TODO: pc should be const here
```
micheleCTDEAdmin, line 649: there is no need for the TODO comment to still be there

micheleCTDEAdmin (changes requested):
as per comment

micheleCTDEAdmin (approved):
Looks good!

PR #3686, opened 2022-05-30
Add missing option check.

dundargoc:
Uncrustify may group includes containing filename even if
mod_sort_incl_import_prioritize_filename is set to false. This will
prevent that.


dundargoc:
There is something weird going on that I cannot  pinpoint. I'm having trouble recreating it in the tests for some reason. I'll need some more time. Closing this in the meantime.

PR #3688, opened 2022-05-31
Null loop

guy-maurel:
There might be some more!

### DIFF
--- a/src/EnumStructUnionParser.cpp
+++ b/src/EnumStructUnionParser.cpp
@@ -1460,7 +1460,8 @@ void EnumStructUnionParser::mark_base_classes(Chunk *pc)
 
    pcf_flags_t flags = PCF_VAR_1ST_DEF;
 
-   while (pc != nullptr)
+   while (  pc != nullptr
+         && pc->IsNotNullChunk())
    {
       chunk_flags_set(pc, PCF_IN_CLASS_BASE);
       /**
--- a/tests/generate_p_files.sh
+++ b/tests/generate_p_files.sh
@@ -5,7 +5,7 @@
 #
 # generate p-files from the sources
 #
-# the test is only usefull for the developpers
+# the test is only useful for the developers
 #
 LIST=`ls -1 ../src/*.cpp`
 LIST_H=`ls -1 ../src/*.h`


src/align_add.cpp
```diff
@@ -18,7 +18,8 @@ void align_add(ChunkStack &cs, Chunk *pc, size_t &max_col)
    size_t min_col;
    Chunk  *prev = Chunk::NullChunkPtr;
 
-   if (pc != nullptr)
+   if (  pc != nullptr
+      && pc->IsNotNullChunk())
```
micheleCTDEAdmin, line 22: ``` if (pc != nullptr)``` was added to prevent SEGV during the transition process of the code improvement. At the end this will simply be ```prev = pc->GetPrev();``` without any ```if```, because Chunk pointers won't be null.
Checking for ```pc->IsNotNullChunk()``` is not wrong, but it is not necessary either. If ```pc``` is a null chunk, ```pc->GetPrev()``` will return a Chunk::NullChunkPtr, with no effect on the current value of ```prev```

src/indent.cpp
```diff
@@ -493,7 +493,8 @@ static Chunk *oc_msg_block_indent(Chunk *pc, bool from_brace,
    LOG_FUNC_ENTRY();
    Chunk *tmp = Chunk::NullChunkPtr;
 
-   if (pc != nullptr)
+   if (  pc != nullptr
+      && pc->IsNotNullChunk())
```
micheleCTDEAdmin, line 497: ```if (pc != nullptr)``` was added to prevent SEGV during the transition process of the code improvement. At the end this will simply be ```tmp = pc->GetPrevNc();``` without any ```if```, because Chunk pointers won't be null.
Checking for ```pc->IsNotNullChunk()``` is not wrong, but it is not necessary either. If ```pc``` is a null chunk, ```pc->GetPrevNc()``` will return a Chunk::NullChunkPtr, with no effect on the current value of ```tmp```

src/indent.cpp
```diff
@@ -4707,7 +4708,8 @@ void indent_preproc()
       }
       Chunk *next = pc->GetNextNcNnl();
 
-      if (next->IsNullChunk())
+      if (  pc != nullptr
+         && pc->IsNullChunk())
```
micheleCTDEAdmin, line 4712: Wrong replacement (```next``` to ```pc```) and not necessary because ```pc->GetNextNcNnl();``` will never return a ```nullptr``` by design.
This part of code should remain unchanged.

src/indent.cpp
```diff
@@ -4731,7 +4733,8 @@ void indent_preproc()
       log_rule_B("pp_space");
 
       if (  (options::pp_space() != IARF_IGNORE)
-         && next != nullptr)
+         && next != nullptr
```
micheleCTDEAdmin, line 4736: Here ```next``` is guaranteed to be non ```nullptr```, so we only need to check for ``` && next->IsNotNullChunk())```

src/lang_pawn.cpp
```diff
@@ -275,7 +276,8 @@ static Chunk *pawn_process_variable(Chunk *start)
    LOG_FUNC_ENTRY();
    Chunk *pc = Chunk::NullChunkPtr;
 
-   if (start != nullptr)
+   if (  start != nullptr
+      && start->IsNotNullChunk())
```
micheleCTDEAdmin, line 280: ```if (start != nullptr)``` was added to prevent SEGV during the transition process of the code improvement. At the end this will simply be ```pc = start;``` without any ```if```, because Chunk pointers won't be null.
Checking for ```start->IsNotNullChunk()``` is not wrong, but it is not necessary either. If ```start``` is a null chunk, there is no effect on the current value of ```pc```

src/newlines.cpp
```diff
@@ -2734,7 +2735,8 @@ static void newline_before_return(Chunk *start)
 
    Chunk *pc = Chunk::NullChunkPtr;
 
-   if (start != nullptr)
+   if (  start != nullptr
+      && start->IsNotNullChunk())
```
micheleCTDEAdmin, line 2739: ``` if (start != nullptr)``` was added to prevent SEGV during the transition process of the code improvement. At the end this will simply be ```pc = start->GetPrev();``` without any ```if```, because Chunk pointers won't be null.
Checking for ```start->IsNotNullChunk()``` is not wrong, but it is not necessary either. If ```start ``` is a null chunk, ```start ->GetPrev()``` will return a Chunk::NullChunkPtr, with no effect on the current value of ```pc```

src/newlines.cpp
```diff
@@ -2879,7 +2881,8 @@ void newline_iarf(Chunk *pc, iarf_e av)
    log_func_stack(LNFD, "CallStack:");
    Chunk *after = Chunk::NullChunkPtr;
 
-   if (pc != nullptr)
+   if (  pc != nullptr
+      && pc->IsNotNullChunk())
```
micheleCTDEAdmin, line 2885: As comment above, no need to check for IsNotNullChunk() here.

src/newlines.cpp
```diff
@@ -3279,8 +3282,9 @@ static void newline_func_def_or_call(Chunk *start)
                a = options::nl_func_type_name_class();
             }
 
-            if (  a != IARF_IGNORE
-               && prev != nullptr)
+            if (  (a != IARF_IGNORE)
+               && prev != nullptr
```
micheleCTDEAdmin, line 3286: If we get here ```prev``` is not a nullptr for sure, so we only need to check for ```&& prev->IsNotNullChunk())```

src/newlines.cpp
```diff
@@ -6594,7 +6598,8 @@ static void newlines_enum_entries(Chunk *open_brace, iarf_e av)
 
    Chunk *pc = Chunk::NullChunkPtr;
 
-   if (open_brace != nullptr)
+   if (  open_brace != nullptr
+      && open_brace->IsNotNullChunk())
```
micheleCTDEAdmin, line 6602: As some comments above, this is only a temporary check to prevent possible SEGV. No need to check for ```IsNotNullChunk()```

src/newlines.cpp
```diff
@@ -6623,7 +6628,8 @@ static void newlines_double_space_struct_enum_union(Chunk *open_brace)
...
```
micheleCTDEAdmin, line 6632: As some comments above, this is only a temporary check to prevent possible SEGV. No need to check for ```IsNotNullChunk()```

micheleCTDEAdmin (changes requested):
Please see comments and amend as necessary.

micheleCTDEAdmin:
> There might be some more!

Quite possible that during the code improvement transition phase some checks for nullptr need adjustment. As we are progressing, more and more methods guarantee to return a valid Chunk pointer (either a real chunk or a NullChunkPtr), so nullptr is used progressively less and less.

guy-maurel:
Thanks for the many comments

src/indent.cpp
micheleCTDE: We can leave the original as is, as explained in previous comment.

src/indent.cpp
micheleCTDE: Need to use ```next```, not ```pc```

src/lang_pawn.cpp
micheleCTDE: We can leave the original as is, as explained in previous comment.

micheleCTDE (changes requested):
There are still a few points that need to be adjusted (from the first review...)

micheleCTDE (approved):

micheleCTDEAdmin (approved):

micheleCTDEAdmin:
@guy-maurel  looks good. Will leave it to you to decide whether to merge as is or to squash the various commit and then merge

gmaurel (approved):

PR #3693, opened 2022-06-01
sp before unnamed byref

### LINKED ISSUE #3691: sp_before_unnamed_byref failed in std::function
current version: Uncrustify-0.75.1
source commit: e8564e37a28150eafabd6a74d03c13cb4efb2417

test options:
```
sp_type_func                = add       # 
sp_after_type               = add       #
sp_fparen_brace             = force

sp_before_byref             = add
sp_after_byref              = remove    # "int * a" vs "int *a"
sp_before_unnamed_byref     = remove    # "int&" vs "int &" (return type)
sp_before_byref_func        = remove
sp_after_byref_func         = add
sp_byref_paren              = add
```

test code:
```
void test(std::function<void(const int&)> func);
```

my test environment is win 10，and test command is:
{path}\uncrustify.exe -c {path}\test.cfg {path}\test.h

output code:
```
void test(std::function<void(const int &)> func);
}
```

my expectation is same as original code, 
no space between "const int" and "&"
thanks for any advice.

guy-maurel:
fixes #3691

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -1376,6 +1376,8 @@ static iarf_e do_space(Chunk *first, Chunk *second, int &min_sp)
       if (  next->IsNotNullChunk()
          && (  chunk_is_token(next, CT_COMMA)
             || chunk_is_token(next, CT_FPAREN_CLOSE)
+            || (  chunk_is_token(next, CT_PAREN_CLOSE)            // Issue #3691
+               && get_chunk_parent_type(next) == CT_CPP_CAST)
             || chunk_is_token(next, CT_SEMICOLON)))
       {
          if (options::sp_before_unnamed_byref() != IARF_IGNORE)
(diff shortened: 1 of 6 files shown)


micheleCTDE:
I think a better solution would be to look at the precedence between ```sp_before_unnamed_byref``` and ```sp_before_byref```. Currently
```
sp_before_byref                 = add    # ignore/add/remove/force/not_defined
sp_before_unnamed_byref         = remove    # ignore/add/remove/force/not_defined
```
adds the space before the unnamed byref, which should not happen IMO.

guy-maurel:
`adds the space before the unnamed byref,...`
No! it doesn't at me

micheleCTDEAdmin:
Closing this because it has been replaced by PR #3885.

PR #3695, opened 2022-06-03
Complete replacement of chunk_is_not_token() with Chunk::IsNot()

micheleCTDE:
As per title.
There is a subtle difference between the two functions when a nullptr/null chuck is used. This cause several tests to fail at first. Upon investingating the fail, this was caused by wrong "expected" test files in the source code, which were missing a last line newline character when the last character was a closing brace. That is, the source code test files had been purposely edited to pass the test with the old chunk_is_not_token(), even if that subtle difference was causing an unexpected missing newline. The new code correctly adds the last newline.

As a further test, formatting my usual 3200 file test repo gave no regressions and no single difference at all.


micheleCTDE:
@guy-maurel 
before merging this, would like to hear your opinion on the modified tests which were missing the newline at the end.
I will be away for 3 days, so feel free to merge if you don't have any objections.

guy-maurel (approved):
fine

gmaurel (approved):
thanks for the changes

PR #3696, opened 2022-06-03
Add mod_infinite_loop option

alexhenrie:
This is definitely the largest new feature that I have tried to add to Uncrustify. To keep it as simple as possible, `mod_infinite_loop` is pretty strict about what input it accepts and what output it creates. I welcome any feedback you may have!

### DIFF
--- a/src/chunk.cpp
+++ b/src/chunk.cpp
@@ -471,6 +471,11 @@ void Chunk::Delete(Chunk * &pc)
 void Chunk::MoveAfter(Chunk *ref)
 {
    LOG_FUNC_ENTRY();
+
+   if (ref == this)
+   {
+      return;
+   }
    g_cl.Pop(this);
    g_cl.AddAfter(this, ref);
 
--- a/src/rewrite_infinite_loops.cpp
+++ b/src/rewrite_infinite_loops.cpp
@@ -1,0 +1,322 @@
+/**
+ * @file rewrite_infinite_loops.cpp
+ *
+ * @author  Alex Henrie
+ * @license GPL v2+
+ */
+
+#include "rewrite_infinite_loops.h"
+
+#include "chunk.h"
+#include "newlines.h"
+#include "uncrustify.h"
+
+
+using namespace uncrustify;
+
+
+static bool for_needs_rewrite(Chunk *pc, E_Token desired_type)
+{
+   // The 'for' statement needs to be rewritten if `for(;;)` is not the
+   // preferred syntax for infinite loops and this 'for' is an infinite loop
+   // with no extra tokens (such as inline comments).
+
+   if (desired_type == CT_FOR)
+   {
+      return(false);
+   }
+   pc = pc->GetNext();
+
+   if (!pc->Is(CT_SPAREN_OPEN))
+   {
+      return(false);
+   }
+   pc = pc->GetNext();
+
+   if (!pc->Is(CT_SEMICOLON))
+   {
+      return(false);
+   }
+   pc = pc->GetNext();
+
+   if (!pc->Is(CT_SEMICOLON))
+   {
+      return(false);
+   }
+   pc = pc->GetNext();
+
+   if (!pc->Is(CT_SPAREN_CLOSE))
+   {
+      return(false);
+   }
+   return(true);
+}
+
+
+static bool while_needs_rewrite(Chunk *keyword, E_Token desired_type, const char *desired_condition)
+{
+   // The 'while' statement needs to be rewritten if it has only the tokens that
+   // are strictly necessary (keyword, condition, two parentheses, and semicolon
+   // if do-while) and either the keyword or the condition needs to be changed.
+
+   Chunk *oparen    = keyword->GetNext();
+   Chunk *condition = oparen->GetNext();
+   Chunk *cparen    = condition->GetNext();
+
+   if (!oparen->Is(CT_SPAREN_OPEN))
+   {
+      return(false);
+   }
+
+   if (  strcmp(condition->Text(), "true") != 0
+      && strcmp(condition->Text(), "1") != 0)
+   {
+      return(false);
+   }
+
+   if (!cparen->Is(CT_SPAREN_CLOSE))
+   {
+      return(false);
+   }
+
+   if (keyword->Is(CT_WHILE_OF_DO))
+   {
+      Chunk *semicolon = cparen->GetNext();
+
+      if (!semicolon->Is(CT_SEMICOLON))
+      {
+         return(false);
+      }
+   }
+
+   if (!keyword->Is(desired_type))
+   {
+      return(true);
+   }
+
+   if (  strcmp(condition->Text(), "true") == 0
+      && strcmp(desired_condition, "true") != 0)
+   {
+      return(true);
+   }
+
+   if (  strcmp(condition->Text(), "1") == 0
+      && strcmp(desired_condition, "1") != 0)
+   {
+      return(true);
+   }
+   return(false);
+} // while_needs_rewrite
+
+
+void rewrite_loop_keyword(Chunk *keyword, E_Token new_type)
+{
+   set_chunk_type(keyword, new_type);
+
+   switch (new_type)
+   {
+   case CT_DO:
+      keyword->orig_col_end += strlen("do") - keyword->Len();
+      keyword->str           = "do";
+      break;
+
+   case CT_WHILE:
+   case CT_WHILE_OF_DO:
+      keyword->orig_col_end += strlen("while") - keyword->Len();
+      keyword->str           = "while";
+      break;
+
+   case CT_FOR:
+      keyword->orig_col_end += strlen("for") - keyword->Len();
+      keyword->str           = "for";
+      break;
+
+   default:
+      break;
+   }
+}
+
+
+static void move_one_token(Chunk * &source, Chunk * &destination, E_Token parent_type)
+{
+   Chunk *next_source = source->GetNext();
+
+   // Place the source token immediately after the destination token, without
+   // any whitespace.
+
+   source->MoveAfter(destination);
+   source->column       = destination->column + destination->Len();
+   source->orig_col     = destination->orig_col + destination->Len();
+   source->orig_col_end = source->orig_col_end + source->Len();
+   source->orig_prev_sp = 0;
+   set_chunk_parent(source, parent_type);
+
+   destination = source;
+   source      = next_source;
+}
+
+
+static void rewrite_loop_condition(Chunk * &source, Chunk * &destination,
+                                   E_Token desired_type, const char *desired_condition)
+{
+   // Move the opening parenthesis
+   move_one_token(source, destination, desired_type);
+
+   // Move the condition
+   if (desired_type == CT_FOR)
+   {
+      set_chunk_type(source, CT_SEMICOLON);
+      set_chunk_parent(source, CT_FOR);
+      source->str = ";";
+      move_one_token(source, destination, desired_type);
+      destination = (destination)->CopyAndAddAfter(destination);
+   }
+   else
+   {
+      set_chunk_type(source, CT_WORD);
+      source->str = desired_condition;
+      move_one_token(source, destination, desired_type);
+   }
+
+   // If converting a 'for' to a 'while', delete the second semicolon
+   if (source->Is(CT_SEMICOLON))
+   {
+      Chunk *next_source = source->GetNext();
+      Chunk::Delete(source);
+      source = next_source;
+   }
+   // Move the closing parenthesis
+   move_one_token(source, destination, desired_type);
+}
+
+
+void rewrite_loop_in_place(Chunk *keyword, E_Token desired_type, const char *desired_condition)
+{
+   Chunk *top    = keyword->GetNext();
+   Chunk *bottom = keyword;
+
+   rewrite_loop_keyword(keyword, desired_type);
+   rewrite_loop_condition(top, bottom, desired_type, desired_condition);
+}
+
+
+static Chunk *find_end_brace(Chunk *pc)
+{
+   while (!pc->IsBraceOpen())
+   {
+      pc = pc->GetNextNcNnl();
+   }
+   return(pc->SkipToMatch());
+}
+
+
+void rewrite_infinite_loops()
+{
+   LOG_FUNC_ENTRY();
+
+   E_Token    desired_type;
+   const char *desired_condition;
+
+   switch (options::mod_infinite_loop())
+   {
+   case 1: // for(;;)
+      desired_type      = CT_FOR;
+      desired_condition = nullptr;
+      break;
+
+   case 2: // while(true)
+      desired_type      = CT_WHILE;
+      desired_condition = "true";
+      break;
+
+   case 3: // do...while(true)
+      desired_type      = CT_WHILE_OF_DO;
+      desired_condition = "true";
+      break;
+
+   case 4: // while(1)
+      desired_type      = CT_WHILE;
+      desired_condition = "1";
+      break;
+
+   case 5: // do...while(1)
+      desired_type      = CT_WHILE_OF_DO;
+      desired_condition = "1";
+      break;
+
+   default:
+      return;
+   }
+
+   for (Chunk *pc = Chunk::GetHead(); pc->IsNotNullChunk(); pc = pc->GetNextNcNnl())
+   {
+      if (pc->Is(CT_DO))
+      {
+         Chunk *while_keyword = find_end_brace(pc)->GetNextNcNnl();
+
+         if (  !while_keyword->Is(CT_WHILE_OF_DO)
+            || !while_needs_rewrite(while_keyword, desired_type, desired_condition))
+         {
+            continue;
+         }
+
+         if (desired_type == CT_WHILE_OF_DO)
+         {
+            rewrite_loop_in_place(while_keyword, desired_type, desired_condition);
+         }
+         else
+         {
+            Chunk *top    = pc;
+            Chunk *bottom = while_keyword->GetNext();
+
+            // Change the 'do' at the top of the loop to a 'for' or a 'while'
+            rewrite_loop_keyword(top, desired_type);
+
+            // Delete the 'while' at the bottom of the loop
+            Chunk::Delete(while_keyword);
+
+            // Move the rest of the tokens from the bottom to the top
+            rewrite_loop_condition(bottom, top, desired_type, desired_condition);
+
+            // Delete the final semicolon
+            Chunk::Delete(bottom);
+         }
+      }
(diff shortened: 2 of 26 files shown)


micheleCTDE:
Hi @alexhenrie,
I will have a look at this during the weekend and feedback. Sorry for the wait.

src/log_levels.h
```diff
@@ -132,6 +132,7 @@ enum log_sev_t
    LOUTPUT   = 101, //! output
    LUNC      = 102, //! rules used in uncrustify.cpp
    LQT       = 103, //! load/save options for Qt
+   LMINFLOOP = 104, //! mod_infinite_loop
```
micheleCTDE, line 135: I think we don't need a specific log option for this, unless you can explain a good use case for it. Using 'log_rule_B' should suffice, but happy to hear @guy-maurel 's opinion as well.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,328 @@
...
+   {
+      pc = pc->GetNextNcNnl();
+   }
+   return(pc->SkipToMatch());
+}
+
+
+static bool for_needs_rewrite(Chunk *pc, E_Token desired_type)
+{
+   assert(chunk_is_token(pc, CT_FOR));
```
micheleCTDE, line 29: better to use 'if' rather than assert and return if the condition is not met, IMO. Likewise in the rest of the file.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,328 @@
...
+      desired_condition = "1";
+      break;
+
+   case 5: // do...while(1)
+      desired_type      = CT_WHILE_OF_DO;
+      desired_condition = "1";
+      break;
+
+   default:
+      assert(false);
```
micheleCTDE, line 246: Don't 'assert', simply print a log message about the wrong value and return without doing any changes.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,328 @@
...
+   if (  strcmp(condition->Text(), "1") == 0
+      && strcmp(desired_condition, "1") != 0)
+   {
+      return(true);
+   }
+   return(false);
+} // while_needs_rewrite
+
+
+static void move_one_token(Chunk **source, Chunk **destination, E_Token parent_type)
```
micheleCTDE, line 120: There is a Swap method in Chunk. Could we use that to swap to chunk perhaps? Just suggesting to have a look, maybe what you need here is more that what the function does.

micheleCTDE (changes requested):
Some things to clean up, but it seems promising.
Would be better if it was a bit less stringent, allowing separators to be accepted too, like for example
```for ( ; ; )``` or ```while ( true )```

**Review by micheleCTDE — CHANGES_REQUESTED** [review]

File: tests/expected/c/10081-infinite-loops.c [hunk 5]
```diff
@@ -0,0 +1,39 @@
+int main(void)
+{
+	int x;
+
+	for (;;)
```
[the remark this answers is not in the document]
alexhenrie: I'm expecting the user to set `sp_before_sparen`, `sp_inside_sparen`, and `sp_inside_for` in addition to `mod_infinite_loop`. Otherwise, there is no guarantee that loops will have similar spacing. Do you want me to set `sp_before_sparen` in the tests so that all the loops in the expected output look the same?

src/chunk.cpp
```diff
@@ -471,6 +471,7 @@ void Chunk::Delete(Chunk * &pc)
 void Chunk::MoveAfter(Chunk *ref)
 {
    LOG_FUNC_ENTRY();
+   assert(ref != this);
```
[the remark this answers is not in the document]
alexhenrie: Done.

src/log_levels.h
alexhenrie: I have removed the LOG_FMT calls as requested.

alexhenrie:
@micheleCTDE Thanks for the feedback. I have tried to address all of your comments, but I may need a little more direction to understand what you'd like to see.

micheleCTDE:
> @micheleCTDE Thanks for the feedback. I have tried to address all of your comments, but I may need a little more direction to understand what you'd like to see.

I will take a look this weekend and comment back. Apologies for the wait, it's another busy week...

tests/expected/c/10081-infinite-loops.c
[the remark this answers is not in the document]
micheleCTDE: Ah, I see now. The ones with the additional space are those cases which were not replaced and so are the same as the original code. In such case I think it should be fine, because other options should take care of normalizing spacing later on. I have not tested, but if you have verified that doing so is enough, then we can just ignore my original comment.

tests/expected/c/10082-infinite-loops.c
```diff
@@ -0,0 +1,39 @@
+int main(void)
+{
+	int x;
+
+	while(true)
+	{
+	}
+
+	while (true)
```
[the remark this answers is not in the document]
micheleCTDE: as per comment above

tests/expected/c/10083-infinite-loops.c
```diff
@@ -0,0 +1,43 @@
...
+
+	do
+		x++; while(true);
+
+	do
+		x++; while(true);
+
+	do
+		x++;
+	while (true);
```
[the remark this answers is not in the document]
micheleCTDE: If settings other options results in a consistent indentation/spacing, that would be fine. But if that results in do-while loops with different formatting in the end, then we should consider doing some extra work here.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,310 @@
...
+      if (!chunk_is_token(semicolon, CT_SEMICOLON))
+      {
+         return(false);
+      }
+   }
+
+   if (!chunk_is_token(keyword, desired_type))
+   {
+      return(true);
+   }
```
micheleCTDE, line 98: This seems wrong. If the result is true, it says it is ok to rewrite, without checking that the condition is "true" or "1".

src/rewrite_infinite_loops.cpp
micheleCTDE: ok, makes sense. It was just a thought I wanted to share.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,310 @@
...
+   return(false);
+} // while_needs_rewrite
+
+
+static void move_one_token(Chunk **source, Chunk **destination, E_Token parent_type)
+{
+   Chunk *next_source = (*source)->GetNext();
+
+   // Place the token immediately after the previous token, without any
+   // whitespace.
```
micheleCTDE, line 120: Comment should refer to source and destination, which not necessarily are adjacent.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+   (*source)->MoveAfter(*destination);
+   (*source)->column       = (*destination)->column + (*destination)->Len();
+   (*source)->orig_col     = (*destination)->orig_col + (*destination)->Len();
+   (*source)->orig_col_end = (*source)->orig_col_end + (*source)->Len();
+   (*source)->orig_prev_sp = 0;
```
micheleCTDE, line 126: I have not done a real test, but once 'source' is moved after 'destination', do the destination's column/orig_col/orig_col_end/orig_prev_sp also need some updating? why only source's values are modified? if I am missing something obvious, please let me know.

micheleCTDE:
@alexhenrie 
Sorry for the long delay. Please see my comments above, some changes are likely to be required.
As we progress with refining this PR, I will do some actual testing (not done yet), so I may have more comments in future.

micheleCTDE:
Forgot to mention that it would be good to add some negative tests, that is tests where loops don't get replaced because they don't meet the required/allowed syntax. This will help to catch eventual mistakes.

src/rewrite_infinite_loops.cpp
alexhenrie: You're right; when changing from `while` to `for`, the function should still make sure that the original condition was `true` or `1`. I have now fixed that and added a test for it.

src/rewrite_infinite_loops.cpp
alexhenrie: I have rephrased the comment to use the words "source" and "destination".

alexhenrie:
> Forgot to mention that it would be good to add some negative tests, that is tests where loops don't get replaced because they don't meet the required/allowed syntax. This will help to catch eventual mistakes.

Thanks for the feedback. I have added several more test cases, including some related to the problem you found where the loop condition was not being checked. Do you see anything else amiss?

src/rewrite_infinite_loops.cpp
micheleCTDE: Still worth to check and make sure the values in the dumps are correct, even if uncrustify sorts things out fine in the end.
You can use the -ds option to see intermiediate steps and see if there is any inconsistent column values.

tests/expected/c/10083-infinite-loops.c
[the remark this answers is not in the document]
micheleCTDE: ok, seems reasonable

tests/expected/c/10085-infinite-loops.c
```diff
@@ -0,0 +1,83 @@
...
+	{
+	}while(1);
+
+	do
+	{
+	}
+	while (1);
+
+	do
+		x++; while(1);
```
micheleCTDE, line 28: I think we could add a new like before ```while(1)``` when replacing into do-while loop. That should be an easy change and probably what everyone would expect, regardless of what settings is used for other options.

tests/expected/c/10083-infinite-loops.c
```diff
@@ -0,0 +1,83 @@
...
+	{
+	}while(true);
+
+	do
+	{
+	}
+	while(true);
+
+	do
+		x++; while(true);
```
micheleCTDE, line 28: I think we could add a new like before while(true) when replacing into do-while loop. That should be an easy change and probably what everyone would expect, regardless of what settings is used for other options.

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,316 @@
...
+   // preferred syntax for infinite loops and this 'for' is an infinite loop
+   // with no extra tokens (such as inline comments).
+
+   if (desired_type == CT_FOR)
+   {
+      return(false);
+   }
+   pc = pc->GetNext();
+
+   if (!chunk_is_token(pc, CT_SPAREN_OPEN))
```
micheleCTDE, line 39: you can replace ```chunk_is_token()``` with ```->Is()``` since we have started to do this in the code.

src/rewrite_infinite_loops.cpp
micheleCTDE: great

micheleCTDE (changes requested):
Seems ready in terms of functionality. Few minor code improvements (see comments).
After that I will do a test of a ~3200 file repo and if no regression are highlighted, we can pass this to @guy-maurel for his comments and final approval.

src/rewrite_infinite_loops.cpp
alexhenrie: I tried the `-ds` option as you suggested and it created six parse dumps, numbered `-c_002.log` through `-c_007.log`. The tokens are shown as overlapping in `-c_002.log` through `-c_004.log`. However, in `-c_005.log` through `-c_007.log`, the tokens are moved so that they do not overlap. I am not sure how to solve this, and I'm still not sure that it's a real problem. Perhaps we can leave it as it is for now, and revisit it later if it turns out to be important?

`infinite-loops.c`:

```
	do // comment
		x++;
	while (1); // comment
```

`-c_002.log`:

```
# 113>                FOR|               NONE|     PARENT_NOT_SET[  9/  9/ 12/  0][1/1/0][        c 0000][0-1]         for
# 115>        SPAREN_OPEN|                FOR|     PARENT_NOT_SET[ 12/ 12/ 14/  0][1/1/0][   2 0000 0000][0-0]            (
# 115>          SEMICOLON|                FOR|     PARENT_NOT_SET[ 13/ 13/ 16/  0][1/2/0][        8 0020][0-0]             ;
# 115>          SEMICOLON|                FOR|     PARENT_NOT_SET[ 13/ 13/ 16/  0][1/2/0][        8 0020][0-0]             ;
# 115>       SPAREN_CLOSE|                FOR|     PARENT_NOT_SET[ 14/ 14/ 17/  0][1/1/0][   2 0000 0000][0-0]              )
# 113>        VBRACE_OPEN|                 DO|     PARENT_NOT_SET[ 12/  9/  0/  0][1/1/0][             0][0-0]
# 113>        COMMENT_CPP|        COMMENT_END|     PARENT_NOT_SET[ 12/ 12/ 22/  1][2/2/0][             0][0-0]            // comment
# 113>            NEWLINE|               NONE|     PARENT_NOT_SET[ 22/ 22/ 17/  0][2/2/0][             0][1-0]
# 114>               WORD|               NONE|     PARENT_NOT_SET[ 17/ 17/ 18/  0][2/2/0][        c 0000][0-1]                 x
# 114>       INCDEC_AFTER|               NONE|     PARENT_NOT_SET[ 18/ 18/ 20/  0][2/2/0][   2 0000 0000][0-0]                  ++
# 114>          SEMICOLON|               NONE|     PARENT_NOT_SET[ 20/ 20/ 21/  0][2/2/0][   2 0000 0000][0-0]                    ;
# 114>       VBRACE_CLOSE|                 DO|     PARENT_NOT_SET[  0/ 20/  0/  0][1/1/0][             0][0-0]
# 114>            NEWLINE|               NONE|     PARENT_NOT_SET[ 21/ 21/  9/  0][1/1/0][             0][1-0]
# 115>        COMMENT_CPP|        COMMENT_END|     PARENT_NOT_SET[ 20/ 20/ 30/  1][1/1/0][             0][0-0]                    // comment
```

`-c_005.log`:

```
# 113>                FOR|               NONE|     PARENT_NOT_SET[  9/  9/ 12/  0][1/1/0][        c 0000][0-1]         for
# 115>        SPAREN_OPEN|                FOR|     PARENT_NOT_SET[ 12/ 12/ 14/  0][1/1/0][   2 0000 0000][0-0]            (
# 115>          SEMICOLON|                FOR|     PARENT_NOT_SET[ 13/ 13/ 16/  0][1/2/0][        8 0020][0-0]             ;
# 115>          SEMICOLON|                FOR|     PARENT_NOT_SET[ 14/ 13/ 16/  0][1/2/0][        8 0020][0-0]              ;
# 115>       SPAREN_CLOSE|                FOR|     PARENT_NOT_SET[ 15/ 14/ 17/  0][1/1/0][   2 0000 0000][0-0]               )
# 113>        VBRACE_OPEN|                 DO|     PARENT_NOT_SET[ 16/  9/  0/  0][1/1/0][             0][0-0]
# 113>        COMMENT_CPP|        COMMENT_END|     PARENT_NOT_SET[ 17/ 12/ 22/  1][2/2/0][             0][0-0]                 // comment
# 113>            NEWLINE|               NONE|     PARENT_NOT_SET[ 27/ 22/ 17/  0][2/2/0][             0][1-0]
# 114>               WORD|               NONE|     PARENT_NOT_SET[ 17/ 17/ 18/  0][2/2/0][        c 0000][0-1]                 x
# 114>       INCDEC_AFTER|               NONE|     PARENT_NOT_SET[ 18/ 18/ 20/  0][2/2/0][   2 0000 0000][0-0]                  ++
# 114>          SEMICOLON|               NONE|     PARENT_NOT_SET[ 20/ 20/ 21/  0][2/2/0][   2 0000 0000][0-0]                    ;
# 114>       VBRACE_CLOSE|                 DO|     PARENT_NOT_SET[ 22/ 20/  0/  0][1/1/0][             0][0-0]
# 114>            NEWLINE|               NONE|     PARENT_NOT_SET[ 22/ 21/  9/  0][1/1/0][             0][1-0]
# 115>        COMMENT_CPP|      COMMENT_WHOLE|     PARENT_NOT_SET[  9/ 20/ 30/  1][1/1/0][             0][0-0]         // comment
```

src/rewrite_infinite_loops.cpp
alexhenrie: Done

src/rewrite_infinite_loops.cpp
```diff
@@ -0,0 +1,316 @@
...
+      keyword->str           = "for";
+      break;
+
+   default:
+      break;
+   }
+}
+
+
+static void rewrite_loop_condition(Chunk **source, Chunk **destination,
```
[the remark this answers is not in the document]
alexhenrie: Done

alexhenrie:
Hi Michele, I have addressed your comments above and added a bunch more test cases just to be sure. The output does look ugly in a few places, but as we discussed, the user can easily get prettier output by setting more options. I hope that the code is good enough to move forward now.

src/rewrite_infinite_loops.cpp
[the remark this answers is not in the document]
micheleCTDE: Great!

micheleCTDE (approved):

micheleCTDE:
> Hi Michele, I have addressed your comments above and added a bunch more test cases just to be sure. The output does look ugly in a few places, but as we discussed, the user can easily get prettier output by setting more options. I hope that the code is good enough to move forward now.

Hi @alexhenrie,
thanks for the patience and the adjustments. Seems good to go.
The output does look a bit ugly at times, I agree. But setting other options should take care of it as discussed.

I could be more picky here and say for example that a
```
for (;;) {
...
}
```
should look like 
```
do {
...
}
while (true);
```
for example. This will result in less ugly replacements.

@guy-maurel as maintainer, what is your opinion? Any other comment about the PR overall?

alexhenrie:
> I could be more picky here and say for example that a
> 
> ```
> for (;;) {
> ...
> }
> ```
> 
> should look like
> 
> ```
> do {
> ...
> }
> while (true);
> ```
> 
> for example. This will result in less ugly replacements. Similar observations for other replacements, the test files have some good example of things that could be better formatted.

That would be a good future improvement. I considered trying to preserve the original formatting as much as possible (like you are suggesting), but I decided that it was too complicated to do as part of the initial pull request. What we have now is a good start and although it's not absolutely perfect, it works well and would be a very helpful feature for me.

micheleCTDEAdmin:
> @guy-maurel as maintainer, what is your opinion? Any other comment about the PR overall?

Guy has confirmed to me that he is happy with this PR, so I will merge it.

micheleCTDEAdmin:
@alexhenrie thanks for the good work!

alexhenrie:
Thank you!

PR #3697, opened 2022-06-06
Make --check output more readable and relevant.

dundargoc:
The current behavior of the --check flag is very suboptimal. It shows
the number of bytes that has changed or the current number of bytes,
which I don't believe is relevant for the user.

Additionally, it also shows which files have passed alongside those that
have failed. This is fine for a small project but becomes practically
impossible to read for projects with hundreds of files. Not using the -q
flag already tells us which files are being parsed, it's obvious files
that aren't shown to have failed have passed.

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -1441,9 +1441,8 @@ static bool bout_content_matches(const file_mem &fm, bool report_status)
    {
       if (report_status)
       {
-         fprintf(stderr, "FAIL: %s (File size changed from %u to %u)\n",
-                 cpd.filename.c_str(), static_cast<int>(fm.raw.size()),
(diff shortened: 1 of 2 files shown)


micheleCTDE:
Not sure of what are the reasons why "--check" works as it is. Maybe some users may use those additional info about bytes for scripts or other things.
@guy-maurel any ideas? 

IMO, being something already there and not causing any harm in printing those extra bytes, I don't see a good enough reason to remove that part. 
@guy-maurel: what is your opinion in this?

gmaurel (changes requested):
It is not false to get those info, even if you don't need them.
Please, do not suppress.

dundargoc:
Yes, I understand. Thank you for your time.

PR #3708, opened 2022-06-20
Start working on replacing chunk_is_token() with Chunk::Is()

metmo:
While looking into the code base I saw that there's ongoing work in replacing the pattern to use the `Chunk` class and it's methods, so I thought I'll do some instances of replacing to familiarize myself with the code before I attempt any features or bug fixes.

This is the first part in replacing occurrences of `chunk_is_token()` with the newer and preferred `Chunk::Is()`. There's roughly 1700 occurrences left after this and more parts will follow if deemed OK and if no one else is already preparing the same changes. 

This completely changes all the occurrences in the following files:
```console
src/combine.cpp
src/newlines.cpp
src/parens.cpp
src/semicolons.cpp
```

Since there's no logic or structural changes I squashed everything into one commit, but I can separate the changes to a commit per file if that's preferred.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -347,7 +347,7 @@ static bool can_increase_nl(Chunk *nl)
          return(rv);
       }
 
-      if (  chunk_is_token(next, CT_PREPROC)
+      if (  next->Is(CT_PREPROC)
          && (  get_chunk_parent_type(next) == CT_PP_ELSE
             || get_chunk_parent_type(next) == CT_PP_ENDIF)
          && (  next->level > 0
@@ -361,7 +361,7 @@ static bool can_increase_nl(Chunk *nl)
       }
    }
 
-   if (chunk_is_token(next, CT_BRACE_CLOSE))
+   if (next->Is(CT_BRACE_CLOSE))
    {
       if (  options::nl_inside_namespace() > 0
          && get_chunk_parent_type(next) == CT_NAMESPACE)
@@ -373,7 +373,7 @@ static bool can_increase_nl(Chunk *nl)
       }
 
       if (  options::nl_inside_empty_func() > 0
-         && chunk_is_token(prev, CT_BRACE_OPEN)
+         && prev->Is(CT_BRACE_OPEN)
          && (  get_chunk_parent_type(next) == CT_FUNC_DEF
             || get_chunk_parent_type(next) == CT_FUNC_CLASS_DEF))
       {
@@ -392,7 +392,7 @@ static bool can_increase_nl(Chunk *nl)
       }
    }
 
-   if (chunk_is_token(prev, CT_BRACE_CLOSE))
+   if (prev->Is(CT_BRACE_CLOSE))
    {
       if (  options::nl_before_namespace()
          && get_chunk_parent_type(prev) == CT_NAMESPACE)
@@ -404,7 +404,7 @@ static bool can_increase_nl(Chunk *nl)
       }
    }
 
-   if (chunk_is_token(prev, CT_BRACE_OPEN))
+   if (prev->Is(CT_BRACE_OPEN))
    {
       if (  options::nl_inside_namespace() > 0
          && get_chunk_parent_type(prev) == CT_NAMESPACE)
@@ -416,7 +416,7 @@ static bool can_increase_nl(Chunk *nl)
       }
 
       if (  options::nl_inside_empty_func() > 0
-         && chunk_is_token(next, CT_BRACE_CLOSE)
+         && next->Is(CT_BRACE_CLOSE)
          && (  get_chunk_parent_type(prev) == CT_FUNC_DEF
             || get_chunk_parent_type(prev) == CT_FUNC_CLASS_DEF))
       {
@@ -512,7 +512,7 @@ void double_newline(Chunk *nl)
    }
    LOG_FMT(LNEWLINE, "%s(%d): add newline after ", __func__, __LINE__);
 
-   if (chunk_is_token(prev, CT_VBRACE_CLOSE))
+   if (prev->Is(CT_VBRACE_CLOSE))
    {
       LOG_FMT(LNEWLINE, "VBRACE_CLOSE ");
    }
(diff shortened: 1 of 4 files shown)


micheleCTDE:
Hi Marcus,
welcome. I am actually working on the software upgrade in bit and pieces each week, but help is surely welcomed.
Replacing of chunk_is_token() should be relatively straightforward in the majority of the cases, but you should review each replacement and judge whether a check for nullptr is required or (even better) if an existing check for nullptr could be removed or at most replace with a check for NullPtr.

I will take a look at your work in details possibily tomorrow or Wed. If all good, it will be merged. In the meantime you can probably do the check I mentioned above (and yes, it is time consuming :-) )

metmo:
> Hi Marcus, welcome. I am actually working on the software upgrade in bits and pieces each week, but help is surely welcomed. Replacing of chunk_is_token() should be relatively straightforward in the majority of the cases, but you should review each replacement and judge whether a check for nullptr is required or (even better) if an existing check for nullptr could be removed or at most replace with a check for NullPtr.
> 
> I will take a look at your work in details possibily tomorrow or Wed. If all good, it will be merged. In the meantime you can probably do the check I mentioned above (and yes, it is time consuming :-) )

Thank you for your reply! I'll have a second look and make sure there's no risk of any invalid dereferencing sneaking in.

metmo:
I've now updated the PR and split it into separate commits for each file to be easier to review. After going through the usages I could find a possible nullptr dereference in `do_symbol_check()`, that one is taken care of in 1da879c1dd17421a032482fd8eb43e986e4b781f

micheleCTDE:
Hi Marcus,
there are a number of adjustments required in this PR.
I am going thorugh the changes and commenting where changes are needed. I hope to finish by tonight or tomorrow.
Once you fix them up, could you please add a commit on top of the existing ones? It would help me to focus on the requested changes and avoid having to go through the whole PR again.

src/newlines.cpp
```diff
@@ -1945,8 +1945,8 @@ static void newlines_do_else(Chunk *start, iarf_e nl_opt)
 
 static bool is_var_def(Chunk *pc, Chunk *next)
 {
-   if (  chunk_is_token(pc, CT_DECLTYPE)
-      && chunk_is_token(next, CT_PAREN_OPEN))
+   if (  pc->Is(CT_DECLTYPE)
+      && next->Is(CT_PAREN_OPEN))
```
micheleCTDE, line 1949: need check against nulptr for both pc and next

src/newlines.cpp
```diff
@@ -5082,8 +5082,8 @@ void newline_after_label_colon()
 
 static bool is_class_one_liner(Chunk *pc)
 {
-   if (  (  chunk_is_token(pc, CT_FUNC_CLASS_DEF)
-         || chunk_is_token(pc, CT_FUNC_DEF))
+   if (  (  pc->Is(CT_FUNC_CLASS_DEF)
```
micheleCTDE, line 5085: needs check against nullptr

micheleCTDE (commented):
Please see comments and amend as necessary

src/newlines.cpp
metmo: I've found other cases where the arguments passed into functions are not checked.
Is it preferred to always have all pointers checked, even if the function is static and used in one or two cases, and checked by the caller?

src/newlines.cpp
metmo: Same question as above, I see one usage of this function where the caller checks it. Should I safeguard from any future potential wrong usage? I have no objection in doing it, just so I know for future similar cases :)

metmo:
> Please see comments and amend as necessary

Thank you for you review, seems I missed quite a few.
I'll keep the next PR smaller to avoid similar mistakes.
I have addressed your comments in [3ce1a59](https://github.com/uncrustify/uncrustify/pull/3708/commits/3ce1a59503c81f2e5769e143ba291bc2de46f754) except the two concerning function calls with an open question if we want to safeguard static functions where the (current) usage is known.

src/newlines.cpp
micheleCTDE: If all the callers of the function check the pointer before the call, then we don't need to add a check here. Reason being that one of the objective of the code improvememnt effort is to make sure all Chunk pointers will never be nullptr (they will be NullChunkPtr in case). So when the improvement is completed, there will be no checks against nullptr, therefore if we know all the calls to a function are safe, then we can skip adding a check that would later been removed anyway.

src/newlines.cpp
micheleCTDE: Same answer, if the callers already check it, we can skip adding a check here since it would be temporary anyway.

micheleCTDE:
> > Please see comments and amend as necessary
> 
> Thank you for you review, seems I missed quite a few. I'll keep the next PR smaller to avoid similar mistakes. I have addressed your comments in [3ce1a59](https://github.com/uncrustify/uncrustify/pull/3708/commits/3ce1a59503c81f2e5769e143ba291bc2de46f754) except the two concerning function calls with an open question if we want to safeguard static functions where the (current) usage is known.

Hi Marcus,
just wondering whether you noticed my comments from a few days ago or not.
If you did and verified the two remaining points, I take this PR is now completed.
In such case would you mind squashing the commits together? They relates to the same thing, so it's fine to group them together.

metmo:
> Hi Marcus, just wondering whether you noticed my comments from a few days ago or not. If you did and verified the two remaining points, I take this PR is now completed. In such case would you mind squashing the commits together? They relates to the same thing, so it's fine to group them together.

Hi, I checked the comments and as far as i can see, `is_var_def()` and `is_class_one_liner()` are used safely with 1 and 2 calls respectively, so I have squashed and rebased on master, thanks!

micheleCTDE:
Great, thanks. Will merge once the tests are passed :-)

micheleCTDEAdmin (approved):

micheleCTDE:
Merged to master. Thanks for the good work :smile:

PR #3713, opened 2022-06-26
Replace chunk is token() with Chunk::Is() pt2

metmo:
Here's some more replacing of `chunk_is_token()`. I've split it over three commits to be easier to review, I'll squash them before merging.

Most I would say are straight forward except for occurrences in the `skip_*` functions in `combine_skip.cpp` that do not null check the Chunk pointer passed in. After looking through the calls I would say that they're used safely, but on the other hand they are public and some are used in 6-10 places.


micheleCTDEAdmin (approved):
Looks good, thanks

micheleCTDEAdmin:
After you squash, I will merge.

metmo:
> After you squash, I will merge.

Great, thanks!

PR #3717, opened 2022-06-27
Fix interaction of mod_infinite_loop and mod_full_brace_*

### LINKED ISSUE #3716: mod_infinite_loop should be applied before mod_full_brace_*
Config file:

```
mod_infinite_loop = 1
mod_full_brace_for = add
nl_for_brace = add
```

Original code:

```
int main()
{
	while(1)
		x++;
}
```

Expected output:

```
int main()
{
	for(;;)
	{
		x++;
	}
}
```

Actual output:

```
int main()
{
	for(;;)
		x++;
}
```

If I run Uncrustify a second time, it does add the braces. However, they should have been added the first time.

alexhenrie:
Fixes #3716

### DIFF
--- a/src/rewrite_infinite_loops.cpp
+++ b/src/rewrite_infinite_loops.cpp
@@ -200,13 +200,13 @@ void rewrite_loop_in_place(Chunk *keyword, E_Token desired_type, const char *des
 }
 
 
-static Chunk *find_end_brace(Chunk *pc)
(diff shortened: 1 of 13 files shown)


micheleCTDEAdmin (approved):

micheleCTDE:
Looks good. Seems like some of those "open points" have finally show up ;-)

PR #3725, opened 2022-07-01
Make mod_*_int ignore qualifiers that come after the 'int' keyword

### LINKED ISSUE #3724: mod_*_int duplicates 'int' keyword if a qualifier comes after it
Input code:

```
short x;
int short x;
```

Config file:

```
mod_short_int = add
```

Expected results:

```
short int x;
int short x;
```

Actual results:

```
short int x;
int short int x;
```

alexhenrie:
Fixes #3724

My apologies for the chaotic development of this feature. I think this is the last patch it needs.

### DIFF
--- a/src/change_int_types.cpp
+++ b/src/change_int_types.cpp
@@ -14,20 +14,26 @@
 using namespace uncrustify;
 
 
-static void add_or_remove_int_keyword(Chunk *pc, iarf_e action)
+static bool is_int_qualifier(Chunk *pc)
+{
+   return(  strcmp(pc->Text(), "const") == 0
+         || strcmp(pc->Text(), "long") == 0
+         || strcmp(pc->Text(), "short") == 0
+         || strcmp(pc->Text(), "signed") == 0
+         || strcmp(pc->Text(), "static") == 0
+         || strcmp(pc->Text(), "unsigned") == 0
+         || strcmp(pc->Text(), "volatile") == 0);
+}
+
+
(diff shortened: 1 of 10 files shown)


micheleCTDE:
Hi @alexhenrie 
I am confused. If we set for example "remove unsigned int" and we have the following code
```
unsigned int unsigned_int_var;
int unsigned int_unsigned_var;
```
I would expect that "int" gets removed in both lines, regardless of whether it is before or after "unsigned".
Is the current behavior by choice (and why)?

alexhenrie:
@micheleCTDE All I really needed was the ability to _add_ the `int` keyword after `short`, `long`, `signed`, and `unsigned`. I included the ability to _remove_ the `int` keyword for completeness, but since it's complicated to handle the cases where `int` comes first or is separated from the size keyword by `const` or `volatile`, I didn't try to remove `int` in all of those cases. And the current behavior seems fairly logical to me, because setting `remove` removes the `int` keyword in the exact same contexts where setting `add` would add it.

If people are very interested in removing `int`, it might be worth extending these options' behavior to match what you expected. Or, we might add options like `mod_int_short` to complement `mod_short_int`, and use the combination `mod_int_short=remove` plus `mod_short_int=add` to transform `int short` to `short int`. I'll have to give it some more thought, but you are of course welcome to edit the code as you see fit ;-)

micheleCTDE:
Ok, thanks for the explanation. I guess we can merge this PR as is for the time being and work on the other options at a later stage or if people ask for it.

micheleCTDEAdmin (approved):

PR #3729, opened 2022-07-02
Ignore storage keywords when searching for last keyword in mod_*_int

### LINKED ISSUE #3728: mod_*_int should work despite 'const' before 'int'
Input code:

```
long const long_const_var;
long const int long_const_var;
```

`mod_long_int=add` should produce:

```
long int const long_const_var;
long const int long_const_var;
```

And `mod_long_int=remove` should produce:

```
const long const_long_var;
long const long_const_var;
```

Currently `mod_long_int` does not change either of them.

alexhenrie:
Fixes #3728

Thanks @micheleCTDE for your feedback last night. Your comment made me realize that it wouldn't be that hard to handle these cases better.

### DIFF
--- a/src/change_int_types.cpp
+++ b/src/change_int_types.cpp
@@ -14,24 +14,23 @@
 using namespace uncrustify;
 
 
-static bool is_int_qualifier(Chunk *pc)
-{
-   return(  strcmp(pc->Text(), "const") == 0
-         || strcmp(pc->Text(), "long") == 0
-         || strcmp(pc->Text(), "short") == 0
-         || strcmp(pc->Text(), "signed") == 0
-         || strcmp(pc->Text(), "static") == 0
-         || strcmp(pc->Text(), "unsigned") == 0
-         || strcmp(pc->Text(), "volatile") == 0);
-}
-
-
 static bool add_or_remove_int_keyword(Chunk *pc, iarf_e action)
 {
    Chunk *next = pc->GetNextNcNnl();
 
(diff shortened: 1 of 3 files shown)


micheleCTDE:
Hi @alexhenrie,
this PR improves some cases, but still does not really solve the problem I mentioned yesterday. You can see in your expected tests results cases where ```int``` is before ```short``` or ```long``` and it is not removed when requested.
I can merge this PR as is if you want. Then as we said yesterday we can work on the other case separately/if needed. Just let me know.

alexhenrie:
Hi @micheleCTDE, I think this PR is worth merging on its own. Later this week I will work on new mod_int_(short|long|signed|unsigned) options to support adding or removing `int` before the size/sign keyword rather than after. I think we need new options for three reasons:

1. `mod_int_short=add` would allow converting `short` to `int short`, which some users might prefer over `short int`.

2. The combination of `mod_int_short=remove` and `mod_short_int=add` would allow changing `int short` to `short int`.

3. If we make mod_*_int remove `int` in any position, it's not clear whether `mod_short_int` or `mod_unsigned_int` is responsible for removing the `int` in `unsigned int short foo`. With new options, `mod_short_int` would add or remove `int` after `short` and `mod_int_short` would add or remove `int` before `short`.

micheleCTDE:
ok, that soulds very good. Thanks for the explanation @alexhenrie

micheleCTDEAdmin (approved):

PR #3730, opened 2022-07-03
Replace set_chunk_type{_real}() with {Chunk::}SetType{Real}()

micheleCTDE:
As per title.

### DIFF
--- a/documentation/htdocs/options_Align.html
+++ b/documentation/htdocs/options_Align.html
@@ -3,9 +3,9 @@
 align_pp_define_gap
 The minimum space between label and value of a preprocessor define
 
-#define set_chunk_type(pc, tt)      do { \
+#define SetType(tt)      do { \
       LOG_FUNC_CALL(); \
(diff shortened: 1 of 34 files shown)


micheleCTDEAdmin (approved):

documentation/htdocs/options_Align.html
```diff
@@ -3,7 +3,7 @@
 align_pp_define_gap
 The minimum space between label and value of a preprocessor define
 
-#define set_chunk_type(pc, tt)      do { \
+#define pc->SetType(tt)      do { \
```
micheleCTDEAdmin, line 6: incorrect format, although inside a html file.

micheleCTDEAdmin (changes requested):
See comment in options_Align.html

documentation/htdocs/options_Align.html
micheleCTDEAdmin: fixed

micheleCTDEAdmin (approved):

PR #3731, opened 2022-07-05
Fix compilation with older versions of GCC

dbartolini:
This PR fixes compilation with older GCC versions (>= 4.9).


gmaurel (approved):
Thanks, but I cannot check if it is correct

PR #3732, opened 2022-07-06
Add mod_int_* options

alexhenrie:
See previous discussion at https://github.com/uncrustify/uncrustify/pull/3725#issuecomment-1172827591 and https://github.com/uncrustify/uncrustify/pull/3729#issuecomment-1173163510

This turned into a much bigger rewrite than I expected, but I'm pretty happy with the result. The new mod_int_* options allow, for example, converting `short` to `int short` and changing `int short` to `short int`. Furthermore, I've added a `mod_int_prefer_int_on_left` option to specify which `int` keyword to keep in case of conflict. The refactoring to support these options also allows the old mod_*_int options to be a little more aggressive about enforcing the user's preferred style.


gmaurel (approved):
Fine

alexhenrie:
Thank you!

PR #3737, opened 2022-07-09
Code impr/get parent type

micheleCTDE:
As per title


micheleCTDE:
Sorry, I messed up something in my local repo. I will create a new PR after cleaning up.

Issue #3734, opened 2022-07-08
Could nl_oc_msg_args be configured with a line width?


PoeticPete:
Currently `nl_oc_msg_args` forces all params to be in a new line.

```
[Object param1:nil param2:nil]; // input
[Object param1:nil
        param2:nil]; // output
```

Would it be possible set a line width to determine if parameters should be new lined? It would be nice to keep short functions on one line.

PR #3739, opened 2022-07-09
Support conditionally newlining objc params based on a number of args

PoeticPete:
Currently we only support an "all or nothing" approach to new-lining Objective-C message parameters. This is not ideal because some would argue functions such as `[object param1:nil param2:nil]` are more readable in a single line.

This change supports configuring `nl_oc_msg_args` based on a number of parameters, a new option `nl_oc_msg_args_min_params`.

Issue: https://github.com/uncrustify/uncrustify/issues/3734

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -2027,7 +2027,7 @@ nl_template_end                 = false    # true/false
 nl_oc_msg_args                  = false    # true/false
 
 # (OC) Minimum number of Objective-C message parameters before applying nl_oc_msg_args.
-nl_oc_msg_args_min_params       = 0
+nl_oc_msg_args_min_params       = 0        # unsigned number
 
 # Add or remove newline between function signature and '{'.
 nl_fdef_brace                   = ignore   # ignore/add/remove/force/not_defined
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -2026,6 +2026,9 @@ nl_template_end                 = false    # true/false
 # See nl_oc_msg_leave_one_liner.
 nl_oc_msg_args                  = false    # true/false
 
+# (OC) Minimum number of Objective-C message parameters before applying nl_oc_msg_args.
+nl_oc_msg_args_min_params       = 0
+
 # Add or remove newline between function signature and '{'.
 nl_fdef_brace                   = ignore   # ignore/add/remove/force/not_defined
 
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -3453,9 +3453,8 @@ static void newline_oc_msg(Chunk *start)
    {
       return;
    }
-
    // Get count of parameters
-   size_t  parameter_count = 0;
+   size_t parameter_count = 0;
 
    for (Chunk *pc = start->GetNextNcNnl(); pc->IsNotNullChunk(); pc = pc->GetNextNcNnl())
    {
@@ -3470,9 +3469,10 @@ static void newline_oc_msg(Chunk *start)
       }
    }
 
-   int    min_params = options::nl_oc_msg_args_min_params();
+   int min_params = options::nl_oc_msg_args_min_params();
 
-   if (parameter_count < min_params && min_params != 0) {
+   if (parameter_count < min_params && min_params != 0)
+   {
       return;
    }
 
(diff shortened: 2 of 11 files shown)


documentation/htdocs/config.txt
```diff
@@ -2026,6 +2026,9 @@ nl_template_end                 = false    # true/false
 # See nl_oc_msg_leave_one_liner.
 nl_oc_msg_args                  = false    # true/false
 
+# (OC) Minimum number of Objective-C message parameters before applying nl_oc_msg_args.
+nl_oc_msg_args_min_params       = 0
```
micheleCTDE, line 2030: Missing "# unsigned number" after the option. Refer to documentation/htdocs/default.cfg where instead you add it correctly.
Please amend the other files in similar manner.

micheleCTDE (changes requested):
Hi and welcome.
Beside the failing tests to fix, please also amend as per comment below.

PoeticPete:
I got some weird html output for the `--tracking_space` test: https://pastebin.com/rRKP4Pav

Checking out master, rebuilding, then checking out my branch and rebuilding seems to have solved the issue. Everything is working now, but if you know what happened, I'd be curious to know.

PoeticPete:
@micheleCTDE thanks for the quick turnaround. I've fixed the failing tests and added the commends.

src/newlines.cpp
```diff
@@ -3453,6 +3453,28 @@ static void newline_oc_msg(Chunk *start)
...
+
+      if (pc->Is(CT_OC_MSG_NAME))
+      {
+         parameter_count++;
+      }
+   }
+
+   int min_params = options::nl_oc_msg_args_min_params();
+
+   if (parameter_count < min_params && min_params != 0)
```
micheleCTDEAdmin, line 3474: I didn't spot this yesterday. As you can see in the rest of the code, we have a specific style for indenting multiple conditional expressions. ```&&``` will go to a new line and some extra spaces are needed in the first line. You can refer to any if or while statement with multiple conditional expressions in it for an example.

tests/expected/oc/50631-nl_oc_msg_args_min_params.m
```diff
@@ -0,0 +1,8 @@
+static void function() {
+	[object param1:nil
+	 param2:nil
+	 param3:nil
+	 param4:nil];
+
+	[object param1:nil param2:nil param3:nil];
```
micheleCTDEAdmin, line 7: I don't use OC, so maybe I am saying something wrong. Based on the description of the new option, I would expect functions/objects/[whatever is the right term in OC] to be broken up in multiple lines when they have 3 (in this specific test) or more parameters. So why is this second case not broken up? I would expect this
```
[object param1:nil param2:nil];
```
not to break up, while the two examples in the test file to be broken up.
Am I missing something?

micheleCTDEAdmin (changes requested):
Please see further feedback

micheleCTDEAdmin:
> I got some weird html output for the `--tracking_space` test: https://pastebin.com/rRKP4Pav. Expected output is here: https://github.com/uncrustify/uncrustify/blob/master/tests/cli/output/tracking_space.html
> 
> Checking out master, rebuilding, then checking out my branch and rebuilding seems to have solved the issue. Everything is working now, but if you know what happened, I'd be curious to know.

Not 100% sure but I think it was related to the failing cli-options test. I also had some problems occasionally with the tracking file in the past, although I did not sit down to investigate it in detail. Usually updating the cli option results solves the issue for me.

tests/expected/oc/50631-nl_oc_msg_args_min_params.m
PoeticPete: Good catch :) We're off by one here because param1 counts as a function name chunk. I've updated the code to instead look for colons to count parameters. Updated the tests as well.

src/newlines.cpp
PoeticPete: Oh nice, just updated. Maybe we should add an uncrustify rule for that haha

PoeticPete:
Fixed format, and updated logic to fix the off by 1 issue. Thanks for the thorough review @micheleCTDEAdmin!

micheleCTDEAdmin (approved):
Looks good, thanks!

micheleCTDE (approved):

PoeticPete:
@micheleCTDE any idea when the next release is going to be?

micheleCTDE:
Guy makes a release every May and November, so it should be a while before the next release.

PR #3741, opened 2022-07-10
Fix detection of multidimensional array type in Vala

dbartolini:
Fixes detection of multidimensional arrays, making sp_*_mdatype_commas work in Vala too.


micheleCTDEAdmin:
Looks good, thanks Daniele.

PR #3743, opened 2022-07-11
Fix mod_remove_extra_semicolon generating invalid code in Vala

dbartolini:
Fixes Uncrustify generating invalid lambda code in some cases when ``mod_remove_extra_semicolon = true`` in Vala.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -603,8 +603,7 @@ void do_symbol_check(Chunk *prev, Chunk *pc, Chunk *next)
          return;
       }
 
-      if (  language_is_set(LANG_CS)
-         && (  pc->Is(CT_LAMBDA)
+      if (  (  pc->Is(CT_LAMBDA)
             || pc->Is(CT_DELEGATE))
          && next->Is(CT_BRACE_OPEN))
       {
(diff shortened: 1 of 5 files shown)


src/combine.cpp
```diff
@@ -603,8 +603,7 @@ void do_symbol_check(Chunk *prev, Chunk *pc, Chunk *next)
          return;
       }
 
-      if (  language_is_set(LANG_CS)
```
micheleCTDEAdmin, line 606: Ciao Daniele,
would it be better to add a check for the language to be Vala, rather than removing the check for C#? Just wondering if removing the check for C# would have any side effects on other languages who ue lambda functions. Something like:
```
if (  (language_is_set(LANG_CS) ||   language_is_set(LANG_VALA)) && (  pc->Is(CT_LAMBDA) ....
```

dbartolini: Ciao Michele,
that check is inside a ``if (language_is_set(LANG_CS | LANG_VALA))`` block so it should only affect those two languages, right? Actually, I added that and other LANG_CS checks myself in a previous PR and now I am gradually disabling them and fixing Vala issues as they pop.

micheleCTDEAdmin: Ah perfect. I was a bit in a rush this morning and didn't delved further in the code, so I just posted a comment to ask. 
Looks good, will merge it.

micheleCTDEAdmin (approved):

PR #3746, opened 2022-07-14
Rename Chunk::parent to Chunk::m_parent and make it a protected member

micheleCTDE:
As per title


micheleCTDE:
Sorry, didn't intend to close this... PR #3747 is a equal replacement for this one.

PR #3749, opened 2022-07-15
Rename 'pp_space' option to 'pp_space_after' and add 'pp_space_before' option

micheleCTDE:
This allows more flexibility in preprocessor formatting.


src/option.cpp
```diff
@@ -392,6 +389,20 @@ bool process_option_line_compat_0_74(const std::string &cmd,
...
 } // process_option_line_compat_0_74
 
+
+bool process_option_line_compat_0_75(const std::string &cmd, const char *filename)
+{
+   if (cmd == "pp_space")
+   {
+      OptionWarning w{ filename, OptionWarning::MINOR };
+      w("option '%s' is deprecated; did you want to use '%s' instead?",
+        cmd.c_str(), options::pp_space_after.name());
```
gmaurel, line 399: and pp_space_before ??

gmaurel (changes requested):
some little questions

forUncrustifySources.cfg
```diff
@@ -166,7 +166,7 @@ pos_enum_comma                            = trail_force
 
 # Preprocessor options
 pp_indent                                 = remove
-pp_space                                  = remove
+pp_space_after                            = remove
```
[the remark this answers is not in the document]
micheleCTDE: 'pp_space' has been renamed into 'pp_space_after' and that is what is needed to preserve the same behavior.
'pp_space_before' offers additional functionality, but its default value is set to 'ignore', so there is no need to add it in this file, IMO.

src/option.cpp
micheleCTDE: 'pp_space' has been renamed into 'pp_space_after' and that is what is needed to preserve the same behavior. 'pp_space_before' offers additional functionality only.
Nevertheless I have amended the message to inform the user of the new options.

micheleCTDE:
> some little questions

Ready for a new review :-)

gmaurel (approved):

micheleCTDE:
Great, thanks for reviewing @gmaurel

PR #3750, opened 2022-07-15
Fix .editorconfig's src/* pattern

dbartolini:
Removes unnecessary braces that prevent (at least some versions of) vim-editorconfig and probably other editors/plugins from applying the config.


micheleCTDEAdmin (approved):
Looks good

Issue #3752, opened 2022-07-15
indent_continue breaks nested continuation indents?


dbartolini:
Not sure if it's intended behavior but I think `indent_continue = 4` should make continuations 4-spaces wide relative to the start of the statement being continued.

Version: `Uncrustify_d-0.75.1_f`.

Config:
```
input_tab_size = 4
output_tab_size = 4
indent_columns = 4
indent_with_tabs = 0

indent_ignore_first_continue = false
indent_continue = 4
```
Input:
```
namespace ns
{
first(foo
, bar
, second(foo
, bar
)
);

first(begin
, end
, [](foo, bar) {
second(foo
, bar
);
}); 
}
```
Output:
````{verbatim}
namespace ns
{
first(foo
    , bar
    , second(foo
    , bar           
    )               
    );

first(begin
    , end
    , [](foo, bar) {
        second(foo
        , bar 
        );    
    });
}
````
Expected:
```
first(foo
    , bar
    , second(foo
        , bar
        )
    );

first(begin
    , end
    , [](foo, bar) {
        second(foo
            , bar
            );
    });
}
```

PR #3765, opened 2022-07-26
Fix indent_cpp_lambda_body in some circumnstances

dbartolini:
This PR fixes the indentation of lambda bodies in cases similar to the following:
```
a([]() {
        b();
    });
```

Previously, the above would have been indented like this:
```
a([]() {
    b();
});
```
which is inconsistent with how an almost identical expression is and was indented instead (you would've put `);` below `}` to trigger the correct lambda body indentation):
```
a([]() {
        b();
    }
);
```

Please notice that some indents are still wrong (see tests/expected/cpp/60055-issue_3116.cpp:151 or tests/expected/cpp/60056-issue_3116-2.cpp:3) but that is a different issue (#3752) that will be fixed in a separate PR.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1759,7 +1759,8 @@ void indent_text()
                (options::indent_continue() > 0)
                && enclosure
                && linematch
-               && toplevel;
+               && toplevel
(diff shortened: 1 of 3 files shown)


gmaurel (approved):
Thanks

Issue #2656, opened 2020-02-12
FUNC_PROTO not correctly identified when OC block is present as param


samishchandra:
`FUNC_PROTO` not correctly identified when OC block is present as param.

### ⬇️Input:
```
void Events(NSString *    identifier, void (^handler)());
```

### ✅Expected Output:
```
void Events(NSString *identifier, void (^handler)());
```

### ❌Actual Output:
```
void Events(NSString *  identifier, void (^handler)());
```

### ⚙️Config options used:
> sp_after_ptr_star               = remove

### Debugging info (before fix)
```
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>               TYPE|               NONE|     PARENT_NOT_SET[  1/  1/  5/  0][0/0/0][      70000][0-0] void
#   1>      FUNC_CTOR_VAR|               NONE|     PARENT_NOT_SET[  6/  6/ 12/  1][0/0/0][    1800000][0-0]      Events
#   1>        FPAREN_OPEN|      FUNC_CTOR_VAR|     PARENT_NOT_SET[ 12/ 12/ 13/  0][0/0/0][  100000000][0-0]            (
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 13/ 13/ 21/  0][0/1/0][20000040000][0-0]             NSString
#   1>              ARITH|               NONE|     PARENT_NOT_SET[ 22/ 22/ 23/  1][0/1/0][20100000000][0-0]                      *
#   1>               WORD|               NONE|     PARENT_NOT_SET[ 24/ 27/ 37/  4][0/1/0][20000040000][0-0]                        identifier
```

### Debugging info (after fix)
```
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp      Flag   Nl  Text
#   1>               TYPE|         FUNC_PROTO|     PARENT_NOT_SET[  1/  1/  5/  0][0/0/0][      70000][0-0] void
#   1>         FUNC_PROTO|               NONE|     PARENT_NOT_SET[  6/  6/ 12/  1][0/0/0][          0][0-0]      Events
#   1>        FPAREN_OPEN|         FUNC_PROTO|     PARENT_NOT_SET[ 12/ 12/ 13/  0][0/0/0][  100000000][0-0]            (
[...]

PR #3769, opened 2022-07-27
Fix func proto with qualifier in OC

### LINKED ISSUE #3766: FUNC_PROTO not correctly identified when OC block is present as param with qualifier
[//]: # 
[//]: # 
[//]: # 
[//]: # 
[//]: # "      or"

Similar issue to #2656, but with qualifiers. See example below:

cfg file
```
sp_arith                        = force
```

input
```
void test(NSString *param1, void (^_Nullable completionBlock)(int));
```

output
```
void test(NSString * param1, void (^_Nullable completionBlock)(int));
```

The * is being detected incorrectly as ARITH.

Will have a PR up soon

PoeticPete:
This PR fixes #3766 

Similar issue to #2656, but with qualifiers. See example below:

cfg file
```
sp_arith                        = force
```

input
```
void test(NSString *param1, void (^_Nullable completionBlock)(int));
```

output
```
void test(NSString * param1, void (^_Nullable completionBlock)(int));
```

The * is being detected incorrectly as ARITH.

### DIFF
--- a/src/combine_tools.cpp
+++ b/src/combine_tools.cpp
@@ -150,6 +150,14 @@ bool can_be_full_param(Chunk *start, Chunk *end)
          }
          Chunk *tmp2 = tmp1->GetNextNcNnl(E_Scope::PREPROC);
 
+         if (tmp2->Is(CT_QUALIFIER))
+         {
+            // tmp2 is the "nullable" qualifier in this case:
+            // void (^nullable name)(params)
+            // skip the qualifier
+            tmp2 = tmp2->GetNextNcNnl(E_Scope::PREPROC);
+         }
+
          if (tmp2->IsNullChunk())
          {
             return(false);
--- a/tests/config/oc/3766.cfg
+++ b/tests/config/oc/3766.cfg
@@ -1,0 +1,3 @@
+sp_arith                        = force
+sp_before_ptr_star              = force
+sp_after_ptr_star               = remove
--- a/tests/expected/oc/50909-3766.m
+++ b/tests/expected/oc/50909-3766.m
@@ -1,0 +1,15 @@
+void test(NSString *param1, void (^_Nullable completionBlock)(int));
+
+void test(NSString *param1, void (^_Nullable completionBlock)(int));
+
+void test(void (^_Nullable completionBlock)(int), NSString *param1);
+
+void test(void (^_Nullable completionBlock)(int), NSString *param1, void (^_Nullable completionBlock)(int));
+
+void test(
+	void (^_Nullable completionBlock)(int),
+	NSString *param1,
+	void (^_Nullable completionBlock)(int)
+	);
+
+void test(NSString *param1, void (^_Nullable completionBlock)(NSString *, NSString *, void (^_Nullable completionBlock2)(int)));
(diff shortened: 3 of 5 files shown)


PoeticPete:
I have another PR that I want to open here: https://github.com/PoeticPete/uncrustify/pull/2

However Github doesn't automatically point PRs to the new HEAD with forked repos. Would it be possible to get write access to non-master branches?

micheleCTDEAdmin:
Hi @PoeticPete,
would you mind squashing the commits into one? No point into committing temporary log stuff that gets removed in subsequent commits, it just adds noise :-)

Regarding write access to non-master branches, I will leave this to @guy-maurel, since he is the official maintainer. But I am not sure if that helps with the problem you mentioned. For example, if I need to add more commits to a PR, I add then to the original branch and push and the branch HEAD seems to get updated. And I avoid having PR branches on top of other PR branches, it seems to be a bit messy. I usually wait till the first PR is merged, then open a second one. Or adds the commits on the branch of the first PR alternatively.

micheleCTDEAdmin (changes requested):
Please squash to avoid clutter.

PoeticPete:
@micheleCTDEAdmin there's a way to squash and merge on github. We might need to enable it for this repo: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

I'll do it manually for the time being

PoeticPete:
There's a good chance I'm just missing something for forked repos.. but github does automatically switch the base branch of stacked PRs for you.

>As soon as you merge and delete the A branch, the base branch of the B pull request — which was A originally — will switch to main.
https://pspdfkit.com/blog/2021/how-to-handle-stacked-pull-requests-on-github/

micheleCTDEAdmin:
> There's a good chance I'm just missing something for forked repos.. but github does automatically switch the base branch of stacked PRs for you.

Github does indeed update the base point of the B branch, but will show the differences between B's HEAD and the common ancestor with the main/master branch. Depending on what was done in the A branch, this may show more commits than necessary.

```
Remember that this command shows the contents of B, starting at the common ancestor between main and B. In other words, this is equivalent to git diff $(git merge-base main B) B. The problem is that the common ancestor of both branches is the commit where the A branch started diverging from main.
```

In most of my projects, I work with self-hosted gitea rather than github and we avoid merge commit on purpose. We instead rebase and fast-forward. This may seems a bit problematic (especially if commits are GPG-signed and we need to ask users to rebase them), but comes with two main benefit:
1. history is linear. It's a great thing when you often need to go back in history to track/debug stuff
2. merge commits can sometimes (not often) introduce subtle bugs because two independent PR may affect the same area and what was working in one of the PR may no longer work after merging on top of the other (yes, it happened to us). By rebasing and testing again, this can be caught before committing to the main/master branch.

micheleCTDEAdmin:
> @micheleCTDEAdmin there's a way to squash and merge on github. We might need to enable it for this repo: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests
> 
> I'll do it manually for the time being

Yes, it is indeed possible to do it from github. But I consider it bad practice because the resulting commit will get a new hash and the user local repo may get out of sync. For example if the user has already perform a local merge to its main, their copy and the online repo would see different commit hashes (even if the changes were the same).

micheleCTDEAdmin (approved):
Looks good, thanks.

PR #3773, opened 2022-08-02
sort (or not) the tracking info.

guy-maurel:
replace the previous PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 A source code beautifier for C, C++, C#, Objective-C, D, Java, Pawn and Vala.
 
 ## Features
-* Highly configurable - 810 configurable options as of version 0.75.0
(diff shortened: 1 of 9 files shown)


micheleCTDE:
@guy-maurel shouldn't we update all related files when we adda new option? If I am not wrong, the release process script will do that anyway, but better to update the required files before the actual release to keep things up to date, IMO.

gmaurel:
Yes, but doing ctest --output-on-failure -I 1
I don't get any warning.
What is wrong?

micheleCTDEAdmin:
ctest will simply run the tests. I don't recall any test that checks for non-updated files in case of new option.
But if you use the scripts/release_tool.py script on a copy of your repo, you can get the additional changes required. Then you can copy those to your PR branch and add them to this PR.

gmaurel (approved):

PR #3776, opened 2022-08-05
Keep vars with '[]' types in variable blocks

fwojcik:
Variable definitions which have square braces in their types are not parsed as being variable definitions for purposes of newline addition.

For example, ```char[] ca = cast(char[])iterator.next;``` would not be considered a variable definition, and ```nl_var_def_blk_end``` would not handle it correctly.

For another example:
```
         Token  tok;
         int    linnum;
         char[] filespec;
         Loc    loc = this.loc;
```
would be considered 2 different variable definition blocks.

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -2082,6 +2082,11 @@ static Chunk *newline_def_blk(Chunk *start, bool fn_top)
             next = next->GetNextNcNnl();
          }
 
+         while (next->Is(CT_TSQUARE))
+         {
+            next = next->GetNextNcNnl();
+         }
+
          if (next->IsNullChunk())
          {
             break;
--- a/tests/expected/d/40000-HashMap.d
+++ b/tests/expected/d/40000-HashMap.d
@@ -547,10 +547,11 @@ class HashMap
                 rehash();
 
             HashEntry[] tab;
+
             volatile tab = table;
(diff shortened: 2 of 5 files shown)


fwojcik:
Hi! I have _many_ bug fixes and features (about 40) that I'm going to start queuing up, ranging from tiny doc fixes to some significant but isolated refactoring. This is going to be the first of a few PRs that I submit today. Please let me know if my PRs could be improved to make your lives easier, so I can adjust my workflow for future ones.

micheleCTDEAdmin:
> Hi! I have _many_ bug fixes and features (about 40) that I'm going to start queuing up, ranging from tiny doc fixes to some significant but isolated refactoring. This is going to be the first of a small number of PRs that I submit today. Please let me know if my PRs could be improved to make your lives easier, so I can adjust my workflow for future ones.

Hi and welcome @fwojcik,
it would be helpful if you could add more info showing what was wrong. For example, regarding the example above, you could show what the wrong result looks like and what the expected result should be. Then you could add a new test in the code to check that the new code does what is expected correctly.
It would make reviewing PRs easier for us.

Please do the same for whatever other PR you would like to submit. Looking forward for your contributions :-)

micheleCTDEAdmin (approved):
Looks good thanks. The changes in the existing test files seems enough to provide a good check.

PR #3777, opened 2022-08-05
Fix behavior of nl_var_def_blk_end to match docs

fwojcik:
Variable definiton blocks which are "not at the top of a function body" should have
newlines after them controlled by the ```nl_var_def_blk_end``` option, according to
```options.h```. Currently it seems like all variable definition blocks are controlled by
```nl_func_var_def_blk```, because the value of ```first_var_blk``` is not set correctly.

This behavior change was introduced in commit
d76e41304b628eba98debd70e05f4ab36797ea6c. It seems like it would be hard for this to
be an accidental change, so perhaps there is simply some misunderstanding about the
phrase "at the top if a function body".

The unit test changes I chose to make were a balance of changing configs versus
expected results, and wanting to test the ```nl_var_def_blk_end``` option enabled versus
disabled. I am happy to do this some other way.

Note also that this change breaks the `java` unit tests  unless the fix in https://github.com/uncrustify/uncrustify/pull/3776 is also applied!


micheleCTDEAdmin (changes requested):
This PR does not pass the automated tests. Please review and fix accordingly.

fwojcik:
Now that #3776 has been merged, I expect these unit tests to pass. However, please do not yet merge this PR.

fwojcik:
There is a surprising amount of complexity when it comes to newline processing around variable blocks. This diff does not go nearly far enough. I have a new diff working, but it is not quite complete. When it is, I'll post it with a very long explanation, but under a new PR to avoid the clutter from this one.

PR #3778, opened 2022-08-05
Stop indent_ignore_asm_block indenting pp lines

fwojcik:
When asm blocks are indented as code and indent_ignore_asm_block is true, then
preprocessor lines inside the asm block are indented along with the asm body.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -2613,8 +2613,11 @@ void indent_text()
 
          do
          {
-            pc->column = pc->orig_col + move;
-            pc         = pc->GetNext();
+            if (!pc->TestFlags(PCF_IN_PREPROC))
+            {
+               pc->column = pc->orig_col + move;
+            }
+            pc = pc->GetNext();
          } while (pc != tmp);
 
          reindent_line(pc, indent_column);
--- a/tests/config/cpp/asm_block_pp.cfg
+++ b/tests/config/cpp/asm_block_pp.cfg
@@ -1,0 +1,2 @@
+indent_columns                  = 8
+indent_ignore_asm_block         = true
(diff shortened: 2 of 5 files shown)


fwojcik:
An example of this failure is in the included unit test, where without this fix the configuration of 
```
indent_columns                  = 8
indent_ignore_asm_block         = true
```
and input file of:
```cpp
void add128( uint64_t & rlo, uint64_t & rhi, uint64_t addlo ) {
#if defined(HAVE_X86_64_ASM)
  __asm__ ("addq %2, %0\n"
           "adcq $0, %1\n"
#if defined(__clang__)
             // clang cannot work properly with "g" and silently
             // produces hardly-workging code, if "g" is specified;
           : "+r" (rlo), "+r" (rhi)
           : "m" (addlo)
#else
           : "+g" (rlo), "+g" (rhi)
           : "g" (addlo)
#endif
    );
#else
    rlo += addlo;
    rhi += (rlo < addlo);
#endif
}
```
becomes:
```cpp
void add128( uint64_t & rlo, uint64_t & rhi, uint64_t addlo ) {
#if defined(HAVE_X86_64_ASM)
        __asm__ ("addq %2, %0\n"
                 "adcq $0, %1\n"
      #if defined(__clang__)
                   // clang cannot work properly with "g" and silently
                   // produces hardly-workging code, if "g" is specified;
                 : "+r" (rlo), "+r" (rhi)
                 : "m" (addlo)
      #else
                 : "+g" (rlo), "+g" (rhi)
                 : "g" (addlo)
      #endif
        );
#else
        rlo += addlo;
        rhi += (rlo < addlo);
#endif
}
```
Note that the inner preprocessor lines have become indented, which was not requested.

micheleCTDE (approved):

micheleCTDE:
Looks good. Thanks for adding the extra explanation, it was very useful.

PR #3779, opened 2022-08-05
Stop altering spaces when nl cont disabled

fwojcik:
If ```disable_processing_nl_cont=true``` and ```sp_before_nl_cont=add```, then spaces will be
added before continuation characters regardless. Furthermore, every loop through
```do_space()``` will add another one, so in some unlucky cases this will lead to hilarious diffs like:

```diff
-#define HASH_FLAGS                                     \
+#define HASH_FLAGS                                                                                                 \
```

### DIFF
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -1984,7 +1984,7 @@ static bool parse_macro(tok_ctx &ctx, Chunk &pc, const Chunk *prev_pc)
                && nl))
          && pc.str.size() > 0)
       {
-         pc.SetType(CT_IGNORED);
+         pc.SetType(CT_PP_IGNORE);
          return(true);
       }
       else if (nl)
--- a/tests/config/cpp/disable_nl_cont.cfg
+++ b/tests/config/cpp/disable_nl_cont.cfg
@@ -3,4 +3,4 @@ sp_compare                    = force
 sp_assign                     = force
 sp_arith                      = force
 indent_class                  = true
-sp_before_nl_cont             = ignore
+sp_before_nl_cont             = add


micheleCTDEAdmin:
I think I saw this problems in the past too. As commented on another PR of yours, please add some example of wrong-expected use case and add a test for it in the code.

fwojcik:
For this PR, I decided to leverage the existing test cases. If you still prefer that I add more and/or don't modify the existing ones, then please let me know explicitly which to do and I'll update my code.

To give one example of a failure, this is from the existing test cases where the config was updated as in this PR, so it becomes:
```
disable_processing_nl_cont    = true
sp_compare                    = force
sp_assign                     = force
sp_arith                      = force
indent_class                  = true
sp_before_nl_cont             = add
```
and the input file (`tests/input/cpp/disable_macro.cpp`):
```cpp
#include <stdio.h>

// this macro should NOT be modified ...
#define CHK(...)   \
  do               \
  {                \
      a+=1;        \
        a=b=0;     \
          c<<1;    \
  } while (0+0)


// ... whereas this should be indented and formatted
int main()
{
	int a,b,c=0;
	if (a<c)
	{
	c+=1;
	}
	a=b=0;
	c<<1;
	CHK;
}
```
becomes:
```cpp
#include <stdio.h>

// this macro should NOT be modified ...
#define CHK(...)    \
  do                \
  {                 \
      a+=1;         \
        a=b=0;      \
          c<<1;     \
  } while (0+0)


// ... whereas this should be indented and formatted
int main()
{
	int a,b,c = 0;
	if (a < c)
	{
		c += 1;
	}
	a = b = 0;
	c << 1;
	CHK;
}
```
which has a diff of:
```diff
diff --git a/home/fwojcik/progs/uncrustify/tests/expected/cpp/33213-disable_macro.cpp b/home/fwojcik/progs/uncrustify/build/tests/results/cpp/33213-disable_macro.cpp
index 7e64f416..993eb784 100644
--- a/home/fwojcik/progs/uncrustify/tests/expected/cpp/33213-disable_macro.cpp
+++ b/home/fwojcik/progs/uncrustify/build/tests/results/cpp/33213-disable_macro.cpp
@@ -1,12 +1,12 @@
 #include <stdio.h>
 
 // this macro should NOT be modified ...
-#define CHK(...)   \
-  do               \
-  {                \
-      a+=1;        \
-        a=b=0;     \
-          c<<1;    \
+#define CHK(...)    \
+  do                \
+  {                 \
+      a+=1;         \
+        a=b=0;      \
+          c<<1;     \
   } while (0+0)
 
 
MISMATCH: cpp:33213
```

micheleCTDEAdmin (approved):
Looks good, thanks.

PR #3784, opened 2022-08-06
Remove unnecessary pp_space_before option.

micheleCTDE:
1. PR # 3749 introduced the option pp_space_before. This is not needed, since pp_indent was already doing the exact same thing. This PR removes such option.

2. Using pp_space_after = force and pp_space_count = 0 was actually using a count of 1 for each level. This has now been fixed: the actual value of pp_space_count is used in all cases.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -4720,34 +4720,15 @@ void indent_preproc()
                               ? pc->pp_level - pp_level_sub : 0;
 
       // Adjust the indent of the '#'
-      log_rule_B("pp_indent");
-
       if (options::pp_indent() & IARF_ADD)
       {
-         log_rule_B("pp_indent_count");
+         log_rule_B("pp_indent ADD");
          reindent_line(pc, 1 + pp_level * options::pp_indent_count());
       }
       else if (options::pp_indent() & IARF_REMOVE)
       {
-         log_rule_B("pp_indent");
+         log_rule_B("pp_indent REMOVE");
          reindent_line(pc, 1);
-      }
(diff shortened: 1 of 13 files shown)


src/indent.cpp
```diff
@@ -4717,36 +4717,17 @@ void indent_preproc()
                               ? pc->pp_level - pp_level_sub : 0;
 
       // Adjust the indent of the '#'
-      log_rule_B("pp_indent");
-
       if (options::pp_indent() & IARF_ADD)
       {
-         log_rule_B("pp_indent_count");
+         log_rule_B("pp_indent ADD");
```
gmaurel, line 4722: Please use the exact text from the option name

gmaurel (changes requested):
see the comment

src/indent.cpp
micheleCTDE: @gmaurel 
Well, IMO it is correct to use 'pp_indent' rather than 'pp_indent_count'. 
''pp_indent' determines whether we add or remove indentation here. 'pp_indent_count' only gives the amount of indent when 'pp_indent' is set to add/force the indentation.
You can see the genetic `log_rule_B("pp_indent");` has been removed from line 4720 and instead added to the body of the if/else-if only when executed. If we don't execute any of the if/else-if bodies, we shouldn't print the log rule at all. I also add ADD or REMOVE to specify the actual path of the if/else-if block

gmaurel (approved):

micheleCTDEAdmin:
@gmaurel  Thanks for reviewing and approving the PR

PR #3788, opened 2022-08-08
Fix typo in log_rule_B statement

fwojcik:
This makes the logged rule name match the real rule name for
"mod_remove_duplicate_include".


micheleCTDEAdmin (approved):
Looks good, thanks.

PR #3789, opened 2022-08-08
Makes mod_full_brace_if keep commented one liners

fwojcik:
If `mod_full_brace_if = add` and an unbraced `if()` statement has a trailing comment,
then the added brace will be on a new line after the comment. However, this is true
even when `nl_if_leave_one_liners = true`.

This change will keep the added brace on the same line as the `if()`, before the
comment, if it was a one-liner that was configured to be left.

Without this patch, the result looks like:
```cpp
unsigned a;

int foo(void) {
        if (a == 1) { printf("It's 1\n"); }
}

int bar(void) {
        if (a == 2) { printf("It's 2\n"); /* comment */
        }
}
```

### DIFF
--- a/src/braces.cpp
+++ b/src/braces.cpp
@@ -875,11 +875,13 @@ static void convert_vbrace(Chunk *vbr)
       /*
        * If the next chunk is a comment, followed by a newline, then
        * move the brace after the newline and add another newline after
-       * the close brace.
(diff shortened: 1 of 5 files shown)


micheleCTDEAdmin (approved):
Excellent! Thanks.

PR #3791, opened 2022-08-08
Fix trailing cmt alignment when braces move

fwojcik:
If a newline is deleted before a brace, the brace may move to be before a comment
which was on the previous line. For example: `do //comment\n{` may become
`do { //comment`. This breaks the existing logic for aligning comments.

This change skips over any tokens that are currently earlier in the line than a to-be-aligned comment, but are originally from later lines. In this way, `prev` will point to the previous token from the same original line, and so accurate original spacing can be determined.

Without this change, the included test case fails like:
```
 2/14 Test  #2: c ................................***Failed   30.76 sec
Tests: ['c']
Processing /home/fwojcik/progs/uncrustify/tests/c.test
diff --git a/home/fwojcik/progs/uncrustify/tests/expected/c/00470-align_trailing_do_cmt.c b/home/fwojcik/progs/uncrustify/build/tests/results/c/00470-align_trailing_do_cmt.c
index 7c21b3d9..453ba441 100644
--- a/home/fwojcik/progs/uncrustify/tests/expected/c/00470-align_trailing_do_cmt.c
+++ b/home/fwojcik/progs/uncrustify/build/tests/results/c/00470-align_trailing_do_cmt.c
@@ -4,7 +4,7 @@ int foo(void) {
                r = bar();
        } while (r);
 
-       do { // this comment should be aligned with the one above
+       do {      // this comment should be aligned with the one above
                r = baz();
        } while (r);
 }
MISMATCH: c:00470
389 / 390 tests passed
1 tests did not match the expected outputm
```

I feel like this change is a little _too_ targeted and special-case; my instinct is that a more general fix would be better so that when braces are moved it won't break other options also, not just comment-aligning. However, when I tried a patch like:
```diff
diff --git a/src/newlines.cpp b/src/newlines.cpp
index f9eb105c..9854005b 100644
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -874,7 +874,18 @@ void newline_del_between(Chunk *start, Chunk *end)
          || start->Is(CT_DO)
          || start->Is(CT_ELSE)))
    {
+      // move the brace to be just after start; this adjusts its orig_col* variables
       end->MoveAfter(start);
+      // bump out the orig_col* variables for the remaining chunks on this line
+      size_t delta = end->orig_col_end - start->orig_col_end;
+      pc = end->GetNext();
+      while (  !pc->IsNewline()
+            && (pc->orig_line == start->orig_line))
+      {
+          pc->orig_col     += delta;
+          pc->orig_col_end += delta;
+          pc                = pc->GetNext();
+      }
    }
 } // newline_del_between
 ```
many things broke in the unit tests. So maybe this patch is OK?

### DIFF
--- a/src/align_trailing_comments.cpp
+++ b/src/align_trailing_comments.cpp
@@ -185,14 +185,12 @@ void align_right_comments()
       {
          if (pc->GetParentType() == CT_COMMENT_END)
          {
-            Chunk *prev = pc->GetPrev();
-
             log_rule_B("align_right_cmt_gap");
 
-            if (pc->orig_col < prev->orig_col_end + options::align_right_cmt_gap())
+            if (pc->orig_prev_sp < options::align_right_cmt_gap())
             {
-               LOG_FMT(LALTC, "NOT changing END comment on line %zu (%zu <= %zu + %u)\n",
-                       pc->orig_line, pc->orig_col, prev->orig_col_end,
+               LOG_FMT(LALTC, "NOT changing END comment on line %zu (%u < %u)\n",
+                       pc->orig_line, pc->orig_prev_sp,
                        options::align_right_cmt_gap());
             }
             else
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -274,7 +274,7 @@ void align_to_column(Chunk *pc, size_t column)
       else if (almod == align_mode_e::KEEP_REL)
       {
          // Keep same relative column
-         auto orig_delta = static_cast<int>(pc->orig_col) - static_cast<int>(prev->orig_col);
+         int orig_delta = static_cast<int>(pc->orig_prev_sp) + static_cast<int>(prev->Len());
          orig_delta = max<int>(orig_delta, min_delta);  // keeps orig_delta positive
 
          pc->column = prev->column + static_cast<size_t>(orig_delta);
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -750,7 +750,7 @@ void output_text(FILE *pfile)
                      && prev->IsNotNullChunk()
                      && prev->nl_count == 0)
                   {
-                     int orig_sp = (pc->orig_col - prev->orig_col_end);
+                     int orig_sp = pc->orig_prev_sp;
 
                      if ((int)(cpd.column + orig_sp) < 0)
                      {
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -1797,6 +1797,7 @@ static bool parse_whitespace(tok_ctx &ctx, Chunk &pc)
    while (  ctx.more()
          && unc_isspace(ctx.peek()))
    {
+      int lastcol = ctx.c.col;
       ch = ctx.get();   // throw away the whitespace char
 
       switch (ch)
@@ -1825,8 +1826,7 @@ static bool parse_whitespace(tok_ctx &ctx, Chunk &pc)
          break;
 
       case '\t':
-         log_rule_B("input_tab_size");
-         pc.orig_prev_sp += calc_next_tab_column(cpd.column, options::input_tab_size()) - cpd.column;
+         pc.orig_prev_sp += ctx.c.col - lastcol;
          break;
 
       case ' ':
@@ -2732,6 +2732,9 @@ void tokenize(const deque<int> &data, Chunk *ref)
       }
       // Store off the end column
       chunk.orig_col_end = ctx.c.col - num_stripped; // Issue #1966 and #3565
+
+      // Make the whitespace we disposed of be attributed to the next chunk
+      prev_sp = num_stripped;
 
       // Add the chunk to the list
       rprev = pc;
--- a/tests/c.test
+++ b/tests/c.test
@@ -193,6 +193,8 @@
 00468  c/align_func_proto_star_amp-8.cfg          c/align_func_proto_star_amp.h
 00469  c/align_func_proto_star_amp-9.cfg          c/align_func_proto_star_amp.h
 
+00470  c/align_trailing_do_cmt.cfg                c/align_trailing_do_cmt.c
+
 # boolean and comma positioning
 
 00501  c/bool-pos-eol.cfg                         c/bool-pos.c
--- a/tests/config/c/align_trailing_do_cmt.cfg
+++ b/tests/config/c/align_trailing_do_cmt.cfg
@@ -1,0 +1,6 @@
+align_right_cmt_span = 10
+nl_brace_while       = remove
+nl_do_brace          = remove
+sp_do_brace_open     = add
+sp_brace_close_while = add
+sp_while_paren_open  = add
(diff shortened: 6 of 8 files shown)

### CODE REVIEW

**Comment by micheleCTDEAdmin** [comment]
Hi @fwojcik,
I looked at the patch and I share your view, a more general solution would be better.
Also removing the newline as in the example above raising the question: should the comment be on the same line or moved to the line after? As in:
```
       do {
                // this comment should be aligned with the one above
                r = baz();
        } while (r);
```
Note the comment is now not on the same line as the brace. It's a very subjective point, so happy to hear your opinion as well.

**Comment by fwojcik** [comment]
One of my upcoming PRs (not _very_ soon; soon-ish) is going to add this "pulling braces across comments when deleting newlines" functionality to `case` statements, and I would very much want the comment to stay on the `case` line for that. I'd rather have consistent behavior across these `nl_brace_* = remove` options, so my desire would be to keep the behavior of this patch.

**Comment by fwojcik** [comment]
OK, here is another take on the problem. From the commit: this change fixes the issue by taking a broader approach than a pinpoint fix for this specific case. First it makes `Chunk::orig_prev_sp` be calculated correctly, and second it makes `Chunk::orig_prev_sp` be used as the definitive source of "columns between this chunk and the previous one", instead of the difference between `current->orig_col` and `previous->orig_col_end`.

This patch has the following components:
- All of the unit tests are the same as the previous version of the patch.
- `src/tokenize.cpp` now computes `orig_prev_sp` correctly. Before this it was using `cpd.column`, which is has not been used on the input side of `uncrustify` in quite some time, if I'm reading `git`'s history correctly. This meant `orig_prev_sp` was always a round multiple of the tab size if they were used, for one thing, since `cpd.column` is always 0 at this point, as far as I can tell.
- `src/output.cpp`, `src/indent.cpp`, and `src/align_trailing_comments.cpp` now all use `orig_prev_sp` to determine spacing between a chunk and its predecessor, even if that chunk changes. This gets `orig_col` and `orig_col_end` closer to consistent usage.
    
So after this patch, `orig_col` and `orig_col_end` are mostly only used in the following ways which worry me only a little:
- logging
- being copied as chunks are duplicated
- being set artificially as new, synthetic chunks are inserted
- compared to or used to set the `column` field of the _same_ chunk
- used to set _minimum_ indents for subsequent chunks (mostly in `indent.cpp`)
- `Chunk::ComparePosition`, which is only used in one place for some preprocessor things, which probably don't get rearranged
- various things in the parsing stages, where is seems like chunks wouldn't be rearranged yet
    - This includes `combine.cpp`, `tokenize_cleanup.cpp`, and `EnumStructUnionParser.cpp`
- various places for comment processing
    - `output_comment_multi()` and `output_comment_multi_simple()` in `output.cpp` 
- inter-chunk spacing in `space.cpp`
    - ordinarily these uses would worry me a lot, but some experimentation seems to show that this code is already prepared for the cases where chunks could get reordered
    - I could be wrong, of course

and in the following ways which worry me a little more:
- `align_add.cpp` where `prev` is always a comment
- `indent_comment()` in `indent.cpp`
- `calc_comment_next_col_diff()` in `indent.cpp`, used for aligning doxygen comments
- `get_indent_first_continue()` in `indent.cpp`, used for the `indent_ignore_first_continue` option
- the `indent_preserve_sql` option functionality in `indent.cpp`

So it could be that there are still some places which might be appropriate to convert to use `orig_prev_sp`, but I would strongly advocate that that should be done in a future patch.

All that said, this approach turned out much nicer and cleaner than I'd hoped! I'm pretty happy with it. I'm good with either this version or the original. If you or the other devs would like some other approach, then I'm definitely going to need some guidance on what that might look like.

**Comment by micheleCTDEAdmin** [comment]
Thanks @fwojcik.
The approach description seems good. I will run a test on about 3200 test files to see if there is any regression.

**Comment by micheleCTDE** [comment]
I did a comparison test on about 3200 c/c++ files. Overall it works quite well, but there is a regression.
I attached an example with a minimum test file, config used for the test and example of output with current master and this PR.
The reqressions is that multiline comments are aligned even if further away than the specified threshold (3 spaces in my config), which sometimes gives awkward results.

Please work on a fix and feel free to use the attached files and add them to uncrustify as a test case.

**Review by micheleCTDEAdmin — CHANGES_REQUESTED** [review]
Need to fix the regression mentioned in comments.

**Comment by fwojcik** [comment]
OK! This took a fair bit of investigation and is kind of amazing, but I don't think this is a regression; I think this is another class of bug that was fixed since we agreed to go with a broader approach!

First consider `main()`. The original code is:
```cpp
int main()
{
    if(( ev1 == -2 && ev2 == -2 ) // unknown
        || ev1 > 0 // configured to have presentation
        || ( ev1 == -2 && ev2 > 0 )) { // not configured, has default presentation
```
and the code after being formatted (using master, for clarity) is:
```cpp
int main()
{
	if ((ev1 == -2 && ev2 == -2) || // unknown
	        ev1 > 0 || // configured to have presentation
	        (ev1 == -2 && ev2 > 0))      // not configured, has default presentation
	{
```

The key difference between them being that the boolean operators (`||`) were pulled up a line! So in the `debug.txt` from master, we see (lines broken for rendering sanity):
```
#   4>               BOOL|               NONE|     PARENT_NOT_SET[ 32/ 35/ 37/  0][1/2/0][   2 0000 0020][0-0]
                                ||
#   3>        COMMENT_CPP|        COMMENT_END|     PARENT_NOT_SET[ 35/ 35/ 45/  1][1/2/0][             0][0-0]
                                   // unknown
```
and the `// unknown` comment does not get turned into a trailing (`PCF_RIGHT_COMMENT`) comment because of the faulty test in `align_right_comments()` where the `prev` pointer is the `CT_BOOL` chunk from the previous line:
```
log_rule(align_right_comments:190): rule is 'align_right_cmt_gap'
NOT changing END comment on line 3 (35 <= 37 + 0)
```
So what the code in master sees as "the column that the comment is in" (`pc->orig_col`) is 35, and what it sees as "the column that the previous chunk ends at" (`prev->orig_col_end`) is 37. IOW, it sees the comment at starting two columns _before_ the previous chunk ends. This is exactly the bug being fixed!

With my change, the gap between the comment and the boolean operator remains at 1 (the same distance that originally existed between the comment and the closing paren), and so it does get considered as a trailing comment because `align_right_cmt_gap` is configured to be 0:
```
log_rule(align_right_comments:188): rule is 'align_right_cmt_gap'
Changing END comment on line 3 into a RIGHT-comment (1 >= 0)
SetResetFlags(775): 0000000000000000^0000000008000000=0000000008000000
   orig_line is 3, orig_col is 35, Text() is '// unknown', type is COMMENT_CPP, parent type is COMMENT_END,
   [CallStack: SetResetFlags:767, align_right_comments:178]
```

Just the same with all the other comments in `main()`, so they all get aligned.

As for the comments in `isColonImmune()`, it is kind-of the same story, except that in master only the comment in line 18 `/*&& prev2 != '*' && prev1 != '/'*/` was erroneously excluded and now it is included. Lines 16 and 17 are included as trailing comments in both branches.

So basically I think this is working!

micheleCTDEAdmin (approved):

micheleCTDEAdmin:
That was quick! Indeed if the boolean operators were at the end of the previous line in the original code (instead of being pulled up), the output file would have given the same awkward results. So you are correct in saying that this is not a regression, it just exposed another existing bug.
Good work!

PR #3795, opened 2022-08-12
Fix typo in docs for indent_ternary_operator

fwojcik:
This just more accurately describes what the '1' option does.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1537,7 +1537,7 @@ indent_using_block              = true     # true/false
 # How to indent the continuation of ternary operator.
 #
 # 0: Off (default)
-# 1: When the `if_false` is a continuation, indent it under `if_false`
+# 1: When the `if_false` is a continuation, indent it under the `if_true` branch
 # 2: When the `:` is a continuation, indent it under `?`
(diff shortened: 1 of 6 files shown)


micheleCTDE (changes requested):
Although changes in the comment would be propagated to other files when making a release, it is always better to keep files in sync if possible.
Whenever an option is modified/added (including comments), there are a bunch of files to update (for example etc/defaults.cfg). Suggest you look up for the existing places of a comment and then modify accordingly.

fwojcik:
Hmm. I cannot seem to find any documentation on this. Are the extra files in, say, commit d3768757d3f4cddb42d916a9ba75a785f0f6a972 a complete list? If so, I'll update this PR as well as #3797. 

You say that this would be done as part of a release. Is there any existing automation for this process? It seems like it really needs it.

gmaurel:
Have a look to:
https://github.com/uncrustify/uncrustify/pull/3773#issuecomment-1206001431
and use scripts/release_tool.py  update build/uncrustify

fwojcik:
Great, thanks!

micheleCTDEAdmin (approved):
Looks good, thanks!

micheleCTDE (approved):

PR #3797, opened 2022-08-12
Fix newline options for variable blocks

fwojcik:
This is what PR #3777 has evolved into.

These code changes make the options relating to newlines for variable definition blocks (referred to as var_def_blks from here on out) act much more as the documentation says they do. It also documents (in `options.h`) their behavior for a number of edge cases. The options covered are:
- `nl_func_var_def_blk`
- `nl_var_def_blk_start`
- `nl_var_def_blk_end`

Technically, `nl_var_def_blk_in` is also covered by this change, but I believe that its behavior is unmodified, except that it also has the more consistent var_def_blk demarcation logic.

I am very happy with these changes in general. The choices I made regarding the edge cases of how comments and/or preprocessor statements affect what is considered the bounds of a given var_def_blk are strange to me, but were made to reflect what most unit test cases seemed to expect/rely on. Importantly, those are now documented more clearly. If additional changes to those need to be made and/or options need to be added, then I think I need to leave that to someone else for the time being.

PR walkthrough
==============

My PRs so far have been small, single-commit fixes. This is the first one that I feel makes sense to use multiple commits. In particular, while the first commit could be a PR all on its own, basically all the following PRs depend on it so I felt that submitting them all together was important. I have several other planned PRs that will be like this, so I welcome any style suggestions for how to present these PRs to you all.

There are 8 commits for this PR. They are:

Skip over qualifiers to check for variable blocks
-------------------------------------------------

When detecting which chunks might be part of var_def_blks, the current code doesn't consider type qualifiers. This causes it to misclassify code as not being in a var_def_blk when it is part of one and also the inverse.

This change skips over qualifiers before presenting chunks to `is_var_def()`. You can see there are two unit tests where items in var_def_blks were previously missed. Also, here is a snippet from a `debug.txt` diff, showing that the chunks given to `is_var_def()` make more sense now:
```diff
-newline_def_blk(2123): pc->orig_line is 130, pc->orig_col is 5, type is QUALIFIER, Text() is 'private'
-newline_def_blk(2125): next->orig_line is 130, next->orig_col is 13, type is QUALIFIER, Text() is 'const'
-newline_def_blk(2130): 'typ==var' found: 'private const' at line 130
+newline_def_blk(2130): pc->orig_line is 130, pc->orig_col is 19, type is TYPE, Text() is 'uint'
+newline_def_blk(2132): next->orig_line is 130, next->orig_col is 24, type is WORD, Text() is 'DEFAULT_INITIAL_CAPACITY'
+newline_def_blk(2137): 'typ==var' found: 'uint DEFAULT_INITIAL_CAPACITY' at line 130
```

I'm just going to ignore the fact that two qualifiers in a row (e.g. `private const`) probably should make `is_var_def()` return `false`.  :-)
Since it's only ever called by `newline_def_blk` which now skips over qualifiers it's probably not urgent to fix.

Factor out test for function context
------------------------------------

NFC. This code is going to be reused, so it's just factored out into a function.

Make a new log level for var def blks
-------------------------------------

NFC except for logging. Given how many changes have been made in this area, and how many times people have found issues, I felt that being able to isolate newline handling in var_def_blks was worth its own logging tag.

Fixes nl handling for variable definition blocks
------------------------------------------------

This is the real meat of this PR.

The big change is that `newline_def_blk()` is no longer called multiple times over the same code (per main loop). Previously, `newlines_brace_pair()` would call it once per function, class, or struct, and then once at the end starting at `Chunk::GetHead()`, which would then re-process all those same functions etc. but this time with `fn_top` set to `false`. This was one major source of
surprising/incorrect behavior.

I altered it to only be called once at the end of `newlines_brace_pair()`, and to handle `fn_top` detection internally. I also renamed it to `newline_var_def_blk` for naming consistency.

First chunk handling was fixed to correctly handle var_def_blks at the top of the file (see below for details). The type-skipping code was unified. State tracking (`var_blk` and `first_var_blk`) was changed to reflect reality.

The preprocessor-related changes are made to mirror existing behavior, as mentioned near the beginning of this PR.

Existing unit tests configs were altered to reflect their tests' intentions as interpreted from their expected results. A few expected results were altered to match the fixed var_def_blk detection (near `[]` chunks in types). One was changed to reflect the corrected behaviors of the nl options.

Improves logging for variable definition blocks
-----------------------------------------------

NFC except for logging. This adds some logging statements, unifies others, and deduplicates some others. A rule log was moved to reflect my understanding of when those should be logged.

Makes nl_var_def_blk_end be ignored at top-of-fn
------------------------------------------------

The documented behavior of `nl_func_var_def_blk` said:
```
// The number of blank lines after a block of variable definitions at the top
// of a function body.
```
and `nl_var_def_blk_end` said:
```
// The number of empty newlines after a block of variable definitions
// not at the top of a function body.
```

This strongly implied to me (and others) that these were designed to be mutually-exclusive options. In reality, **both** options would apply to "a block of variable definitions at the top of a function body", and the number of newlines added would be the higher of the two options. (Further discussions of actual behavior are below.)

This change makes them act mutually-exclusively, per the previous docs.

Adds unit tests for nl_\*var_def_blk\*
------------------------------------

NFC. Does what it says on the tin. These tests intentionally do not yet cover option behavior in the face of preprocessor directives and/or comments.

Update options.h docs for nl_\*var_def_blk\*
--------------------------------------------------

NFC. The list of edge cases/exceptions that I know about for each option has been updated.

Old behaviors
=============

`nl_var_def_blk_start`
----------------------

This behaved mostly as documented. The exceptions were that newlines would not be added if a var_def_blk was just after an opening brace, and var_def_blks at the top of the file would have newlines inserted into the middle, such as in:

```cpp
unsigned a;

char b;
int bar(void);

unsigned c;
char d;
int foo(void) {
	char buf[1000] = {0};
	size_t len;
	float amount;
	double remainder;
	bar();

	uint64_t quux;
	uint64_t quack;
	for (int i = 0; i < 1000; i++) {
		uint64_t checksum;
		double remainder;
		buf[i] += len;

		uint32_t xyzzy;
		uint32_t quark;
	}

	float amount;
	double remainder;
	baz();

	uint32_t llama;
	uint32_t treehouse;
}

char * w;
uint8_t x;
int baz(void);

long y;
int z;
```

`nl_var_def_blk_end`
--------------------

This sort-of behaved as documented. Newlines would be added after _every_ var_def_blk in the file, except for those at the end of a brace block or the end of the file. The number of newlines specified was also treated as a minimum; if more newlines were
specified by something else, that would take precedence.

```cpp
unsigned a;
char b;

int bar(void);
unsigned c;
char d;

int foo(void) {
	char buf[1000] = {0};
	size_t len;
	float amount;
	double remainder;

	bar();
	uint64_t quux;
	uint64_t quack;

	for (int i = 0; i < 1000; i++) {
		uint64_t checksum;
		double remainder;

		buf[i] += len;
		uint32_t xyzzy;
		uint32_t quark;
	}
	float amount;
	double remainder;

	baz();
	uint32_t llama;
	uint32_t treehouse;
}
char * w;
uint8_t x;

int baz(void);
long y;
int z;
```

`nl_func_var_def_blk`
---------------------

This was just weird. Newlines would be added after _every_ var_def_blk that was _at the top brace-level_ of a function, class, or struct, except for those at the end of the brace block. The number of newlines specified was also treated as a minimum; if more newlines were specified by something else (such as `nl_var_def_blk_end`), that would take precedence.

```cpp
unsigned a;
char b;
int bar(void);
unsigned c;
char d;
int foo(void) {
	char buf[1000] = {0};
	size_t len;
	float amount;
	double remainder;

	bar();
	uint64_t quux;
	uint64_t quack;

	for (int i = 0; i < 1000; i++) {
		uint64_t checksum;
		double remainder;
		buf[i] += len;
		uint32_t xyzzy;
		uint32_t quark;
	}
	float amount;
	double remainder;

	baz();
	uint32_t llama;
	uint32_t treehouse;
}
char * w;
uint8_t x;
int baz(void);
long y;
int z;
```

Behavior changes
================

IMHO, the behavior changes in this PR are purely fixing bugs. However, it seems like these options' effects (especially `nl_func_var_def_blk`) could be somewhat surprising to existing users. For example, note that some unit tests had `nl_var_def_blk_end` options added to their configs (in the "Fixes nl handling for variable definition blocks" commit) because they expected those newlines to be added purely from setting `nl_func_var_def_blk`.

I am not sure if you would feel this is too big a change for users to accept as-is. If so, perhaps a new `nl_var_def_blk_func_top` variable (or similar) could be added, and then `nl_func_var_def_blk` could be marked as deprecated and internally set that value for both `nl_var_def_blk_func_top` and `nl_var_def_blk_end`?

Please let me know what you think. Thanks!


gmaurel:
... nl_var_def_blk_func_top variable (or similar) could be added, and then nl_func_var_def_blk could be marked as deprecated ...
is the better solution.

gmaurel (approved):
fine work, thanks

fwojcik:
OK, I followed the guidance to use a new option name and deprecate the old one.  The new option name is `nl_var_def_blk_end_func_top`, which was chosen based off the existing scheme for variables like `nl_func_call_start_multi_line` and `nl_func_def_end_single`.

The deprecation code enables `nl_var_def_blk_end_func_top` when ``nl_func_var_def_blk_end` is used. I considered having it enable `nl_var_def_blk_end` also, but that broke more things than it fixed.

I also updated all the things that `release_tool.py` updates for option changes.

micheleCTDEAdmin:
Looks good, tested on 3200 files and no regressions at all. Well done.

micheleCTDEAdmin:
> My PRs so far have been small, single-commit fixes. This is the first one that I feel makes sense to use multiple commits. In particular, while the first commit could be a PR all on its own, basically all the following PRs depend on it so I felt that submitting them all together was important. I have several other planned PRs that will be like this, so I welcome any style suggestions for how to present these PRs to you all.

Hi Frank, I usually follow the rule of thumb below.

1. If the commits relate to work on the same topic and are necessary to complete the development of a particular feature/fix, then presenting them together in the same PR makes perfect sense.
2. If the commits are implementing 2 or more features/bug fixes, they should be splitted into separate PR.

PR #3798, opened 2022-08-12
Fix parsing of a postfix incdec operator before a multiply

fwojcik:
Without this fix, `result = *var++ * int(0x1111);` will have the second star parsed as `CT_DEREF` instead of the correct `CT_ARITH`.

The only functional change here is the addition of `prev->Is(CT_INCDEC_AFTER)`. The remainder is just a reformatting of the existing code so it is much easier to read.

I looked through the other C++ postfix operators and didn't see any other easy-to-add-and-test cases.


gmaurel (approved):
looks fine

micheleCTDEAdmin:
Looks good.

Issue #3783, opened 2022-08-05
cli_options test breaks for me


fwojcik:
When checking out a clean copy of uncrustify master and building it, and then running `ctest`, the `cli_options` test fails for me. This seems to be related to the fix for the similar issue #3196. When I alter the test file as described in that issue (updated for more recent code) as follows:
```diff
diff --git a/tests/cli/test_cli_options.py b/tests/cli/test_cli_options.py
index 06b658bc..cec01b15 100755
--- a/tests/cli/test_cli_options.py
+++ b/tests/cli/test_cli_options.py
@@ -715,7 +715,7 @@ def main(args):
     #
     if os_name != 'nt':
         # &lt; &gt; doesn't work under windows
-        if parsed_args.config == 'Debug':
+        if True or parsed_args.config == 'Debug':
             eprint("Debug")
             if not check_uncrustify_output(
                     uncr_bin,
```
then the test passes for me.

While I did not do a full `git bisect`, checking out commit ac9b0f481159cfad5cfb21b841208c4bfbe09ed6 does break for me and the previous commit works. The behavior is identical whether or not I specify ` CMAKE_BUILD_TYPE=Release` on the cmake command-line.

I am using GCC 9.3 on Slackware 14.2. I'm happy to add any additional info about the failure.

Output of `ctest --output-on-failure   -R cli_options` from commit 90400185ca2b64bbc77486c2edb745d5c23ff11e:
```
Test project /home/fwojcik/progs/uncrustify/buildrel
    Start 13: cli_options
1/1 Test #13: cli_options ......................***Failed    0.15 sec

Uncrustify binary found: /home/fwojcik/progs/uncrustify/buildrel/uncrustify

************************************
Problem (1) with /home/fwojcik/progs/uncrustify/buildrel/tests/cli/results/tracking_space.html
************************************
u'  <html>'
u'  <head>'
u'     <meta http-equiv="content-type" content="text/html; charset=utf-8"/>'
u'     <title>Uncrustify: where do the Spaces options work</title>'
u'  </head>'
u'  <body lang="en-US">'
u'  <p>'
u'  </p>'
u'  <pre>'
u'- {<a title="REMOVE: 1"><font color="red">M</font></a>'
[...]

PR #3801, opened 2022-08-16
Adds a command to update options.h document contents

fwojcik:
This adds a command to `release_tool.py` which only updates the contents of the various option documents. It does not alter any version information in them, nor does it run any commands which contact remote git repos or alter the repo it runs in.

Since `universalindent.cfg` is another file that has to be updated when options change, I added it to the new `optiondocs` command only, since I wasn't sure if you'd want it during the normal update command.

This also updates `CONTRIBUTING.md` to show how and when to use it.

It also updates the regexes used to detect uncrustify version numbers to handle the edge cases that come from not having a named tag or other baseline in git.

To see an example of what this ends up looking like, see PR #3805. If this PR needs changes, then I'll update that PR to demo what they end up producing. After this PR is accepted, I'll promote that PR to a real one.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,10 @@ There are lots of ways to contribute to Uncrustify:
       Changes without tests should explain why tests are not present.
       (Changes that are non-functional, such as documentation changes,
       can usually omit tests without justification.)
+    - Any updates to `src/options.h`, including option descriptions, should be
+      accompanied by an update of the option documentation. This can be done by
+      building uncrustify and then running:
+      `./scripts/release_tool.py optiondocs path_to_uncrustify_binary`
 
 * Prepare a Pull Request (PR):
     - To reduce the likelihood of conflicts and test failures,
--- a/scripts/release_tool.py
+++ b/scripts/release_tool.py
@@ -74,13 +74,26 @@ def generate(repo, version, path, *args):
     import subprocess
 
     p = os.path.join(repo.working_tree_dir, path)
+
+    if version is None:
+        with open(p, 'r') as f:
+            content = f.read()
+            version = re.search(r'Uncrustify([a-z_]*)?-([0-9]+(\.[0-9]+)+(-[0-9]+)?)?(-?[0-9a-f]{6,})?', content)
+            if version:
+                version = version.group()
+            else:
+                version = 'Uncrustify'
+            version += "-dev"
+    else:
+        version = 'Uncrustify-{}'.format(version)
+
     with open(p, 'w') as f:
         c = subprocess.check_call(args, stdout=f)
     print('Created: {}'.format(path))
 
     alter(repo, path,
-          r'Uncrustify-[0-9.]+(-[0-9]+-[0-9a-f]+(-dirty)?)?',
-          r'Uncrustify-{}'.format(version))
+           r'Uncrustify([a-z_]*)?-([0-9]+(\.[0-9]+)+(-[0-9]+)?)?(-?[0-9a-f]{6,})?(-?dirty)?(-dev)?',
+           r'{}'.format(version))
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


micheleCTDE:
Hi Frank,
the number of options in README.md (and some other files) needs some review.
Either we don't update the number or we need to update the version number to something like '<next version number> [development]' or something like that.
Probably a good way could be that after an official release, the first thing to do should be to change the release number in that file to '<next version number> [development]', then update all the files as we develop and then update to final release number on the real release. What do you think?
Question extended to @gmaurel too.

On a side note, apologies for the delay in looking at your PRs. Busy period, will probably spend more time on them during the weekend.

fwojcik:
I understand where you're coming from. Not changing the count sort of makes sense, and does avoid some problems, but having it not match the actual count is not ideal. Doesn't seem like a showstopper at first glance, though. Having these scripts add some suffix like "-dev" to the version number might work also, if your concern is just having a different number of options alongside an existing release version string.

But I think the scenario that I'm most concerned about is where a number of different option additions, perhaps from different people, are queued against the same baseline. It would be very unfortunate if merging one of those caused merge conflicts in the rest. Do other aspects of this cause that issue? It seems like the whole option renumbering thing might. So I think maybe a wider net should be cast when trying to solve this problem.

fwojcik:
And no apologies for review timing are needed! I am not in any huge rush, and _completely_ understand that life can get in the way of projects. :)

micheleCTDEAdmin:
> I understand where you're coming from. Not changing the count sort of makes sense, and does avoid some problems, but having it not match the actual count is not ideal. Doesn't seem like a showstopper at first glance, though. Having these scripts add some suffix like "-dev" to the version number might work also, if your concern is just having a different number of options alongside an existing release version string.
> 
> But I think the scenario that I'm most concerned about is where a number of different option additions, perhaps from different people, are queued against the same baseline. It would be very unfortunate if merging one of those caused merge conflicts in the rest. Do other aspects of this cause that issue? It seems like the whole option renumbering thing might. So I think maybe a wider net should be cast when trying to solve this problem.

Yes, making things as "-dev" or "[development]" would be good. I think we can address that as a separate PR.
For this PR we can proceed as it is now. I will review later today or tomorrow.

fwojcik:
I think I may not have been clear enough; sorry about that. To elaborate: since we're considering changing the details of this process, I think it might be good to also reconsider handling of `etc/uigui_uncrustify.ini`. Looking at PR #3805, the diff for this file is huge due to the option renumbering. If I or someone else had another new option PR and 3805 were to be merged as-is, then I think that other new PR would suddenly have a big merge conflict on that file, since many options would be renumbered differently. I could be wrong about that, and GitHub somehow handles this case correctly. But this is a concern I have when thinking about how to handle option documentation for enhancement PRs.

micheleCTDEAdmin:
It's a valid point and concern. Numbering of the option was introduced by @gmaurel recently, if I remember correctly.
IMO, numbering makes sense on a final release, while during development we could probably do without it.
@gmaurel what do you think as well?

micheleCTDEAdmin:
Before merging this PR, let's come to an agreement on how to handle renumbering. I did a test and it seems it takes aways the numbering from `etc/uigui_uncrustify.ini`. As per previous comment, IMO this is perfectly fine during development.
We shall decide whether we want to keep numbering for releases or not and whether we want to go through the pattern 'release with number' --> 'development without numbers' --> 'release with number' --> 'development without numbers' --> .... 
My personal preference would be not to number options, keeps things simple.

guy-maurel:
Something might be wrong.
I have introduce the numbering of the options, to be able to find an option easier in the window created by the universalindent tool, and having a look to the configuration file.
Having a look to universalindentgui.cpp at line 61:
`#if defined (DEBUG) && !defined (WIN32)`
one may see that this is ONLY for debugging purposes.
It is never used with the "Release" command option.

Sorry! I don't understand the PR.

fwojcik:
That was an excellent lead, @guy-maurel. It turns out that I accidentally did have -DDEBUG enabled, which also explains my problems in issue #3783. I had it enabled because the default build type for uncrustify (which is none or "") has -DDEBUG enabled! The build instructions in README.md need to make this much clearer for Makefile users. I will make a PR for that.

As for this PR, @micheleCTDEAdmin, I will update the version string as discussed above and push it to this branch. This may take me a day or two; I have several pots boiling, as it were.

fwojcik:
This should be ready for review, mostly. I cannot update the demo output in PR #3805 until PR #3825 lands and I can rebase against it. Other than that, I think this is complete.

micheleCTDEAdmin (approved):
Tested, work great!

PR #3803, opened 2022-08-16
Adds C++ <cstdint> integer macros as types

fwojcik:
C++11 defines a number of macros for minimum-width integer constants. This adds those as types, so that uncrustify will "bind" them more tightly to their enclosed values than it would for a function, which is how they get interpreted without this change.

For example, without this, uncrustify would be happy to break `UINT64_C(0x0ab8e2a231b49732)` at the opening parenthesis, but with this change it will not do that.

https://en.cppreference.com/w/cpp/types/integer#Function_macros_for_minimum-width_integer_constants

Without this change, the unit test result looks like:
```cpp
#include <stdint.h>

uint64_t values[12] = {
	UINT64_C(1486918591), UINT64_C(28100), ~UINT64_C(4), UINT64_C(
		0x2700f184), UINT64_C(0x69b88dcc4d7e98eb), UINT64_C(
		0x0ab8e2a231b49732),
	UINT64_C(0x8226beae36e7317e), UINT64_C(0x5739289992c21658),
	UINT64_C(0x3acce043), ~UINT64_C(8), UINT64_C(4022), UINT64_C(
		63365009),
};
```


micheleCTDEAdmin (approved):
Looks good

PR #3804, opened 2022-08-16
Fix align_assign* when multiple definitions are encountered

fwojcik:
If assignment aligning is happening for a group of variable definition lines, and a line with multiple definitions is encountered, then align_assign.c will not align the most recent assignments before that line.

Without this fix, the unit test fails like:
```cpp
int foo(void) {
	unsigned a = 123;
	unsigned abcde = 456;
	unsigned abcdefghijklmnop = 789;
	unsigned w, x, y, z;
	unsigned u          = 16730;
	unsigned vvvvv      = 4917;
	unsigned wwwwwwwwww = 12428;
}
```


micheleCTDEAdmin (approved):
Nice one, this was another pending issue that is good to see fixed.

PR #3805, opened 2022-08-16
Adds sp_func_call_inside_fparen option

fwojcik:
This adds an option to add or remove space inside function _call_ '(' and ')' specifically.

This option overrides `sp_inside_fparen` for function calls only. This will allow for extra spaces in things like function definitons and ctor initializer lists, but not when just calling functions with data or variables.

While this _is_ in some sense a real pull request, and will someday be promoted to such, for right now I'll use this as a test/demo for PR #3801.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 A source code beautifier for C, C++, C#, Objective-C, D, Java, Pawn and Vala.
 
 ## Features
-* Highly configurable - 824 configurable options as of version 0.75.0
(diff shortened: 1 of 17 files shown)


gmaurel:
@fwojcik : could you please resolve the conflicts?
Thanks

fwojcik:
I'm abandoning this PR for now. I have several extensive changes that need to be rebased to be on more recent code, and this will be part of that, if it ever happens.

PR #3809, opened 2022-08-17
Check null after malloc() and define trivial destructor C++11

GermanAizek:
@micheleCTDEAdmin, @micheleCTDE

### DIFF
--- a/src/compat_win32.cpp
+++ b/src/compat_win32.cpp
@@ -32,14 +32,18 @@ bool unc_getenv(const char *name, std::string &str)
    if (buf)
    {
       len = GetEnvironmentVariableA(name, buf, len);
+      buf[len] = 0;
+
+      str = buf;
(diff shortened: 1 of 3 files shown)


src/compat_win32.cpp
```diff
@@ -32,14 +32,18 @@ bool unc_getenv(const char *name, std::string &str)
...
+      str = buf;
+      //printf("%s: name=%s len=%zu value=%s\n", __func__, name, len, str.c_str());
+      free(buf);
 
-   return(true);
+      return(true);
+   }
+   else
+   {
+      return(false);
```
micheleCTDEAdmin, line 45: we should make sure 'str' is an empty string in this case, since now it would not be updated if we go through the else part.

micheleCTDEAdmin (changes requested):
Please amend as commented.

micheleCTDEAdmin:
Also this PR is failing the automatic checks. Please fix them too.

guy-maurel:
replaced by #4341

PR #3810, opened 2022-08-17
Update C# and Vala keywords list

dbartolini:
Adds a couple of missing keywords from C# and Vala found during some debugging.


micheleCTDEAdmin (approved):
Looks good.

PR #3814, opened 2022-08-18
Fix incorrectly detecting enum declaration

PoeticPete:
It's valid to have `enum` prefix a type in parameter. In this case `enum` does not define a new enum. This PR adds a check to prevent the enum code from running in these cases.

input
```
int main()
{
  test2(
    [](enum TestEnum lhs) {
      return lhs.first < rhs;
    }
  );
}
```

```
int main()
{
  test2(
    [](enum TestEnum lhs) {
      return lhs.first < rhs;,
    }
  );
}
```


micheleCTDEAdmin (approved):
Looks good

Issue #3815, opened 2022-08-19
RFC: Add new cpp config to mirror indent_oc_block


PoeticPete:
Take this example

input
```
func(^ {
  return object;
});

func2([]() {
  return object;
});
```

output
```
func(^ {
  return object;
});

func2([]() {
    return object;
  });
```

config
```
indent_columns                  = 2
indent_with_tabs                = 0
indent_paren_open_brace         = true
indent_func_call_param          = true
indent_oc_block                 = true
```

The objc block indents because of `indent_oc_block = true`. Currently there is no such option for cpp lambdas. Would it make sense to add a config param, `indent_cpp_lambda_to_brace`, to mirror the objective-c option?

micheleCTDEAdmin:
Hi @PoeticPete, a new option could be good. I am not quite sure I understand how the indent would look like. Could you make a small example?

PoeticPete:
Basically it would look the same as the OC block example

input
```
func2([]() {
    return object;
  });
```

output
```
func2([]() {
  return object;
});
```

micheleCTDEAdmin:
This could fit nicely with the topic being discussed on PR #3816 with @dbartolini as well. Maybe we can first merge PR #3816 and then add the option.
@dbartolini what do you think about this idea?

dbartolini:
It makes sense, I will fix the regressions.

PR #3816, opened 2022-08-19
Fix indent_continue not indenting nested statements

### LINKED ISSUE #3752: indent_continue breaks nested continuation indents?

dbartolini:
Allows indent_continue to increment the indentation level in nested statements.
The LANG_OC check is to ensure previous behavior since this could break Objective-C code in strange ways I will not be able to fix properly at the moment.

This fixes all indenting problems in my C++ codebase without introducing regressions, especially in code inside lambdas, but feedback and broader testing is very much appreciated.

Fixes #3752.


micheleCTDEAdmin (changes requested):
Ciao Daniele,
I tested the PR over 3200 files and it does fix a lot of mis-indentation.
On the other hand I also spotted few regressions, with some code being indented too much. The regressions are similar in concept to the modified "expected" cpp files in this PR, where you can see extra indentation added.
If you could find a way to fix this PR so that the extra indentation is not present, it would be a great PR to merge.

dbartolini:
Assuming the regressions are similar to the one in `tests/expected/cpp/31593-sf593.cpp`: should `indent_continue` avoid incrementing the indentation when "nested statements are on the same line"?

The other modified "expected" .cpp files should not contain more indentation than necessary I think, could you confirm that?

micheleCTDEAdmin:
Yes, the regression I observed are similar to the one in `tests/expected/cpp/31593-sf593.cpp`. If you need, I can provide some examples.

Regarding the other modified tests, it could probably be a subjective thing. For example `tests/expected/cpp/60055-issue_3116.cpp`, the body of the lambda function is intended twice compared to the `B` function. While technically there are three levels of indentation (A -> B -> lambda body) and so it would explain the extra indentation, on the other hand it does look a bit awkward. What's your take on it?

dbartolini:
> Yes, the regression I observed are similar to the one in `tests/expected/cpp/31593-sf593.cpp`. If you need, I can provide some examples.

It would be helpful, thanks!

> Regarding the other modified tests, it could probably be a subjective thing. For example `tests/expected/cpp/60055-issue_3116.cpp`, the body of the lambda function is intended twice compared to the `B` function. While technically there are three levels of indentation (A -> B -> lambda body) and so it would explain the extra indentation, on the other hand it does look a bit awkward.

It does look awkward indeed, but it is consistent with how lambdas are indented normally. In this case, the `)` of B is on the same line with the `}` of its lambda parameter, and that is what makes it look weird, I think. Had the closing parentheses been on its own line, it would have looked nice and correct:

```
A(
    B([] (const std::string &s) -> bool {
            s = "hello";
            return true;
        }
    ), 1
);
```

Moreover, if you add extra parameters to B, with this PR you will get something like this, which I think it is correct:
```
A(
    B(foo,
        bar,
        [] (const std::string &s) -> bool {
            s = "hello";
            return true;
        }
    ), 1
);
```
Compare it with the current behavior; doesn't this look even weirder with those parameters squashed to B()'s level?
```
A(
    B(foo,
    bar,
    [] (const std::string &s) -> bool {
        s = "hello";
        return true;
    }
    ), 1
);
```
Like you said, it may be a subjective thing, and another option to adjust for personal preferences would be needed I guess.

micheleCTDE:
I have attached an example showing some regressions, including the config used to test it. Note that I use negative values for indent_continue.

Re lambda indentation, I see the point in your comment. Also there is an interesting idea proposed in issue #3815, so the solution may be to merge this PR (once the other regressions are fixed) and then add an option as proposed by @PoeticPete.

gmaurel:
@dbartolini 
Do you expected to continue your work? Let me know.
Thanks

dbartolini:
I’d love to see this PR merged, but I don’t have time to work on it right now.

PR #3817, opened 2022-08-20
Fix switch case indent after function call with dictionary

PoeticPete:
Seems like the extra square close condition was added by accident in https://github.com/uncrustify/uncrustify/commit/437aac5692470e543ab0c0821480bcc5bf68bad0 

Fixes #3182

input
```
- (int)test
{
  [obj method:[obj2 method2:@{}]];
  switch (var) {
    case one: return 3;
  }
  return 0;
}
```

```
- (int)test
{
  [obj method:[obj2 method2:@{}]];
  switch (var) {
      case one: return 3;
  }
  return 0;
}
```

config
```
indent_switch_case = 2
indent_with_tabs = 0
indent_columns = 2
```

command
```
./uncrustify -l OC+ -c tmp.cfg -f test_file_original.m -o test_file_original_unc.m -p myExample.p -L A 2>myExample.A -ds dump
```

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1517,9 +1517,7 @@ void indent_text()
                      && (  (  next->Is(CT_BRACE_CLOSE)
                            && next->GetParentType() == CT_OC_AT)
                         || (  next->Is(CT_SQUARE_CLOSE)
-                           && next->GetParentType() == CT_OC_AT)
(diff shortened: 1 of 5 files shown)


micheleCTDEAdmin (approved):
Looks good

Issue #3761, opened 2022-07-22
indent_switch_case broken inside lambda


dbartolini:
Version: `Uncrustify_d-0.75.1_f`

Command: `./build/uncrustify -c tests/config/common/empty.cfg -f indent-case.cpp -p indent-case.txt`

Debug file: [indent-case.txt](https://github.com/uncrustify/uncrustify/files/9169824/indent-case.txt)

Input:
```
namespace ns
{
f([]() {
		switch (x) {
		case 1:
			break;
		}
	});
}
```

Output:
```
namespace ns
{
f([]() {
		switch (x) {
			case 1:
				break;
		}
	});
}
```

Expected:
```
namespace ns
{
f([]() {
		switch (x) {
		case 1:
			break;
		}
	});
}

```

First bad commit: 7eb595caa38ced024a42ba1602d69e0387b8e3a1

dbartolini:
Maybe @alexhenrie could take a look?

PR #3818, opened 2022-08-21
Fix OC switch case indent when in OC block

### LINKED ISSUE #3813: Extra indent added in switch case in ObjC block
input
```
- (void)func {
  func2(^{
    switch (type) {
      case one:
        break;
      default:
        break;
    }
  });
}
```

```
- (void)func
{
  func2(^{
    switch (type) {
        case one:
          break;
        default:
          break;
    }
  });
}

```

config
```
indent_switch_case = 2
indent_with_tabs = 0
indent_columns = 2
```

command
```
./uncrustify -l OC+ -c tmp.cfg -f test_file_original.m -o test_file_original_unc.m -p myExample.p -L A 2>myExample.A -ds dump
```

PoeticPete:
Fixes #3813

input
```
- (void)func {
  func2(^{
    switch (type) {
      case one:
        break;
      default:
        break;
    }
  });
}
```

```
- (void)func
{
  func2(^{
    switch (type) {
        case one:
          break;
        default:
          break;
    }
  });
}

```

config
```
indent_switch_case = 2
indent_with_tabs = 0
indent_columns = 2
```

command
```
./uncrustify -l OC+ -c tmp.cfg -f test_file_original.m -o test_file_original_unc.m -p myExample.p -L A 2>myExample.A -ds dump
```

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -2040,6 +2040,11 @@ void indent_text()
                log_indent();
                indent_column_set(frm.prev().pc->column);
             }
+            // Issue #3813
+            else if (pc->TestFlags(PCF_OC_IN_BLOCK) && pc->GetParentType() == CT_SWITCH)
+            {
+               frm.top().indent = frm.prev().indent_tmp;
+            }
             else
             {
                // We are inside ({ ... }) -- indent one tab from the paren
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -3067,6 +3067,12 @@ static void handle_oc_block_literal(Chunk *pc)
    // mark the braces
    bbo->SetParentType(CT_OC_BLOCK_EXPR);
    bbc->SetParentType(CT_OC_BLOCK_EXPR);
+
+   // mark the OC_BLOCK
+   for (auto *tmp = bbo; tmp != bbc; tmp = tmp->GetNextNcNnl())
+   {
+      tmp->SetFlagBits(PCF_OC_IN_BLOCK);
+   }
 } // handle_oc_block_literal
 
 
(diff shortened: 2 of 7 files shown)


src/indent.cpp
```diff
@@ -2040,6 +2040,11 @@ void indent_text()
                log_indent();
                indent_column_set(frm.prev().pc->column);
             }
+            // Issue #3183
```
micheleCTDEAdmin, line 2043: This is issue #3813, not #3183

micheleCTDEAdmin (changes requested):
Comment needs to be amended.

Other than that, it seems a very custom-made solution. How about blocks that are not `switch`? Would probably be better to have a general solution I think. Maybe in that case we don't even need a new bit in the PCF list...

PoeticPete:
#3761 fixed the same issue in cpp and the solution was to check for `PCF_IN_LAMBDA` before running indent logic.

Re: adding additional flag -- If we have a `PCF_IN_LAMBDA` flag, it makes sense to have a flag for OC blocks too?

Re: making this more generalizable -- the issue is that switch cases are falling into the generalized `else` block [here](https://github.com/uncrustify/uncrustify/blame/6ebc9634a21a8f3e7fba713810bdc51514eb86e0/src/indent.cpp#L2006). Basically we are incorrectly assuming the switch block is `({ ... })`. It seems like the other conditions are looking for cases like `({` to determine indent logic, which isn't the case for switch statements.

micheleCTDEAdmin (approved):
Ok, makes sense. Thanks for the feedback and amending.

PR #3825, opened 2022-08-22
Fix version string in option docs

fwojcik:
Per PR #3801, this resets the version string in all the option-related documents to
be the last release string + "-dev". It also removes all the option numbering in
uigui_uncrustify.ini, since that is only supposed to be there for debug builds.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1,4 +1,4 @@
-# Uncrustify_d-0.75.0-245-6f360ab1
+# Uncrustify-0.75.1-dev
 
 #
 # General options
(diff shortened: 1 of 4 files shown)


documentation/htdocs/config.txt
```diff
@@ -1,4 +1,4 @@
-# Uncrustify_d-0.75.0-245-6f360ab1
+# Uncrustify-0.75.1-dev
```
micheleCTDEAdmin, line 1: IMO, this should be 0.76.0-dev, since it is the development stage of the next release (0.76.0).

micheleCTDEAdmin (changes requested):
Version number should be amended.

documentation/htdocs/config.txt
micheleCTDEAdmin: After having tested PR #3801, I now see why the numbers are 0.75.1-dev.
I think we can discuss about updating to a newer number in a separate PR :-)

micheleCTDEAdmin (approved):

PR #3826, opened 2022-08-24
primary-expression

### LINKED ISSUE #3786: Braces infinitely pushed by `align_struct_init_span` when using commas before closing brace
[//]: # 
[//]: # 
[//]: # 
[//]: # 
[//]: # "      or"

Current version: Uncrustify-0.75.1_f
[debug.txt](https://pastebin.com/CJSRh9yK)

Debug text is generated with the latest version I was able to use with my WSL. Mainly I use `0.75.1` on Windows. There is no difference between these two when it comes to this issue though.

Here's the original file:
```c
unsigned char wow[][2] = {
	{ 0, 24, },
	{ 152, 1, },
};
```
Running the formatter once gives pretty much the results I expect.
```c
unsigned char wow[][2] = {
	{ 0,   24, },
	{ 152, 1,  },
};
```
Every time it's ran again it keeps slightly pushin the braces away. Here's the results after 8 reformats.
```c
unsigned char wow[][2] = {
	{ 0,   24,                           },
	{ 152, 1,                            },
};
```
I noticed that this behaviour only happens because I have the commas inside the braces after the last numbers. For example this here wouldn't have the same issue:
```c
unsigned char wow[][2] = {
	{ 0,   24   },
	{ 152, 1    },
};
```

guy-maurel:
Fixes #3786
Consider the source:
```cpp
unsigned char wow[][2] = {
        {0,24,},
        {152,1,},
};
```
The second COMMA after the primary-expression 24 is not correct, but the compiler doesn't produce any error or warning.
The proposed patch detect that and doesn't shift the BRACE any more.


micheleCTDEAdmin:
I am confused: issue #3786 is now closed but this PR is not merged. So was #3786 a problem? or is this PR not needed?
I also tested on my 3200 files and see not a single difference with or without the PR.

rankaisija64:
From what I know it was a problem and this PR fixed it at least for me.

gmaurel (approved):

micheleCTDEAdmin:
> From what I know it was a problem and this PR fixed it at least for me.

ah I see. Issue #3786 was closed too early, before this PR was merged. Thanks for the explanation.

PR #3827, opened 2022-08-24
Replace 'nl_func_var_def_blk' with with 'nl_var_def_blk_end_func_top

dbartolini:
To avoid annoying warnings.


micheleCTDEAdmin (changes requested):
This is good. There are also other cfg files where 'nl_func_var_def_blk' is used, so it would be good to replace all of them.

dbartolini:
Updates with requested changes. I also renamed some config files that had 'nl_func_var_def_blk' in their name just to avoid confusion.

micheleCTDEAdmin (approved):
Looks good, thanks for amending it.

PR #3828, opened 2022-08-24
Fix indent_cpp_lambda_only_once documentation

dbartolini:
With ``uncrustify -c tests/config/common/empty.cfg -f test.cpp`` you will get:
```
auto x = [] {
                 return 0;
         };
```

Adding ``indent_cpp_lambda_only_once = true`` to ``empty.cfg``, instead:
```
auto x = [] {
        return 0;
};
```

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -3445,10 +3445,10 @@ use_indent_continue_only_once   = false    # true/false
 
 # The indentation can be:
 # - after the assignment, at the '[' character
-# - at the begin of the lambda body
+# - at the beginning of the lambda body
 #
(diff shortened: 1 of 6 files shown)


micheleCTDEAdmin (changes requested):
We have recently had a discussion about numbering the options. The feature was introduced by Guy for debug mode only, so for release mode the options won't be numbered. We have also recently switched the default build to Release instead of Debug.
Please amend `etc/uigui_uncrustify.ini` so that it does not contain the number of each option.

dbartolini:
Fixed.

micheleCTDEAdmin (approved):
Looks good, thanks for amending the PR.

PR #3829, opened 2022-08-26
Fixed #3658 lambda indentation regression with namespaces

### LINKED ISSUE #3658: indent_cpp_lambda_body problem with namespace
Config:
```
# Uncrustify-0.75
indent_cpp_lambda_body = true
```

Input:
```
namespace
{

void f1()
{
	const auto lambda =
		[]()
		{
			return 1;
		};
}

} // namespace

void f2()
{
	const auto lambda =
		[]()
		{
			return 1;
		};
}
```

Output (no changes were expected)
```
namespace
{

void f1()
{
	const auto lambda =
		[]()
	{
		return 1;
	};
}

} // namespace

void f2()
{
	const auto lambda =
		[]()
		{
			return 1;
		};
}
```

As you can see, the body of the lambda is being wrongly indented in f1(), just because it is enclosed in a namespace.  This regression appeared in version 0.75.

JodiTheTigger:
Removed namespace indent code introduced by my patch fix for issue #1813
The code has matured so that that fix is no longer needed, and actually
breaks things.

Secondly, due to misunderstanding of the indent_align_assign and
indent_continue settings, tests for issue #4328 were malformed. Updated
the configuration for those tests so that they are valid.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1700,44 +1700,6 @@ void indent_text()
                }
             }
 
-            // Issues: #1813, #3409, #3428
-            // if enclosure is set, the namespace is already accounted for
-            // so don't apply the correction twice.
-            // And for some reason, it's also accounted for when
-            // indent_namespace_single_indent is set
-            if (!enclosure && !options::indent_namespace_single_indent())
-            {
-               size_t namespace_indent_to_ignore = 0;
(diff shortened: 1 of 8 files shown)


micheleCTDEAdmin (approved):
Looks good, thanks.

PR #3833, opened 2022-08-30
Add new option to configure lambda indent in parameter

PoeticPete:


PoeticPete:
refactoring, updating docs, and doing more testing before submitting

PR #3839, opened 2022-09-04
Consider level when new lining OC messages

PoeticPete:
Fixes an issue where parameters inside messages were being incorrectly newlined. For example:

input
```
static void function() {
    [obj param1:nil param2:nil param3:[obj2 param1:nil param2:nil]];
}
```

output
```
static void function() {
	[obj param1:nil
	 param2:nil
	 param3:[obj2 param1:nil
	         param2:nil]];
}
```

expected
```
static void function() {
	[obj param1:nil
	 param2:nil
	 param3:[obj2 param1:nil param2:nil]];
}
```

config
```
nl_oc_msg_args                  = true
nl_oc_msg_args_min_params       = 3
```

### DIFF
--- a/src/newlines.cpp
+++ b/src/newlines.cpp
@@ -3497,7 +3497,8 @@ static void newline_oc_msg(Chunk *start)
          break;
       }
 
-      if (pc->Is(CT_OC_COLON))
+      if (  pc->Is(CT_OC_COLON)
+         && pc->level - 1 == start->level) // Only consider params on the current level
       {
          parameter_count++;
       }
@@ -3518,7 +3519,8 @@ static void newline_oc_msg(Chunk *start)
          break;
       }
 
-      if (pc->Is(CT_OC_MSG_NAME))
+      if (  pc->Is(CT_OC_MSG_NAME)
+         && pc->level - 1 == start->level) // Only newline messages on the current level
       {
          newline_add_before(pc);
       }
--- a/tests/expected/oc/50631-nl_oc_msg_args_min_params.m
+++ b/tests/expected/oc/50631-nl_oc_msg_args_min_params.m
@@ -13,4 +13,8 @@ static void function() {
 	[object param1:nil];
 
 	[object func];
+
+	[obj param1:nil
+	 param2:nil
+	 param3:[obj2 param1:nil param2:nil]];
 }
--- a/tests/input/oc/nl_oc_msg_args_min_params.m
+++ b/tests/input/oc/nl_oc_msg_args_min_params.m
@@ -8,4 +8,6 @@ static void function() {
     [object param1:nil];
 
     [object func];
+
+    [obj param1:nil param2:nil param3:[obj2 param1:nil param2:nil]];
 }


micheleCTDEAdmin:
Hi @PoeticPete,
I don't use ObjC but the proposed solution does not seem right to me because it ends up with different rules for objects on different levels.
IMO, the correct way to handle this would be for nested objects to re-evaluate the parameters `nl_oc_msg_args` and `nl_oc_msg_args_min_params` at their own level. So for example (using your settings):
```
static void function() {
    [obj param1:nil param2:nil param3:[obj2 param21:nil param22:nil]];
}
```
results in:
```
static void function() {
	[obj param1:nil
	 param2:nil
	 param3:[obj2 param21:nil param22:nil]];
}
```

while 

```
static void function() {
    [obj param1:nil param2:nil param3:[obj2 param21:nil param22:nil param23:nil]];
}
```
results in:
```
static void function() {
	[obj param1:nil
	 param2:nil
	 param3:[obj2 param21:nil
                 param22:nil
                 param23:nil]];
}
```

As further improvement on top of the above, an option to specify the max depth for applying this behavior would allow users to have more freedom on their preferred style.

micheleCTDEAdmin:
What do you think about it?

micheleCTDE:
I converted this to `Draft` till we clarify the point above, to prevent accidental merging.

PoeticPete:
That was my intended behavior originally because in many cases there are short functions / messages as params, which make the code less readable if we new line them.

For example, this is easily readable
```
[obj param1:nil
     param2:[obj2 param1:nil param2:nil]
     param3:nil
     param4:nil];
```

And this makes it less readable
```
[obj param1:nil
    param2:[obj2 param1:nil
                 param2:nil]
    param3:nil
    param4:nil];
```

Thoughts?

micheleCTDE:
I agree, in your example the top one is more readable. You could also have a scenario like:
```
[obj param1:nil
     param2:[obj2 param1:nil param2:nil param3:nil param4:nil param5:nil param6:nil]
     param3:nil];
```
where this would probably be more readable:
```
[obj param1:nil
     param2:[obj2 param1:nil
                  param2:nil
                  param3:nil
                  param4:nil
                  param5:nil
                  param6:nil]
     param3:nil];
```

Reflecting back on my comment from yesterday, if we evaluate things at each level, what would happen in the following scenario if we set the min mumber of arguments for new lines to 4?
```
[obj param1:nil param2:[obj2 param1:nil param2:nil param3:nil param4:nil param5:nil param6:nil] param3:nil];
```
I guess it would end up like this:
```
[obj param1:nil param2:[obj2 param1:nil
                             param2:nil
                             param3:nil
                             param4:nil
                             param5:nil
                             param6:nil] param3:nil];
```
which definitely looks very awkward to say the least.
One possible solution could be that if we do add new lines at a nested level, then we also add new lines at outer levels, which would result in something like this:
```
[obj param1:nil
     param2:[obj2 param1:nil
                  param2:nil
                  param3:nil
                  param4:nil
                  param5:nil
                  param6:nil]
     param3:nil];
```

Thoughts? Any other idea to handle such case?

PoeticPete:
Those are great points, and after looking at other code formatters such as Black for python, it might be best if we use line length to implement this rule. ie. any message containing a line over a given code width would be newlined.

This solves the issue raised above, where child messages are newlined, but the parent message is not.

Here's a PR the demos newlining based on code width: https://github.com/PoeticPete/uncrustify/pull/14/files

micheleCTDEAdmin:
I think the parameter "nl_oc_msg_args_min_params" is good to keep, it's a good one.
The max width "nl_oc_msg_args_max_code_width" can be an additional parameter, that works together with "nl_oc_msg_args_min_params" (a value of 0 would mean no width restriction.).
This gives more choice to the user. What do you think?

PoeticPete:
Is there a case where having both would be useful?

Purpose of these options is to make long function calls more readable, which makes me lean towards just having the code width option.

micheleCTDEAdmin:
> Is there a case where having both would be useful?

Working on open source project is about giving user choices. We all have different preferences, so I can easily foresee some users using only "nl_oc_msg_args_min_params", others using only "nl_oc_msg_args_max_code_width" and perhaps some others using both.
I don't use OC, so this doesn't really affect my own work. But I take a wider view for those users who do use OC :-)

micheleCTDEAdmin:
The work you already did on "nl_oc_msg_args_min_params" is good. It would be a pity to waste it. Instead we can complement it with the new "nl_oc_msg_args_max_code_width" param.

PoeticPete:
Sounds good. How would you expect these 2 options to work with one another? My first inclination is "OR"

ie. if there's a line with X or more params, OR there's a line over Y code width, newline all of the params

micheleCTDEAdmin:
Yes, the 'OR" is what I had in mind too. Just make sure there is a "don't use" value for each option, so a user can decide which option to use and which not. For example:
- "nl_oc_msg_args_max_code_width = 0" --> don't use this parameter
- "nl_oc_msg_args_min_params = -1 (or maybe 0)" --> don't use this parameter

PoeticPete:
Put up https://github.com/uncrustify/uncrustify/pull/3849 to add nl_oc_msg_args_max_code_width

micheleCTDEAdmin:
> Put up #3849 to add nl_oc_msg_args_max_code_width

So should we now close this one?

PoeticPete:
yep, we can close. This is still going to be a known issue when using `nl_oc_msg_args_min_params`, but will work with `nl_oc_msg_args_max_code_width`.

PR #3846, opened 2022-09-15
Add 'pp_multiline_define_body_indent' option to control indentation of multiline #define blocks.

micheleCTDE:
This allows to decide how to indent multiline #define blocks.
The settings allows to choose either:

 - a relative offset to the '#' chunk (>=0)
 - an absolute column (<0)

Indentation of the #define body then follow normal rules.
Several test cases shows the different behavior in action.


gmaurel (approved):
sound well

PR #3849, opened 2022-09-18
Add new option to newline OC messages based on code width

PoeticPete:
Often times we don't want to new line every single function call (OC message). For example, the following is readable as is

```
[obj param1:nil param2:nil];
```

It would be nice to provide the option to newline OC messages based on the line length. This PR adds the `nl_oc_msg_args_max_code_width` option to address this use case.

Related PRs:
#3839
#3739


documentation/htdocs/default.cfg
```diff
@@ -2029,6 +2029,9 @@ nl_oc_msg_args                  = false    # true/false
 # (OC) Minimum number of Objective-C message parameters before applying nl_oc_msg_args.
 nl_oc_msg_args_min_params       = 0        # unsigned number
 
+# (OC) Minimum number of Objective-C message parameters before applying nl_oc_msg_args.
```
micheleCTDEAdmin, line 2032: The comment in this file is not correct.

micheleCTDEAdmin (changes requested):
Minor correction required in a comment.

micheleCTDEAdmin (approved):
Looks good.

PR #3858, opened 2022-10-02
introduce the macro LOG_CURRENT_PC, to make the code simple.

guy-maurel:
created at debugging 3785.
It might be useful to have the same log of pc at some more places.


micheleCTDEAdmin (commented):
The macro is good, but I think a more appropriate name could be "LOG_CHUNK"? since the chunk to log is passed to the macro itself....
What do you think?

guy-maurel:
see better at #3859

PR #3859, opened 2022-10-07
introduce LOG_CURRENT_PC to unify the log

guy-maurel:
replace the PR #3858

### DIFF
--- a/src/align_eigen_comma_init.cpp
+++ b/src/align_eigen_comma_init.cpp
@@ -32,15 +32,7 @@ void align_eigen_comma_init()
 
    while (pc->IsNotNullChunk())
    {
-      if (pc->IsNewline())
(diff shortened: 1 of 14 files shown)


micheleCTDEAdmin (commented):
Same comment about the macro name that I made [here](https://github.com/uncrustify/uncrustify/pull/3858#pullrequestreview-1134622400).
Even better, we could replace the macro with a Chunk::Log() function, so we can simply write `pc->Log(LTOK);` for example.
What do you think?

guy-maurel:
No, the line number will be lost

micheleCTDEAdmin (approved):

micheleCTDEAdmin:
Good point about the line number which would be lost.

PR #3877, opened 2022-11-08
Re-Type C++ Access Modifiers in ObjC Message Send

briantkelley:
Tokens inside an Objective-C message send are parsed differently than in other contexts. Parameter names that are spelled the same as C++ access modifiers (e.g. `private:` or Qt's `signals:`) were previously formatted as access modifiers where the parameter name and colon were formatted on their own line with access modifier indentation (e.g. -2) and the argument value on the following line.

This change extends the re-typing logic in `handle_oc_message_send()` to also check for tokens typed as access modifier related.

### DIFF
--- a/src/combine.cpp
+++ b/src/combine.cpp
@@ -3505,6 +3505,7 @@ static void handle_oc_message_send(Chunk *os)
    }
 
    if (  tmp->Is(CT_WORD)
+      || tmp->Is(CT_ACCESS)
       || tmp->Is(CT_TYPE))
    {
       tmp->SetType(CT_OC_MSG_FUNC);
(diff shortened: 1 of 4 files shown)


src/combine.cpp
```diff
@@ -3517,11 +3518,12 @@ static void handle_oc_message_send(Chunk *os)
 
       if (tmp->level == cs->level + 1)
       {
-         if (tmp->Is(CT_COLON))
+         if (tmp->Is(CT_COLON) || tmp->Is(CT_ACCESS_COLON))
```
gmaurel, line 3521: Please use a newline before ||

gmaurel (changes requested):

src/combine.cpp
briantkelley: Thanks, PR updated!

micheleCTDEAdmin (approved):
Looks good

gmaurel (approved):

PR #3883, opened 2022-11-13
Vala: treat errordomain keyword as enum

dbartolini:
"The error domain declaration syntax is effectively the same as for enumerated types, but the keyword errordomain is used instead of enum." ([Vala Manual](https://www.vala-project.org/doc/manual.pdf), p. 98).


micheleCTDEAdmin (approved):
Looks good

PR #3884, opened 2022-11-14
Rearrange order of Chunk members to ease debugging

micheleCTDE:
@guy-maurel 
is the suggested order ok for you? If not, let me know your preference and I can adjust further.

I tried to rearrange based on your email request but also keep group of similar names together.


micheleCTDEAdmin (approved):
As per email from @guy-maurel, this is fine.

PR #3885, opened 2022-11-14
over sp_before_unnamed_byref

### LINKED ISSUE #3691: sp_before_unnamed_byref failed in std::function

guy-maurel:
fixes #3691
replaces PR #3693

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -1387,10 +1387,11 @@ static iarf_e do_space(Chunk *first, Chunk *second, int &min_sp)
 
       if (  next->IsNotNullChunk()
          && (  next->Is(CT_COMMA)
+            || next->Is(CT_PAREN_CLOSE)                            // Issue #3691
             || next->Is(CT_FPAREN_CLOSE)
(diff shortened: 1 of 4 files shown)


micheleCTDEAdmin (approved):
Looks good.

Issue #3910, opened 2022-12-14
sp_after_sparen fails react to changes in setting


SebastianBilek:
In my uncrustify configuration file I have defined:
```
sp_after_sparen = ignore
```
Which does: "add or remove space after ')' of control statements."
While my code looks something like this:
```
if (true)  printf("true");
if (false) printf("false");
```
When I run the uncrustify config on this code it turns into this:
```
if (true) printf("true");
if (false) printf("false");
```
What I have also found is that over-all any changes to the sp_after_sparen setting are not reflected in any way and all ways result in the previous conversion.

[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

PR #3913, opened 2022-12-15
Fix uncrustify#3910: sp_after_sparen doesn't reflect settings

SebastianBilek:
This is a response to the issue that I opened up here: [https://github.com/uncrustify/uncrustify/issues/3910](https://github.com/uncrustify/uncrustify/issues/3910)
Basically what I did was change the order around because some default operations were preventing the actual functions from running and I ended up fixing the way IARF_ADD events are handled in that now it doesn't have trouble with CT_VBRACE_OPEN tokens.
Also I had some test problems that I didn't consider before so now those should be fixed up.


SebastianBilek:
More problems...  :(

SebastianBilek:
fixed those problems :)

SebastianBilek:
Hi there, I had problems passing the test_cli_options.py test and sort of just assumed that --apply would help.
Could you please explain to me how I can compile with RELEASE?
Thanks, Seb.

tests/cli/output/universalindent.cfg
```diff
@@ -777,12 +777,12 @@ ValueDefault=ignore
...
 [Sp After Sparen]
 Category=1
-Description="<html>Add or remove space after ')' of control statements.</html>"
+Description="<html>Add or remove space after ')' of control statements.<br/><br/>Default: force</html>"
 Enabled=false
 EditorType=multiple
 Choices=sp_after_sparen=ignore|sp_after_sparen=add|sp_after_sparen=remove|sp_after_sparen=force|sp_after_sparen=not_defined
 ChoicesReadable="Ignore Sp After Sparen|Add Sp After Sparen|Remove Sp After Sparen|Force Sp After Sparen"
-ValueDefault=ignore
+ValueDefault=force
```
micheleCTDEAdmin, line 785: Default should be ignore like all the other options.

tests/cli/output/tracking_space.html
```diff
@@ -7,17 +7,17 @@
...
-}
-template<typename ...A, int... B>
-struct foo1 : foo1<A..., (sizeof...(A) + B)...>
-{
-   foo1() {
-      int x = sizeof...(A);
-      bool b = x > 1;
-   }
-};
+{<a title="REMOVE: 1"><font color="red">M</font></a>
```
micheleCTDEAdmin, line 10: This file should remain unchanged

micheleCTDEAdmin (changes requested):
Please check the provided feedback and adjust accordingly.

micheleCTDEAdmin:
> Hi there, I had problems passing the test_cli_options.py test and sort of just assumed that --apply would help. Could you please explain to me how I can compile with RELEASE? Thanks, Seb.

Hi @SebastianBilek,
the readme file explains how to build with release or debug :-)

SebastianBilek:
Hi @micheleCTDEAdmin, Sorry for that amateur question - I don't know how I missed that.
My reasoning for defaulting the sp_after_sparen setting to force is that before I applied my changes the sp_after_sparen setting it acted like it always was force - no matter the setting, so I wasn't sure whether you would be alright with me correcting all your tests to ignoring sp_after_sparen - which none of them do.

tests/cli/output/tracking_space.html
SebastianBilek: Yep this is done.

gmaurel:
the default value should be "ignore"

SebastianBilek:
So, has it been decided?
If so, may I do it automatically or should I manually modify all the tests?

micheleCTDEAdmin:
yes, both @guy-maurel and I think the default should be `ignore`. Please update the PR accordingly.

src/space.cpp
```diff
@@ -17,12 +17,11 @@
  * @license GPL v2+
  */
 
-#include "space.h"
-
 #include "add_space_table.h"
 #include "log_rules.h"
 #include "options_for_QT.h"
 #include "punctuators.h"
+#include "space.h"
```
gmaurel, line 24: Leave the include at the old place, before the block of many includes.

gmaurel (changes requested):
keep the old include place

src/space.cpp
SebastianBilek: Yep got it, sorry for that :sweat_smile:

micheleCTDEAdmin (approved):
Looks good to me. Still need @guy-maurel approval before merging.

gmaurel (approved):

micheleCTDEAdmin:
Thanks @SebastianBilek

PR #3923, opened 2023-01-16
Replace remaining chunk's nullptr logic with Chunk::NullChunkPtr logic

micheleCTDE:
As per title

PR #3926, opened 2023-01-17
Replace remaining chunk's nullptr logic with Chunk::NullChunkPtr logic

micheleCTDE:
As per title


micheleCTDE:
For some reasons tests on PR #3923 were not being executed. This PR was created as a replacement for it, but as soon as this was created, tests on PR #3923 also started (don't ask me why....).
So this PR is no longer needed. Closing.

PR #3930, opened 2023-01-24
First step in improving parsing frame code: renaming of the object names

micheleCTDE:
This commit is intended as a first step to improve the code related to the parsing frame stack. It renames some of the classes/enums to provide more meaningful names. More improvements will come in future commits.

@guy-maurel: before I merge, please let me know if you are ok with the new names.


micheleCTDE:
Ignore this PR, the names I chose were not good. I will create a new PR with better names to replace this PR.

PR #3944, opened 2023-02-11
some if(pointer) need to be changed to if(pointer->IsNotNullChunk())

### LINKED ISSUE #3943: will never been false
at indent.cpp line 1268
 if (next) will never been false, as next is a pointer to Chunk

guy-maurel:
fixes #3943


micheleCTDEAdmin (approved):

micheleCTDEAdmin (approved):
It looks good, thanks for catching those missing ones.

PR #3947, opened 2023-02-12
use a better comparision

### LINKED ISSUE #3946: if (stringType) is not sure
seen at tokenize.cpp line 1279/1318 should not use the value.
stringType is a cs_string_t variable of enum cs_string_t.

Preferred is.
```.cpp
if (stringType == CS_STRING_NONE)
...
if (stringType != CS_STRING_NONE)
```

guy-maurel:
fixes #3946


micheleCTDEAdmin (approved):
Looks good

PR #3963, opened 2023-03-18
Further improve parsing frame code

micheleCTDE:
Rework code and avoid `genDummy()` and `clear()` functions.


micheleCTDE:
Uhm... looks like online testing fails while on my computer I have no problem. Will need to further investigate why.

micheleCTDE:
@guy-maurel 
The PR reworks some functions related to the parsing stack. In particular it properly initializes a ParsingStackEntry rather than using dummy partially initialize ones.
One OC test is affected by this, but looking at the code and the result, IMO the original expected test result didn't seem quite right.
What do you think? If you are ok with it, I can then merge the PR.

gmaurel (approved):

PR #3971, opened 2023-03-28
the first parent muss get also the new flag

guy-maurel:


micheleCTDE:
@guy-maurel 
This commits introduces regressions.
You can use my usual test repo and config, compare the formatted output that you get when you build commit 5728ff5d and commit d7ed6153 (this one). There are 12 files which shows different behaviors.
Let me know if you need a specific issue report of this is enough.

PR #3972, opened 2023-03-30
& is CT_BYREF

### LINKED ISSUE #3967: sp_arith incorrectly adds spaces into types
```
[debug.txt](https://github.com/uncrustify/uncrustify/files/11039149/debug.txt)
```
Test file - should be named test.cc but github won't upload it as .cc

[test.txt](https://github.com/uncrustify/uncrustify/files/11039185/test.txt)
```
Results: output of diff... first line is original (and expected), second line is generated

1c1
< std::ostream& xx::operator<<(std::ostream& os, xx::Exception const& ex)
---
> std::ostream & xx::operator<<(std::ostream& os, xx::Exception const& ex)
```

Testing done with version Uncrustify_d-0.76.0_f
Issue found/exists also in version Uncrustify_d-0.74.0_f

Previously we were using a version of uncrustify from around 2007 - unknown version.

Fixing the output required setting "sp_arith = ignore". Obviously, we don't expect sp_arith to be affecting the return type definitions of functions.

guy-maurel:
In the line:
`  std::ostream& xx::operator<<(.....`

Fixes #3967


gmaurel (changes requested):
change the number 30147

PR #3975, opened 2023-04-03
Replace Chunk's SetType and SetParentType macros with real methods

micheleCTDE:
@guy-maurel 
As discussed by email, macros have been replaced with real methods. This allows to define "SetType()" methods on other classes without any clash.

Please check before I merge it.

### DIFF
--- a/documentation/htdocs/options_Align.html
+++ b/documentation/htdocs/options_Align.html
@@ -3,10 +3,6 @@
 align_pp_define_gap
 The minimum space between label and value of a preprocessor define
 
-#define SetType(tt)      do { \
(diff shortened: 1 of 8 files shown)


micheleCTDE:
Second try. Apparently `__func__` gives different results in linux (my box) and windows (CI box), causing the CLI option test "output92" to fail. Using `__PRETTY_FUNCTION__` to see if both environments are happy with it.

micheleCTDE:
Clearly "__PRETTY_FUNCTION__" didn't work in the CI environment. Will need to relook at this and find another solution.

gmaurel (approved):
This is fine

PR #3978, opened 2023-04-05
produces a software error

### LINKED ISSUE #2225: Unbalanced double quote inserting whitespace
Uncrustify-0.68
Configuration file:  default.

Input:

```
#include "Framework/UI/Window/CDlgComponentSubstitute.h>  // Compile error here...

// A lot of code here...

void Test()
{
    const char* const myString = "My string";
    // ...
}
```

Ouput:

```
#include "Framework/UI/Window/CDlgComponentSubstitute.h>

// A lot of code here...

void Test()
{
    const char* const myString = "My string ";     // Note that a whitespace was added!
    // ...
}
```

This is a tricky one, I'm not sure that something can be done about it!
Here's what happened.

- This is a small example. but the real file was much bigger, and the string was in the last function.
- The developer made a mistake.  He added a new include, but got confused and mixed " and <>.
- He ran Unscrutify.  He did not notice that the string was modified (added space) since it was at the end of the file.
- He then compiled the file. There was a compile error at line #1, of course.
- He corrected the error, so everything compiles now.
- He committed his changes.

Fortunately, our regression tests catched this one...

Ideally, Unscrustify shouldn't do anything since the " are not balanced...  but I'm not sure if this is possible!

guy-maurel:
if the include statement is not correct
fixes #2225

### DIFF
--- a/src/tokenize.cpp
+++ b/src/tokenize.cpp
@@ -2776,6 +2776,18 @@ void tokenize(const deque<int> &data, Chunk *ref)
       if (cpd.in_preproc != CT_NONE)
       {
          pc->SetFlagBits(PCF_IN_PREPROC);
+         // Issue #2225
+         LOG_FMT(LBCTRL, "%s(%d): orig line is %zu, orig col is %zu, type is %s, parentType is %s\n",
+                 __func__, __LINE__, pc->GetOrigLine(), pc->GetOrigCol(),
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin (approved):
Looks good. Also helps to catch errors in the original code :-)

Issue #3105, opened 2021-02-27
Inconsistent indentation of long lines


micheleCTDE:
uncrustify b51e2c22 + additional code for quick dumping (see this [branch](https://github.com/micheleCTDE/uncrustify/tree/dump-cycle) - rebase on top of master if necessary).

Problem: indentation of the same long line is different when a few extra characters are added.

The configuration uses
```
indent_continue = -6
```
to indent the continuation of a long line.
In the attached file, the output in test1 folder is correctly intended, while the output in test2 folder is wrongly indented. The difference in the source files is just 3 characters.
The output in test 1 (correct) is:
```
  connectFor(&m_timer, averylongfieldvariable, TQT_SIGNAL(timeoutVeryVeryLong()),
        TQT_SIGNAL(timeout()), this, TQT_SLOT(slotTimeout()));
```
and in test2 (wrong) is:
```
  connect(&m_timer, averylongfieldvariable, TQT_SIGNAL(timeoutVeryVeryLong()), TQT_SIGNAL(
          timeout()), this, TQT_SLOT(slotTimeout()));
```
Notice the different indentation of the continuation line.

To reproduce run: 
```
uncrustify -c test.cfg -f t1.cpp -o o1.cpp --dump -L 21,22,31,71,88,102 2>log1.log
uncrustify -c test.cfg -f t2.cpp -o o2.cpp --dump -L 21,22,31,71,88,102 2>log2.log
```

I will problably work on this myself in the near future. Will prepare a PR when ready.

[inconsistent_continuation.zip](https://github.com/uncrustify/uncrustify/files/6054655/inconsistent_continuation.zip)

micheleCTDE:
Additional info: the problem is caused during the second iteration of the final do-while loop. 
The first iteration splits the code correctly (see dump_4.log), then when calling indent_text() on the second iteration, the intentation is messed up.

micheleCTDE:
Hi Guy, the --dump option is a dirty-n-quick option I added to my local fork to get extra info for troubleshooting. It creates all those dump_X.log files you find in the attachment. You can find the code for such option [here](https://github.com/micheleCTDE/uncrustify/tree/dump-cycle). Since it is quick-n-dirty I didn't craete a PR, it would need some polishing and rework. If you think this could be useful, I can prepare something after April.

micheleCTDE:
For the record, I am working on a fix for this

micheleCTDE:
[...]

PR #3982, opened 2023-04-09
Fix indentation of continuation lines when indent_continue has negative values

micheleCTDE:
If indent_continue has negative values, it means the indentation is absolute from the beginning of the current frame. This relates to issue #3105.

Indentation is correctly maintained across multiple reruns on the same file, so if a long line was split during the first run, a second run (where the long line has already been split) will result in the same indentation.

No changes in behavior when indent_continue is positive or zero.


micheleCTDEAdmin (approved):

micheleCTDEAdmin:
@guy-maurel: as discussed by email and since you are away for a week, I have merged this PR without waiting for a further review from you. If you have any comments, we can discuss after you are back.

PR #3988, opened 2023-04-12
introduce more MS suffixes

### LINKED ISSUE #3980: Add support for integer suffix constants 8, 16, 32, 128 (64 is already supported)
[//]: # 
[//]: # 
[//]: # 
[//]: # 
[//]: # "      or"

Uncrustify supports the i64/I64 integer suffix constants, as detailed in:
https://learn.microsoft.com/en-us/cpp/c-language/c-integer-constants?view=msvc-170

But it only supports I64/i64, not the full range, as mentioned in
https://learn.microsoft.com/en-us/cpp/c-runtime-library/data-type-constants?view=msvc-170

It is missing suport for:
i8/I8, i16/I16, i32/I32, i128/I128

The existing function just needs expanded for the additional unsupported cases:
https://github.com/uncrustify/uncrustify/blob/master/src/tokenize.cpp#L1099

guy-maurel:
fixes #3980


micheleCTDEAdmin (commented):
There is a new input file Issue_3980.cpp but the new test refers to file Issue_3785.cpp.
Maybe a mistake?

gmaurel (approved):

PR #3990, opened 2023-04-12
Make sp_before_ptr_star_func independent of sp_before_unnamed_ptr_star

### LINKED ISSUE #3989: sp_before_unnamed_ptr_star overrides sp_unnamed_ptr_star_func=ignore
Config file:

```
sp_before_unnamed_ptr_star = remove
sp_before_ptr_star_func    = ignore
```

Input file:

```
int * foo(int *);
```

Expected output:

```
int * foo(int*);
```

Actual output:

```
int* foo(int*);
```

These two options should be completely independent of each other. If `sp_unnamed_ptr_star_func` is `ignore`, it's fine to fall back to `sp_before_ptr_star`, but it shouldn't fall back to its sibling `sp_before_unnamed_ptr_star` which is supposed to be for unnamed variables and typecasts.

alexhenrie:
And clarify that sp_before_ptr_star_func overrides sp_before_ptr_star.

Fixes #3989

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -232,7 +232,8 @@ sp_ptr_star_func_type           = ignore   # ignore/add/remove/force/not_defined
 sp_ptr_star_paren               = ignore   # ignore/add/remove/force/not_defined
 
 # Add or remove space before a pointer star '*', if followed by a function
-# prototype or function definition.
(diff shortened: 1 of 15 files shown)


micheleCTDEAdmin (approved):
Looks good and no regression on the big test

alexhenrie:
Thank you!

PR #3994, opened 2023-04-13
Add sp_qualifier_ptr_star options

### LINKED ISSUE #3993: Option to control space between 'const' and pointer star
In general I want to left-align the pointer star, for example:

```
char* const foo;
```

However, to conform to my company's style guide I need a space between the word `const` and a pointer star, for example:

```
char* const * foo;
```

There is currently no combination of Uncrustify options that can do both.

alexhenrie:
Fixes #3993

I personally could get by with only one new option, but it's necessary to add three options in order to preserve Uncrustify's flexibility.


alexhenrie:
Thank you!

PR #3999, opened 2023-04-17
Add options to align_nl_cont that take the existing spacing into account

### LINKED ISSUE #3998: Option to use existing final backslash alignment
I want to be able to align all final backslashes in a column, but do nothing if the author has already aligned them in a column farther to the right than the one that Uncrustify would pick. For example, I want Uncrustify to reformat the following:

```
#define SUM \
    one + \
    two + \
    three + \
    four \
```

But leave the following alone:

```
#define SUM
    one +                       \
    two +                       \
    three +                     \
    four                        \
```

alexhenrie:
Fixes #3998


micheleCTDEAdmin (changes requested):
Hi Alex,
I like the idea of being able to align the backslashes at a different column, but I think we need to rework this a bit.
1. It is better to keep `align_nl_cont` as it was. This allows to quickly enable/disable alignment and spare several file changes
2. We can add a new option that specifies how to align the slashes (`align_backslash_space`). This option would be active only if `align_nl_cont` is set to true. The new option would be numeric and specifies the number of spaces between the end of the longest line and the slash. The values could be:
a. -1: align the back slash to the right most back slash
b. 0 or positive: number of spaces to insert between end of longest line and back slash.
Default value could be 1.

I think this solution offers more flexibility and comes with less code/test changes. What do you think?

alexhenrie:
Your proposed `align_backslash_space` would work for me. In addition to `-1` for "align to rightmost backslash", we could have a `-2` for "align to leftmost backslash", although that's an extra feature that I don't personally need. However, one potential problem is that `align_backslash_space = -1` wouldn't provide a way to enforce a minimum number of spaces between the end of the longest line and the backslash. For example, the backslashes might be aligned already, but the user might want at least two spaces before each backslash instead of one.

It would also be nice to have a way to specify an absolute column for the backslashes. For example: Put all backslashes in column 80 if all lines are 78 characters long or less, otherwise, put all backslashes in the second column after the end of the longest line. I had thought about adding options such as `align_nl_cont_min_col` and `align_nl_cont_max_col` for this purpose. That would still work with your proposed `align_backslash_space`.

It seems to me that to do this properly, we would need `align_backslash_min_space`, `align_backslash_max_space`, `align_nl_cont_min_col`, and `align_nl_cont_max_col`, plus an option to try to align to the rightmost backslash, the leftmost backslash, or neither. My proposed patch adds the choice of left or right to the existing `align_nl_cont` option, which seems most natural to me, but making it a separate option would also be fine.

Like I said, your proposal would work fine for me personally, so I wouldn't mind just doing what you suggested. Still, given the above considerations, what do you think the best way forward is?

micheleCTDEAdmin:
Hi Alex,
will comment on this tomorrow. Today I have had a busy day and didn't have enough time to think about it properly.

micheleCTDEAdmin:
Let's summarize the various requirements that came up in the discussion.
1. being able to align to left most backslash, at its current column or right after longest line
2. being able to align to right most backslash, at its current column or right after longest line
3. being able to specify number of spaces between end of longest line and backslash column
4. being able to align backslashes at specific column. or right after longest line is specific column is too small

To achieve this we could use the following options:
1. `align_nl_cont`: enable/disable alignment. Option already exists and does not need changes
2. `align_nl_cont_mode`: values are `left`, `right`, `absolute`
3. `align_nl_cont_spaces`: numerical value. 
   a. If `align_nl_cont_mode` is `left` or `right`:
        - If `align_nl_cont_spaces == -1`, align to left or right most backslash on it current column.
        - If `align_nl_cont_spaces >= 0`, it specifies number of spaces between end of longest line and backslash.
   b. If `align_nl_cont_mode` is `absolute`, `align_nl_cont_spaces` value specifies absolute column for backslash. If longest line is longer than the spefified value, backslashes will be placed after the end of the longest line plus one space.

Did I miss any point? If not, what do you think about this proposal @alexhenrie ?

alexhenrie:
>     2. being able to align to right most backslash, at its current column or right after longest line

This can be simplified to "being able to align to right most backslash"

>        a. If `align_nl_cont_mode` is `left` or `right`:
>        - If `align_nl_cont_spaces == -1`, align to left or right most backslash on it current column.
>        - If `align_nl_cont_spaces >= 0`, it specifies number of spaces between end of longest line and backslash.

The value -1 is redundant: In the `left` and `right` modes, the current minimum or maximum column is always used if possible. Once the mode is specified, all that is needed is a way to specify the minimum number of spaces between the end of the longest line and its backslash, which can be any positive number including zero. Probably `code_width` should be taken into account too to determine the maximum number of spaces.

>        b. If `align_nl_cont_mode` is `absolute`, `align_nl_cont_spaces` value specifies absolute column for backslash. If longest line is longer than the spefified value, backslashes will be placed after the end of the longest line plus one space.

If `align_nl_cont_mode = absolute`, the length of the longest line plus one is `align_nl_cont_spaces` or longer, and `sp_before_nl_cont = remove`, I don't think there should be a space before the backslash on the longest line.

Otherwise, your proposal looks great! At the moment the only part that I need is `align_nl_cont_mode`. I'd be happy to rework my patch to add that option instead of extending `align_nl_cont`.

micheleCTDEAdmin:
Hi Alex, thanks for the additional comments. Here some extra feedback:

> This can be simplified to "being able to align to right most backslash"

Not necessarily, if the longest line is the last one :-)
```
#define foo(a,b) \
      do_something()    \
      a_very_long_function_name(a, b);
```
would become:
```
#define foo(a,b)                       \
      do_something()                   \
      a_very_long_function_name(a, b);
```
rather than:
```
#define foo(a,b)        \
      do_something()    \
      a_very_long_function_name(a, b);
```
<br/>

> The value -1 is redundant: In the left and right modes, the current minimum...

I was thinking about a case where backslashes are already there and we want to align to left or right most one, without forcing a minimum number of spaces. For example:
```
#define foo(a,b)                           \
      do_something()                              \
      a_very_long_function_name(a, b);
```
would become this (using 'left' and '-1)':
```
#define foo(a,b)                           \
      do_something()                       \
      a_very_long_function_name(a, b);
```
or this (using 'right' and '-1'):
```
#define foo(a,b)                                  \
      do_something()                              \
      a_very_long_function_name(a, b);
```
or this (using 'left' and for example '2'):
```
#define foo(a,b)                        \
      do_something()                    \
      a_very_long_function_name(a, b);
```
Note how using `-1` keeps the backslash on the existing left or right most column (as long as they are past the longest line) while a value `>=0` forces an exact number of spaces.

<br/>

> Probably code_width should be taken into account too to determine the maximum number of spaces.

Good point, otherwise it may make a mess. I didn't think about it.

<br/>

> If align_nl_cont_mode = absolute, the length of the longest line plus one is align_nl_cont_spaces or longer, and sp_before_nl_cont = remove, I don't think there should be a space before the backslash on the longest line.

Another good point that I had not consider. I think we could make `sp_before_nl_cont` obsolete with the new options, because `align_nl_cont_mode` and `align_nl_cont_spaces`  would determine where the backslash would be placed (not only in absolute mode but also in left and right mode). If someone wants to remove the space, he can use `left` and a value of `0` to get the same result. Will be good to add a comment about this too.

What do you think?

micheleCTDEAdmin:
Hi Alex,

> That's fine, just be aware that it's a change in behavior from what Uncrustify does now.

Thanks for pointing this out. Do you think we need an option to enab:e/disable this change of behavior, or is it something deem acceptable given that it is only happening if the last line is longer than all others and the backslashes?

> To implement my company's style guide (which I did not write BTW), I need to be able to use the existing alignment if possible, but also force a minimum number of spaces. So I would need `align_nl_cont_mode = left` combined with `align_nl_cont_spaces = 2` to leave the 5 spaces between the semicolon and the slash:
> [...]
> I think we could both be happy if in addition to `left`, `right`, and `absolute`, we add a `relative` mode to `align_nl_cont_mode` as well. When `align_nl_cont_mode = relative`, `align_nl_cont_spaces` would be the exact number of spaces between the end of the longest line and its backslash. Fair enough?

Happy to include your needs too, it may also be useful for other users :-) But I am confused by your proposed suggestion.
You wrote that in `relative` mode, `align_nl_cont_spaces` would be the exact number of spaces between the end of the longest line and its backslash. But this clashes with the need to "keep existing backslash alignment when possible but also ensure a minimum gap from longest line". Basically the explanation would be the same as `left` + positive space value.
If I understand correctly, you need something like `left` + *min* number of spaces. Instead of `relative`, I propose we add two new modes `left_min` and `right_min`. They are similar to `left` and `right`, but in this case the value of `align_nl_cont_spaces` (which must be positive for these two modes) specifies the minimum number of spaces to use after the longest line. 
Using the same example as before, using `left_min` and `2` we would get:
```
#define foo(a,b)                           \
      do_something()                       \
      a_very_long_function_name(a, b);
```
but using `left_min` and `10` we would get:
```
#define foo(a,b)                                \
      do_something()                            \
      a_very_long_function_name(a, b);
```
Note that in the second case the min value of 10 was enforced, while in the first case the current leftmost backslash place was used.
Is this what you need for your company style? 

> `sp_before_nl_cont` would still matter when not aligning the backslashes. 

Oh yes, good point. Then we need to keep the option but revise its functionality. `sp_before_nl_cont` would only affect spacing when backslash alignment is not used and we also need to add comments in the option explanation to let users know.

> Thanks for the careful thought!

Likewise, great discussion and team cooperation 👍

alexhenrie:
> Thanks for pointing this out. Do you think we need an option to enab:e/disable this change of behavior, or is it something deem acceptable given that it is only happening if the last line is longer than all others and the backslashes?

I don't care about it. Unless someone else does, I wouldn't worry about it.

> Happy to include your needs too, it may also be useful for other users :-) But I am confused by your proposed suggestion. You wrote that in `relative` mode, `align_nl_cont_spaces` would be the exact number of spaces between the end of the longest line and its backslash. But this clashes with the need to "keep existing backslash alignment when possible but also ensure a minimum gap from longest line". Basically the explanation would be the same as `left` + positive space value. If I understand correctly, you need something like `left` + _min_ number of spaces. Instead of `relative`, I propose we add two new modes `left_min` and `right_min`. They are similar to `left` and `right`, but in this case the value of `align_nl_cont_spaces` (which must be positive for these two modes) specifies the minimum number of spaces to use after the longest line.

Now it's my turn to be confused :-P Could you start by explaining what the difference would be between `align_nl_cont_mode = left` and `align_nl_cont_mode = right` when `align_nl_cont_spaces` is positive? It sounds like they would do the exact same thing.

> Using the same example as before, using `left_min` and `2` we would get:
>
> ```
> #define foo(a,b)                           \
>       do_something()                       \
>       a_very_long_function_name(a, b);
> ```
>
> but using `left_min` and `10` we would get:
>
> ```
> #define foo(a,b)                                \
>       do_something()                            \
>       a_very_long_function_name(a, b);
> ```
>
> Note that in the second case the min value of 10 was enforced, while in the first case the current leftmost backslash place was used. Is this what you need for your company style?

Yes, that's the end result that I need. However, the proposed design sounds convoluted: Why not rename `left_min` and `right_min` to simply `left` and `right`, and add a separate `relative` option for specifying an exact number of spaces instead of a minimum number of spaces? (This is related to the question I asked above.)

micheleCTDEAdmin:
> I don't care about it. Unless someone else does, I wouldn't worry about it.

Great, I also think a new option for this is not needed.

> Now it's my turn to be confused :-P Could you start by explaining what the difference would be between `align_nl_cont_mode = left` and `align_nl_cont_mode = right` when `align_nl_cont_spaces` is positive? It sounds like they would do the exact same thing.

You are correct. When `align_nl_cont_spaces` is positive, `align_nl_cont_mode = left` and `align_nl_cont_mode = right` do exactly the same thing. The difference is when `align_nl_cont_spaces == -1`, with the backslashes being aligned to either the left most or right most backslash (or after the longest line in case the last line is longer).

> Yes, that's the end result that I need. However, the proposed design sounds convoluted: Why not rename `left_min` and `right_min` to simply `left` and `right`, and add a separate `relative` option for specifying an exact number of spaces instead of a minimum number of spaces? (This is related to the question I asked above.)

ah ah, I spend 15 minutes typing my point of view, then I suddenly realized I was typing the same thing you also typed. Yes, I agree. So sum up the modes:
- `absolute`: absolute column (or after longest line if needed)
- `relative`: `align_nl_cont_spaces` (positive only) is the exact number of spaces to add from the end of longest line
- `left` and `right`: if positive, `align_nl_cont_spaces` is the minimum number of spaces between the end of longest line and the selected backslash position. If `-1`, use left or right most position of backslash.

alexhenrie:
> ah ah, I spend 15 minutes typing my point of view, then I suddenly realized I was typing the same thing you also typed. Yes, I agree. So sum up the modes:
> 
>     * `absolute`: absolute column (or after longest line if needed)
> 
>     * `relative`: `align_nl_cont_spaces` (positive only) is the exact number of spaces to add from the end of longest line
> 
>     * `left` and `right`: if positive, `align_nl_cont_spaces` is the minimum number of spaces between the end of longest line and the selected backslash position. If `-1`, use left or right most position of backslash.

That sounds great, but what would be the difference between `align_nl_cont_spaces = -1` and `align_nl_cont_spaces = 0`? It doesn't seem like -1 is necessary.

micheleCTDEAdmin:
> That sounds great, but what would be the difference between align_nl_cont_spaces = -1 and align_nl_cont_spaces = 0? It doesn't seem like -1 is necessary.

You are correct, using 0 would result in the same formatting (and again I spent 15 minutes typing an explanation with examples before I realized that 🙃 🙂 )

alexhenrie:
I started implementing this today, then realized that at some point, someone is going to want both `align_nl_cont_mode = absolute` and at least two spaces between the end of the longest line and its backslash, and the proposed design doesn't support that. What if we did something like this instead:

```
align_nl_cont_mode:
< 0: Absolute column number
  0: Relative
  1: Left
  2: Right
```

Then the user would be able to specify both an absolute column and a minimum number of spaces.

While we're on the subject, I still think it would be nice to extend `align_nl_cont` instead of adding a new `align_nl_cont_mode` variable. If we use 0 for "don't align" and 1 for "relative alignment" then the new options would map nicely onto the old options of `false` and `true`. But I will implement the feature in whatever way is necessary to get it accepted.

micheleCTDEAdmin:
Hi Alex,

> While we're on the subject, I still think it would be nice to extend `align_nl_cont` instead of adding a new `align_nl_cont_mode` variable. If we use 0 for "don't align" and 1 for "relative alignment" then the new options would map nicely onto the old options of `false` and `true`. But I will implement the feature in whatever way is necessary to get it accepted.

Ok, that sounds like a good suggestion. So perhaps something like this?
```
0 -> aligned disabled
1 -> relative
2 -> absolute
3 -> left
4 -> right
```

> I started implementing this today, then realized that at some point, someone is going to want both `align_nl_cont_mode = absolute` and at least two spaces between the end of the longest line and its backslash, and the proposed design doesn't support that. What if we did something like this instead:
> 
> ```
> align_nl_cont_mode:
> < 0: Absolute column number
>   0: Relative
>   1: Left
>   2: Right
> ```
> 
> Then the user would be able to specify both an absolute column and a minimum number of spaces.
> 

I think the idea of having absolute column + min space may definitely come up, so it is good to provide such possibility. But I think missing `mode` and `spaces` in the same `align_nl_cont_mode` option is not a very clean design. I would rather have an additional option to specify the absolute column (applicable only when the mode is `absolute`) and use `align_nl_cont_spaces ` to specify the minimum number of spaces for the other options (maybe even rename the option in such case).
What do you think?

alexhenrie:
If we create a separate `align_nl_cont_col` option, we won't need to have an "absolute" option in `align_nl_cont`. Setting `align_nl_cont_col` to a nonzero value would be sufficient to turn on absolute alignment mode. So `align_nl_cont` only has to have values for "disabled", "relative", "left", and "right". Is that all right with you?

micheleCTDEAdmin:
Sounds good. Let's put enough comments to clearly explain that when `align_nl_cont_col` is bigger than 0, the options `left` and `right` will be ignored, while settings for a minimum column gaps are still applicable.

gmaurel:
@alexhenrie , @micheleCTDEAdmin 
It is wonderful to follow the discussion!

alexhenrie:
I've pushed a new commit that implements the first part of our proposal: Adding left and right options to `align_nl_cont` as the numeric values 2 and 3. The new modes are useful on their own even without `align_nl_cont_spaces` and `align_nl_cont_col` being implemented yet.

micheleCTDEAdmin:
> I've pushed a new commit that implements the first part of our proposal: Adding left and right options to `align_nl_cont` as the numeric values 2 and 3. The new modes are useful on their own even without `align_nl_cont_spaces` and `align_nl_cont_col` being implemented yet.

Hi Alex,
how do you prefer to proceed from here? Merge this part and later work on the remaining functionality on separate PR? or complete the implementation and then merge? I am fine either way

alexhenrie:
I would like to merge this pull request now and add `align_nl_cont_spaces` and `align_nl_cont_col` later. Thanks!

micheleCTDEAdmin (approved):
Looks good, well done.

PR #4006, opened 2023-05-01
CI: use python3 interpreter

yegorich:
Modern Linux distributions install neither python2 nor symlink for python3. Hence, `run_ctest.py` will fail.


micheleCTDEAdmin (approved):
Looks good

PR #4014, opened 2023-05-12
Fixes for building with GCC13 and MinGW-w64

brechtsanders:
- Add `#include <cstdint>` in `src/indent.cpp` and `src/unc_text.cpp` to fix undefined types `uint32_t` and `int_fast8_t`.
- Fix format issue in `src/unc_tools.cpp` for MinGW-w64 where `size_t` is a different size than `long`

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -16,6 +16,7 @@
 #include "prototypes.h"
 #include "quick_align_again.h"
 #include "space.h"
+#include <cstdint>
 
 #ifdef WIN32
 #include <algorithm>                   // to get max
--- a/src/unc_text.cpp
+++ b/src/unc_text.cpp
@@ -13,6 +13,7 @@
 
 #include <algorithm>
 #include <stdexcept>
+#include <cstdint>
 
 
 using namespace std;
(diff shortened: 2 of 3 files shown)


brechtsanders (commented):
Changed order of
```
#include <cstdint>
#include <stdexcept>
```
in `src/unc_text.cpp` as recommended by CI

gmaurel (changes requested):
I prefer the change of the format:
```
      LOG_FMT(LDYNKW, "%s: %3zu: %18s, %14s, %12zu, %16s, %s\n",
      ...
```

gmaurel (changes requested):
It would be fine to have something such as:
```
# ifdef GCC13
  # include <cstdint>
...
```

brechtsanders:
Though it is possible to check the GCC version with an `#if` pragma, it should not be needed.

GCC 13 in fact fixes an issue where previously those types were available by accident even if `#include <cstdint>` was missing, so the correct solution `#include <cstdint>` even for other compiler versions. Form more info see: https://gcc.gnu.org/gcc-13/porting_to.html

Maybe I was wrong about `%zu`. The only `-Wformat` warning that is shown is:
```
src/unc_tools.cpp:516:50: warning: format '%ld' expects argument of type 'long int', but argument 7 has type 'size_t' {aka 'long long unsigned int'} [-Wformat=]
  516 |       LOG_FMT(LDYNKW, "%s: %3zu: %18s, %14s, %12ld, %16s, %s\n",
      |                                              ~~~~^
      |                                                  |
      |                                                  long int
      |                                              %12lld
......
  519 |               keyword_for_lang[ii].lang_flags,
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                    |
      |                                    size_t {aka long long unsigned int}
```

gmaurel (approved):

gmaurel:
Thanks for the discussion.

Issue #4026, opened 2023-05-29
the type is INCDEC_AFTER, is not coded. Please make a call.


unity-emilk:
Running a build from current master: 31190617093a8dc1fa553c8df7d0714f05b2dec5

Another issue with the config line `indent_member_single=true`

A file like: 
```
class foo {
public:
    int var;
 
    foo(int x) { var = x; }
};

int main() 
{
    int a = 2;
    a++;

    foo f(3);
    f.var++;
}
```

fails with:
```
ParsingFrame::pop (indent_text:1351): orig line is   14, orig col is   10, type is INCDEC_AFTER,
ParsingFrame::pop (indent_text:1351): the type is INCDEC_AFTER, is not coded. Please make a call.
```

Let me know if you prefer I submit PRs instead of issues.

[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

unity-emilk:
This seems to be all related to unhandled types in `parsing_frame.cpp`

I don't have test cases for all of them, but the following ones (and perhaps more) seem like they should be handled?

```
|| pc->GetType() == CT_OPERATOR
|| pc->GetType() == CT_CARET
[...]

PR #4028, opened 2023-05-31
Fix issue #4026

micheleCTDE:
@guy-maurel 
I have reworked the two big `if` blocks into one only. The difference was only the message that was printed (the first one having a `pushed with` added at the end). Having one big `if` makes it easy to see the CT_* in the test.
Let me know if it is ok for you, then I will merge


gmaurel (approved):
fine work, thanks

PR #4055, opened 2023-07-14
Add sp_between_ptr_ref configuration option to control whether or not a space is inserted between * and & in cpp.

JimmyManning:
This configuration option follows similar use case as sp_between_ptr_star in order to remove spaces between * and & when both sp_before_ptr_star and sp_before_byref are set to add or force.


gmaurel (approved):

micheleCTDE (changes requested):
The option is useful, but  a number of files needs to be updated when adding a new option, which has not been done here. You can either look at the usage of other options or use `uncrustify --update-config-with-doc` as indicated in the readme.
Please amend and then we can merge.

micheleCTDE:
Apologies, wrong command in previous comment (`uncrustify --update-config-with-doc`).
What I meant to say was to use `./scripts/release_tool.py optiondocs path_to_uncrustify_binary`. This was once in `contributing.md` but got lost in the preparation of 0.76 release.

JimmyManning:
It looks like not only the contributing.md changes were lost in preparation of 0.76, but also the optiondocs feature within the release_tool.py. Let me see if I can grab that version of release_tool.py and get the requested documents updated. Would you like another PR that brings these changes back, or an issue created?

JimmyManning:
I was able to run the older version of the script, however it's wanting to add documentation for configurations that I didn't add that might have been missed previously they are as follows:
sp_before_bit_colon
sp_after_bit_colon

Would it be best for these to be added in a separate PR?

Two other potential issues with the script are:
-  etc/uigui_uncrustify.ini file - it appears to contain lots of formatting changes that don't seem to be correct. 
- documentation/htdocs/ files are being modified to # Uncrustify_d-0.76.0-148-ce753d9e9-dev vs # Uncrustify_d-0.76.0-148-ce753d9e9

Thanks!

micheleCTDE:
Thanks for pointing out that the whole `optiondoc` has been removed.
@guy-maurel was this intentional? I think that option was quite useful to quickly update the required files.

> Would it be best for these to be added in a separate PR?
Let's wait for @guy-maurel's answer first, then we decide how to proceed.

> etc/uigui_uncrustify.ini file - it appears to contain lots of formatting changes that don't seem to be correct.
It may depends on the building version: release vs debug. One way it adds numbers to all the options and another it doesn't, so if you compare across two different build types, you would see differences all across the file

gmaurel:
Thanks for the message!

guy-maurel:
I cannot remember exactly.
I think it might not be necessary to update all those files by each PR,
only one time before tagging for a new version.

Other thought? Please comment!

guy-maurel:
This was introduce with f373c8dc851059432b80bdb59e2afcf50276dc61

micheleCTDE (approved):

micheleCTDEAdmin:
Let's merge this as is. I will look at reintroducing the `optiondoc` code and then updating those files which are missing some options. For example `default.cfg` is already missing at least a couple of options.

micheleCTDE:
`optiondoc` code reintroduce in PR #4108.

PR #4069, opened 2023-08-03
Minimal newlines after function body (fix #2787)

### LINKED ISSUE #2787: add newline after function, but don't remove multiple
I want this code:

```c
int foo()
{
    return 5;
}
int bar()
{
    return 3;
}


char test()
{
    return 'c';
}
```

to become:

```c
int foo()
{
    return 5;
}

int bar()
{
    return 3;
}


char test()
{
    return 'c';
}
```

But when I set `nl_after_func_body = 2`, two lines before `test()` becomes one. Is there other way?

Jorengarenar:


gmaurel (changes requested):
please provide some test data
see <uncrustify>/cpp.test

Jorengarenar:
> please provide some test data see /cpp.test

@gmaurel Let me know if 6d86c15a is enough :)

Jorengarenar:
Sorry for that, had to change branch name

PR #4070, opened 2023-08-05
Fix language detection for .inl files

dbartolini:
As per title.

### DIFF
--- a/src/language_names.cpp
+++ b/src/language_names.cpp
@@ -44,7 +44,7 @@ struct lang_ext_t language_exts[] =
    { ".hp",   "CPP"      },
    { ".hpp",  "CPP"      },
    { ".hxx",  "CPP"      },
-   { ".inl",  "PAWN"     },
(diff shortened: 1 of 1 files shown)


micheleCTDE (approved):
Looks good to me.

micheleCTDE:
@guy-maurel You initially set `.inl` files to the `pawn` language. Is that a possible use case too or just an oversight that this PR would correct? I can't find anything related to use of `.inl` files with `pawn` language

dbartolini:
> I can't find anything related to use of .inl files with pawn language

AFAICT Pawn uses the `.inc` extension for [include files](https://github.com/compuphase/pawn/blob/master/include/). I could add it to the `language_exts` array after this PR.

gmaurel:
I only created the file language_names.cpp, as an extract from uncrustify.cpp
Yes [dbartolini](https://github.com/dbartolini), please create a new PR

Issue #4071, opened 2023-08-09
Pointer star followed by qualifier is treated as unnamed


Jorengarenar:
Currently, if I have the following in config:
```
sp_before_ptr_star = add
sp_before_unnamed_ptr_star = remove
```

then code as the following:

```c
#include <stdio.h>

int *const *func(int);

int main()
{
	int *const *x;
	(int*) x;
	(const int*) x;
	(int const *) x;
	return 0;
}
```

gets "mis"-formatted to:

```c
#include <stdio.h>

int*const *func(int);

int main()
{
        int*const *x;
        (int*) x;
        (const int*) x;
        (int const *) x;
        return 0;
}
```

PR #4073, opened 2023-08-10
Option to add space before qualifier, 'operator', class or namespace name, '::'

### LINKED ISSUE #402: sp_before_unnamed_ptr_star=remove removes space before named ptr_star
When I add `sp_before_unnamed_ptr_star=remove` to the configuration file,

```
Adapter::Answer *Adapter::Answers::get()
{
}
```

is formatted as

```
Adapter::Answer*Adapter::Answers::get()
{
}
```

instead of remaining the same.

FWIW, the combination of ptr_star options I would like to use is:

```
sp_before_ptr_star=add
sp_before_ptr_star_func=add
sp_before_unnamed_ptr_star=remove
sp_between_ptr_star=remove
```

but I cannot have `sp_before_unnamed_ptr_star=remove` there because of this bug.

Tested with commit fc5228e.

Jorengarenar:
Fixes #402 and #4071 

New options:

* `sp_before_qualifier_ptr_star` - add or remove space before pointer star '*' that is followed by a qualifier.
* `sp_before_operator_ptr_star` - add or remove space before pointer star '*' that is followed by 'operator' keyword.
* `sp_before_scope_ptr_star` - add or remove space before pointer star '*' that is followed by a class scope (as in `int *MyClass::method()`) or namespace scope (as in `int *my_ns::func()`)
* `sp_before_global_scope_ptr_star` - add or remove space before pointer star '*' that is followed by '::', as in `int *::func()`.


tests/config/cpp/sp_before_operator_ptr_star-a.cfg
```diff
@@ -0,0 +1 @@
+sp_before_operator_ptr_star = ignore
```
micheleCTDEAdmin, line 1: Should the value be `add`?

tests/config/cpp/sp_before_operator_ptr_star-f.cfg
```diff
@@ -0,0 +1 @@
+sp_before_operator_ptr_star = ignore
```
micheleCTDEAdmin, line 1: Should the value be `force`?

tests/config/cpp/sp_before_operator_ptr_star-r.cfg
```diff
@@ -0,0 +1 @@
+sp_before_operator_ptr_star = ignore
```
micheleCTDEAdmin, line 1: Should the value be `remove`?

micheleCTDEAdmin (changes requested):
Three config files seems to have incorrect values (see individual comments). Please amend as necessary.
This also raises the question whether the expected results have been properly reviewed, because I would expect tests 60116 to 60119 to provide different output with the correct options and this wasn't picked up. Please do make sure the expected result is as intended.

Jorengarenar:
@micheleCTDEAdmin 
You are absolutely right. When I've split what was supposed to be `sp_before_keyword_ptr_star`, I only updated tests for `sp_before_qualifier_ptr_star` while forgetting about doing the same for `sp_before_operator_ptr_star`.

micheleCTDEAdmin:
The PR looks good, although when we add new options it would be good to update a bunch of other files too. I know those would get updated at release time eventually, but it is always a good idea to keep etc/defaults.cfg and documentation in line with the contents of options.h
@guy-maurel: ok to merge as is or better to have those files updated as well?

Jorengarenar:
@micheleCTDEAdmin 
Would there be anything beside `etc/defaults.cfg`? Because if that's all, I think it'd simpler if I just update it right away.

micheleCTDEAdmin:
> @micheleCTDEAdmin Would there be anything beside `etc/defaults.cfg`? Because if that's all, I think it'd simpler if I just update it right away.

Yes, there are. Look for example at PR 3842, you will see the new option `pp_indent_with_tabs` is added in a bunch of files.

Jorengarenar:
I've updated config files.  
I didn't bump up the count of options as it refers to release version.  
`etc/uigui_uncrustify.ini` seems to be already quite outdated and options are numbered(?), so I think it may require being generated again in general.

micheleCTDEAdmin:
> etc/uigui_uncrustify.ini seems to be already quite outdated and options are numbered

This depends on the build (release or debug) if I remember correctly.

micheleCTDEAdmin (approved):
Looks good to me.

@guy-maurel ok to merge the PR or you want to review first?

guy-maurel:
OK to merge

PR #4085, opened 2023-08-24
Vala: update keywords list

dbartolini:
Adds a few missing Vala keywords and removes the non-existent "callback" keyword which causes erratic formatting sometimes.


micheleCTDEAdmin (approved):
Looks good

PR #4089, opened 2023-08-28
Options to define minimum and maximum number of new lines after function body

### LINKED ISSUE #2787: add newline after function, but don't remove multiple

Jorengarenar:
Fixes #2787 

New options:

* `nl_min_after_func_body` - the minimum number of newlines after '}' of a multi-line function body
* `nl_max_after_func_body` - the maximum number of newlines after '}' of a multi-line function body

They aren't applied when `nl_after_func_body` is set.

If `nl_min_after_func_body` is bigger than `nl_max_after_func_body`, 
it essentially is doing the same as if `nl_after_func_body = nl_max_after_func_body`


micheleCTDEAdmin (changes requested):
Looks good. Just a small note on the new options' comments.
`nl_min_after_func_body` and `nl_max_after_func_body` only works if `nl_after_func_body` is 0. This is not clear without looking at that code and it would be better to rephrase it by adding something like "Only works when nl_after_func_body is 0" in the comment for the two new options.

Jorengarenar:
@micheleCTDEAdmin Updated 👍

micheleCTDEAdmin (approved):
Looks good, thanks

PR #4099, opened 2023-09-13
solve the if

### LINKED ISSUE #3238: Issue #2236 still present in version 0.73
See #2236  for reference.

guy-maurel:
we need to over spring the statement until the semicolon.
fixes #3236


tests/config/cpp/Issue_3236.cfg
```diff
@@ -0,0 +1,4 @@
+# Whether to fully parenthesize Boolean expressions in 'while' and 'if'
+# statement, as in 'if (a && b > c)' => 'if (a && (b > c))'.
+mod_full_paren_if_bool = true
+sp_bool                = force
```
bertwesarg, line 4: I still consider this a bug, that one needs to set this to `force`. The spaces are there, `sp_bool=ignore` should not remove them.

PR #4105, opened 2023-09-24
Suppress clang warning with -Wno-deprecated-declarations.

dundargoc:
This will prevent warnings such as:
"warning: 'sprintf' is deprecated: This function is provided for
compatibility reasons only.  Due to security concerns inherent in the
design of sprintf(3), it is highly recommended that you use snprintf(3)
instead. [-Wdeprecated-declarations]"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ elseif(CMAKE_COMPILER_IS_GNUCXX)
   endif()
   unset(gcc_warning_flags)
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin:
Personally I would keep the warning active and fix the usage of deprecated functions rather than simply masking the warning.
@guy-maurel what do you think?

dundargoc:
I don't mind fixing the warnings. Shall I use std::string instead since it's in general less error-prone?

micheleCTDEAdmin:
I am fine with either std::string or snprintf, whatever you are more comfortable with.

micheleCTDEAdmin:
Good to see you are still around btw 😃

gmaurel (approved):
fine

PR #4106, opened 2023-09-25
Add target "format" for easy formatting.

dundargoc:


micheleCTDE:
We already had the `format-sources` target for that, see [here](https://github.com/uncrustify/uncrustify/blob/master/CMakeLists.txt#L487). 
Looks like we now have a duplicated target, although with different names.

dundargoc:
Lmao, I missed that. I can make a follow up PR to remove it then.

micheleCTDEAdmin:
> Lmao, I missed that. I can make a follow up PR to remove it then.

It will probably be good to have only one target, so yes IMO we should remove the duplicated target.

PR #4108, opened 2023-09-26
Fix missing documentation for some options

micheleCTDE:
Some options were added without properly updating all required files. This PR fixes that.

@guy-maurel: will wait for you approval before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 A source code beautifier for C, C++, C#, Objective-C, D, Java, Pawn and Vala.
 
 ## Features
-* Highly configurable - 837 configurable options as of version 0.77.1
+* Highly configurable - 849 configurable options as of version 0.77.1
 - <details><summary>add/remove spaces</summary>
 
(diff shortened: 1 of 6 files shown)


gmaurel (commented):
etc/uigui_uncrustify.ini:
I think, the build of these file should be done with a release version to avoid the (numbers)
The (numbers) are more for debugging cases

micheleCTDEAdmin:
> etc/uigui_uncrustify.ini: I think, the build of these file should be done with a release version to avoid the (numbers) The (numbers) are more for debugging cases

I agree. The only reason I kept the numbers is that they were already present in the current master code. I also think we should not have the numbers there.
I will update the PR tomorrow and remove the numbers, then I will post the new version.

micheleCTDE:
PR updated removing the option numbering.

micheleCTDE:
@gmaurel let me know if ok now

gmaurel (approved):
Fine, Thanks

PR #4112, opened 2023-10-03
Make uncrustify respect the -q flag.

dundargoc:
The -q flag is supposed make uncrustify quiet, yet uncrustify still
prints "PASS: grid.c (32140 bytes)". This PR fixes that so uncrustify is
silent if the check passes, but not if it fails as it's likely the
wanted behavior.

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -117,8 +117,9 @@ static void uncrustify_start(const deque<int> &data);
  * @param dump_file    nullptr or the filename prefix for dumping formatting steps debug info
  * @param no_backup    don't create a backup, if filename_out == filename_in
  * @param keep_mtime   don't change the mtime (dangerous)
+ * @param is_quiet     whether output should be quiet
  */
-static void do_source_file(const char *filename_in, const char *filename_out, const char *parsed_file, const char *dump_file, bool no_backup, bool keep_mtime);
+static void do_source_file(const char *filename_in, const char *filename_out, const char *parsed_file, const char *dump_file, bool no_backup, bool keep_mtime, bool is_quiet);
 
 
 static void add_file_header();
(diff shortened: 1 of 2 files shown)


src/uncrustify.cpp
```diff
@@ -1507,8 +1509,11 @@ static bool bout_content_matches(const file_mem &fm, bool report_status)
    if (  is_same
       && report_status)
    {
-      fprintf(stdout, "PASS: %s (%u bytes)\n",
-              cpd.filename.c_str(), static_cast<int>(fm.raw.size()));
+      if (is_quiet)
```
micheleCTDEAdmin, line 1512: Shouldn't this be `!is_quiet`? when it is quiet, we should not print the message.

src/uncrustify.cpp
```diff
@@ -1024,12 +1026,12 @@ int main(int argc, char *argv[])
 
       // Issue #3427
       init_keywords_for_language();
-      uncrustify_file(fm, stdout, parsed_file, dump_file_name);
+      uncrustify_file(fm, stdout, parsed_file, dump_file_name, false);
```
dundargoc, line 1029: TODO: should this be is_quiet?

src/uncrustify.cpp
dundargoc: Hmmm, you're right. Unsure why, but the above gave the right behavior for some reason. I'll investigate, I have a suspicion what the problem is.

src/uncrustify.cpp
micheleCTDEAdmin: If the question is about default behavior, I think it should be "not quiet" since we have a command option to request to be quiet, but not viceversa.
If the question is about whether `false` is right or wrong, then I need to study the code further.

dundargoc: I meant it as a TODO note to myself.

src/uncrustify.cpp
dundargoc: OK, I think I've fixed it now.

dundargoc:
@micheleCTDEAdmin I think it works as expected now.

micheleCTDEAdmin (approved):
Looks good

PR #4116, opened 2023-10-07
Remove redundant function.

dundargoc:


src/uncrustify.cpp
```diff
@@ -1594,7 +1573,7 @@ static void do_source_file(const char *filename_in,
          if (strcmp(filename_in, filename_out) == 0)
          {
             // Create 'outfile.uncrustify'
-            filename_tmp = fix_filename(filename_out);
+            filename_tmp = string{ filename_out } + ".uncrustify";
```
micheleCTDE, line 1576: Forgive my ignorance, but why is the brace notation used? Seems a bit of a weird way to create a string in C++ to me.
`filename_tmp += ".uncrustify";` should do the job since just before the `if` we already have `filename_tmp = filename_out;`

dundargoc: Just brainfart. Fixed.

micheleCTDEAdmin (approved):
Looks good.

PR #4117, opened 2023-10-08
Fix flaws in line splitting logic

micheleCTDE:
The line splitting logic has several flaws, which sometimes results in awkward splits or (worse) in wrong splits, where the lines are longer than the allowed code width configured.
This PR fixes some of the issues. There is definitely more work to be done on line splitting, but this is a first step in the right direction.
Some of the expected test files are affected: some were wrong in first place, some are simply modified due to the new split logic implied.
Extra note: if `indent_continue` is negative, the values are considered absolute. In this case, the continuation lines have space for being quite long, therefore it is helpful to give more chances for splitting as late as possible. the code reflects this by adjusting the priority of open parenthesis in such case, which provides better split results on average.

@guy-maurel: please check and comment. I won't merge till I hear your feedback on this.


guy-maurel (changes requested):
Sorry, the patch needs to be build at new, because of
```.log
Checking patch tests/expected/c/02302-nl-cont.c...
error: while searching for:

#define FOO(bar) create_a_really_long_identifier name(some_function( \
                                                          bar1 + bar2), bar3, \
                                                      bar4);

#define VNV_RECORD_CYCLES(m) do { \

```

guy-maurel:
I download:
  https://github.com/uncrustify/uncrustify/commit/57fa941489536a4beb4e55428cc85628a9423ccc.diff
and try:
  git apply -v ~/Testen/Uncrustify/Issues/I-4117-michele/4117-B.diff
I get the error shown above.
I suspect "blank" and try:
   git apply -v --ignore-space-change ~/Testen/Uncrustify/Issues/I-4117-michele/4117-B.

It works.
Look for space in the first line, and some more...

micheleCTDE:
> I download: https://github.com/uncrustify/uncrustify/commit/57fa941489536a4beb4e55428cc85628a9423ccc.diff and try: git apply -v ~/Testen/Uncrustify/Issues/I-4117-michele/4117-B.diff I get the error shown above. I suspect "blank" and try: git apply -v --ignore-space-change ~/Testen/Uncrustify/Issues/I-4117-michele/4117-B.

Ah I see. I had similar issues too in the past sometimes, when using `git apply`. So I got into the habit to download a zip of the full branch from github when I want to test a patch, it avoids the problem.

micheleCTDEAdmin (approved):

PR #4140, opened 2023-11-02
Don't create uninstall target if used as subproject.

dundargoc:
This allows other projects to use uncrustify while still having their
own "uninstall" target.


micheleCTDEAdmin (approved):
Looks good, thanks

PR #4150, opened 2023-11-18
Set CMAKE_EXPORT_COMPILE_COMMANDS at the start of CMakeLists.txt

dundargoc:
It needs to be set before defining targets in order to work. Also make
it a normal variable instead of a cache variable as the variable cmake
uses is not a cache variable.


micheleCTDEAdmin (approved):
Looks good, thanks.

PR #4155, opened 2023-11-21
Fix VBRACE semicolon

pavelgeiger:
Fixes https://github.com/uncrustify/uncrustify/issues/3191 which was caused by https://github.com/uncrustify/uncrustify/commit/2748cfc041bc59c1de6389945dc0fb1b76c312e4. 
Mentioned commit causes that code like 

```
if (foo)
    return {};
```
is formatted to
```
if (foo)
    return {}

;
```

which is obviously not allright. Original commit fixed some D lang issues, so I have fixed it by wrapping its changes in condition so they are only applied for D lang.  When this is merged, I will be finally able to upgrade uncrustify after more than three years.

### DIFF
--- a/tests/expected/cpp/60164-Issue_3191.cpp
+++ b/tests/expected/cpp/60164-Issue_3191.cpp
@@ -1,0 +1,4 @@
+if (foo)
+	return {};
+
+int i = 0;
--- a/src/brace_cleanup.cpp
+++ b/src/brace_cleanup.cpp
@@ -513,7 +513,8 @@ static void parse_cleanup(BraceState &braceState, ParsingFrame &frm, Chunk *pc)
          frm.pop(__func__, __LINE__, pc);
          print_stack(LBCSPOP, "-Close  ", frm);
 
-         if (  frm.top().GetStage() == E_BraceStage::NONE
+         if (  language_is_set(LANG_D)
+            && frm.top().GetStage() == E_BraceStage::NONE
             && (  pc->Is(CT_VBRACE_CLOSE)
                || pc->Is(CT_BRACE_CLOSE)
                || pc->Is(CT_SEMICOLON))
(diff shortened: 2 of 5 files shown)


tests/expected/cpp/60164-Issue_3191.cpp
```diff
@@ -0,0 +1 @@
+if (foo) return {};
```
micheleCTDEAdmin, line 1: I don't understand. The test config is asking for a newline to be added after an `if` but there is newline here. This seems not correct. Any idea?

micheleCTDEAdmin: I understand that the original code has a bug, but the fix does not seem to be correct because the result does not behave as would be expected given the provided config.
I think we need a better solution, that fix the bug and behaves consistenly

micheleCTDE: `if (true) {...}<here>` yes, that is where it is supposed to add the newline. If `nl_after_if = add` uncrustify make sure there is at least one new line after the `if` block. Keep in mind that the end of the block is `}` and not `};`. Technically the `;` is a second statement.
The new test code is better, the original one was not showing any newline after the `if` statement.
Now the question is "should the newline be after `}` or after `};`? This can be a very subjective point of view. From one point of view, `};` looks better. From another point of view, having `}` followed by newline and a spare `;` would help identifying mistakes in a quicker way.
@pavelgeiger What do you think?
@guy-maurel would also like to hear your opinion.

gmaurel: I think newline be after } is the better option

micheleCTDEAdmin: I also think newline after `}` is most appropriate.
Looking at the original reported example, it seems two new lines are added rather than one, so that is something that should be address. The result should be something like this:
```
if (foo)
    return {}
;
```
which allows to clearly spot unnecessary semicolons. And the option `mod_remove_extra_semicolon` allows to remove them automatically if needed.

gmaurel (approved):

PR #4168, opened 2023-12-04
Check some options against nl_max

guy-maurel:
detect error in the configuration file


gmaurel (changes requested):
not yet

guy-maurel:
is replaced  by #4170

PR #4186, opened 2024-01-02
split newlines.cpp into many little files

guy-maurel:


micheleCTDE:
While the original newline file was too big, this proposed split seems too excessive to me. I think we can find a better split somewhere in between the original version and this version. For example the files named `newline_add_*.cpp` can go into a single `newline_add.cpp` file. Same for `newline_after*.cpp` and so on, basically grouping the files with some common name into a single bigger file.
What do you think?

micheleCTDE:
Sounds good. Will that be in an additional PR or a rework of this one?

gmaurel:
Yes, some more little merged PR will follow

gmaurel (approved):

PR #4190, opened 2024-01-05
merge the files with newline_add

guy-maurel:


gmaurel (approved):

micheleCTDE:
This way it looks much better. We just need to do the same for the rest. Keep going!

PR #4213, opened 2024-01-26
vala: handle missing optional expression after catch

dbartolini:
Vala and C# share optional expr after 'catch' keyword.


gmaurel (approved):
Fine, many thanks

Issue #4216, opened 2024-01-31
C++ scope problem with --tracking


JEAYNE:
Demo (with WIndows)
_(note: in the output I replaced the line numbers by xxxx because my source contains other stuff not related to this issue and this makes these line numbers misleading)_

```
> uncrustify.exe --version
Uncrustify-0.78.1_f

> echo float f1(int); | uncrustify.exe -c - -l cpp
main(xxxx): Parsing: 17 bytes (17 chars) from stdin as language CPP
float f1(int);

> echo float f1(int); | uncrustify.exe -c - -l cpp --tracking space:Space.html
main(xxxx): Parsing: 17 bytes (17 chars) from stdin as language CPP
uncrustify_file: Unable to create f1(int);
&Ý^ü☺: Invalid argument (22)
```

The root cause of this strange output ```Unable to create f1(int);``` and ```&Ý^ü☺: Invalid argument``` is a problem of scope in uncrustify.cpp

``` 
681: if (html_arg != nullptr)
682: {
          . . .
692:      char buffer[max_args_length];
693:      strcpy(buffer, html_arg);
694:
695:      // Tokenize and extract key and value
696:      const char *tracking_art = strtok(buffer, ":");
697:      const char *html_file    = strtok(nullptr, ":");
698:
699:      if (html_file != nullptr)
700:      {
             . . .
720:         cpd.html_file = html_file;  // <-- problem here
721:      }
722: } 
```

`html_file` points some where in `buffer`. But the scope of `buffer` is the block  `if (html_arg != nullptr){ ... }`
and the global variable `cpd.html_file` requires a global scope.

With this fix:
```
699: if (html_arg != nullptr && cpd.html_file == nullptr)

720:         cpd.html_file = strdup(html_file); 
```
the output looks good:
```
> echo float f1(int); | uncrustify.exe -c - -l cpp --tracking space:Spaces.html
main(xxxx): Parsing: 17 bytes (17 chars) from stdin as language CPP

> type Space.html
<html>
<head>
 ...
```

Note:s 
1. the memory allocated by strdup() is used until the end of the program. It is therefore not absolutely necessary to free() this memory.
2. the change made  line 699 is not mandatory because it is impossible to parse two --tracking, the second will be interpreted as a file name. But it may be clever to explicitly avoid memory leaks by showing that we will never call strdup() twice.

Regards

PR #4226, opened 2024-02-03
Fix global pointer pointing to local variable in uncrustify.cpp bugfix

JEAYNE:
BUG: In uncrustify.cpp the **global** pointer `cdp.html_file` points inside the **local** variable `buffer`.
This PR fixes this bug described in detail in #4216

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -696,7 +696,7 @@ int main(int argc, char *argv[])
       const char *tracking_art = strtok(buffer, ":");
       const char *html_file    = strtok(nullptr, ":");
 
-      if (html_file != nullptr)
+      if (html_file != nullptr && cpd.html_file == nullptr)
       {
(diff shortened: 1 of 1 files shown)


src/uncrustify.cpp
```diff
@@ -696,7 +696,7 @@ int main(int argc, char *argv[])
       const char *tracking_art = strtok(buffer, ":");
       const char *html_file    = strtok(nullptr, ":");
 
-      if (html_file != nullptr)
+      if (html_file != nullptr && cpd.html_file == nullptr)
```
gmaurel, line 699: Please, put && on the next line.

gmaurel (changes requested):

src/uncrustify.cpp
JEAYNE: Hi Guy,
Done.

gmaurel (approved):
it is fine, thanks

Issue #4217, opened 2024-02-01
Misplaced initialization in output.cpp


JEAYNE:
Hi,

Look at how the variable `first_text` is used in the function output_text() in the file output.cpp

```
1063: for (size_t track = 0; track < pc->GetTrackingData()->size(); track++)
         ...
1069:    bool              first_text = true;
         ...
1081:       if (first_text)
1082:       {
               ...
1086:          first_text = false;
1087:       }
         ...
1095: }
```

This is a classic pattern, but, because `first_text` is set to false at the begining of **each** loop, the test `if (first_text){ ... }` is always true and the assignment `first_text = false;` is useless.

I think that `first_text` shoud be initialized **before** the loop.

_Of course, the bug is only visible when the loop has to be executed more than once._

Regards

PR #4227, opened 2024-02-03
Fix initialization of first_text in output.cpp bugfix

JEAYNE:
BUG: In output.cpp the flag `first_text` is initialised **inside** the loop but, to distinguish the first text from subsequent ones, it should be **initialised** before the loop.
This PR fixes this bug described in detail in #4217


gmaurel (approved):
Thanks

PR #4229, opened 2024-02-05
Feat/reorganize files: alignment and newlines

micheleCTDE:
This is the first part of an effort to rearrange the source files in a more structured way.
There is no change in functionality, but instead of having 295 files in the `src` folder, many of them have been moved to the subfolders `align` and `newlines`. Further work on more files is planned once this PR is accepted.
Beside not having a huge number of files in a single folder, a second benefit is that most of those files can drop their prefix (like `align_` and `newlines`) since it is already captured in the subfolder name. This gives shorter and cleaner names, IMO.

@guy-maurel This is the first part of the effort. Let me know if you like the idea and if it is ok to merge the PR. If accepted, I will work on the second remaining part.


micheleCTDE:
Looks like building under windows failed due to a change in the name of the targets for the new names. I will fix that tomorrow, in the meantime you can still check the new file structure and let me know your opinio (@guy-maurel)

gmaurel (commented):
It is a good way.
But, have a look to this:
we have newline_xyz and newlines_abc
Do you want to have 1 directory or 2 directories for all the newline(s)?

micheleCTDE:
> we have newline_xyz and newlines_abc
>Do you want to have 1 directory or 2 directories for all the newline(s)?

I have combined them into a single folder. The idea is to group files into a folder by functionality, so to get something like:

- align: code for alignment
- newlines: code related to newlines logic
- indent: code for indentation and line splitting. Currently indent.cpp is a huge file, I expect we will split it at some point in future
- tokenizer: code to do the tokenizing process

I may or may not add one or more subfolders to the list above, I need to see how it looks like first.

gmaurel (approved):
a big work, fine and thanks

PR #4247, opened 2024-02-19
change enum to enum class

guy-maurel:
some more will follow


src/chunk_tag_t_keywords.h
```diff
@@ -0,0 +1,310 @@
+static chunk_tag_t keywords[] =
```
micheleCTDEAdmin, line 1: We should add a #if define guard to the file, to avoid multiple definition in case the file gets included twice in the same compile unit.

src/align/struct_initializers.cpp
```diff
@@ -23,7 +23,7 @@ void align_struct_initializers()
 
       if (  prev->Is(CT_ASSIGN)
          && (  pc->Is(CT_BRACE_OPEN)
-            || (  language_is_set(LANG_D)
+            || (  language_is_set(lang_flag_e::LANG_D)
```
micheleCTDEAdmin, line 26: Instead of the full `lang_flag_e::LANG_XXX` form, we could use `E_LANG_XXX` since they are define in `keywords.h`. It is quicker to read/write. Otherwise if we stick to the long form, it makes no much sense to have the  `E_LANG_XXX` one

src/chunk_tag_t_keywords.h
```diff
@@ -0,0 +1,310 @@
...
+{
+   // TODO: it might be useful if users could add their custom keywords to this list
```
micheleCTDEAdmin, line 3: As an idea, the user coulf specify similar entries in the config file. Those would have to be added to this table at run time. The table would have to be changed from a static array to a vector.

src/language_names.h
```diff
@@ -23,7 +23,7 @@ struct lang_name_t
 };
 
 //! list of all programming languages Uncrustify supports
-enum lang_flag_e
+enum class lang_flag_e
```
micheleCTDEAdmin, line 26: Suggest we use a name like `E_Language`.  The `E` at the beginning indicates it is an Enum and the rest of the name indicates the function. Also it would be a good idea to separate the `FLAG_XXX` into a separate `E_Language_Flag` enum. What do you think?

src/chunk_tag_t_keywords.h
guy-maurel: Yes

src/align/struct_initializers.cpp
guy-maurel: NO!
Using the short form produces:
```
src/align/struct_initializers.cpp:27:35: error: cannot convert ‘const size_t’ {aka ‘const long unsigned int’} to ‘lang_flag_e’
   27 |             || (  language_is_set(e_LANG_D)
      |                                   ^~~~~~~~
      |                                   |
      |                                   const size_t {aka const long unsigned int}
```

src/chunk_tag_t_keywords.h
guy-maurel: mmmh

gmaurel (approved):

PR #4257, opened 2024-03-07
suppress the loop

### LINKED ISSUE #4250: Infinite loop while parsing template with deprecation warning
[//]: # 
[//]: # 
[//]: # 
[//]: # 
[//]: # "      or"

First, I want to thank you for developing and maintaining this tool.  We use it heavily in https://github.com/ros2/ros2 .

I've found what looks to be an infinite loop problem in uncrustify with a particular piece of code in our codebase.  The original code is [here](https://github.com/ros2/rclcpp/blob/1981e1f12acfcbdc5d9202cd3e636cb692621971/rclcpp/include/rclcpp/any_subscription_callback.hpp), but the problematic function is:

```
class AnySubscriptionCallback
{
  template<typename SetT>
#if !defined(RCLCPP_AVOID_DEPRECATIONS_FOR_UNIT_TESTS)
  [[deprecated("use 'void(std::shared_ptr<const MessageT>)' instead")]]
#endif
  void
  set_deprecated(std::function<void(std::shared_ptr<SetT>)> callback)
  {
    callback_variant_ = callback;
  }
};
```

I'm currently running the following version of uncrustify (built from source today):

```
$ ./uncrustify --version
Uncrustify_d-0.78.1-138-e312b0e41
```

For this test, I'm using the [default configuration file](https://github.com/uncrustify/uncrustify/blob/e312b0e410dc41814b69c78e4b3d51b5311f7a49/etc/defaults.cfg) from this project.

When I run the above, like the following:

```
./uncrustify -c etc/defaults.cfg /path/to/code/above.cpp
```

Then uncrustify goes into an infinite loop.  If I remove the template argument, it doesn't loop forever.  If I remove the deprecation warning, it doesn't loop forever.  If I make it a standalone function (not inside the class), it doesn't loop forever.

Going back through `uncrustify` history, here are some working and failing versions:

* `uncrustify-0.72.0` - works
* `uncrustify-0.73` - works
* `uncrustify-0.74.0` - loops forever
* `uncrustify-0.75.1` - loops forever
* `uncrustify-0.78.1-138-e312b0e41` - loops forever

I can't attach a debug.txt (as produced by the `-p` option), since the infinite loop prevents that from being created.  Also, setting `debug_max_number_of_loops` seems to have no effect here, and it loops forever anyway.

I can get some debug information by running:
```
./uncrustify -c etc/defaults.cfg /path/to/code/above.cpp -L A 2>A.txt
```

You can find the results of that in https://gist.github.com/clalancette/5cf4a89ebf3d707d7eba82711c46efba .

Please let me know if you need any other information from me.

guy-maurel:
fixes #4250

### DIFF
--- a/src/chunk.cpp
+++ b/src/chunk.cpp
@@ -207,9 +207,14 @@ Chunk *Chunk::Search(const T_CheckFnPtr checkFn, const E_Scope scope,
    T_SearchFnPtr searchFnPtr = GetSearchFn(dir);
    Chunk         *pc         = const_cast<Chunk *>(this);
 
+   //LOG_FMT(LSETTYP, "\n%s(%d): origLine is %zu, origCol is %zu\n",
+   //        __func__, __LINE__, pc->GetOrigLine(), pc->GetOrigCol());
+
    do                                      // loop over the chunk list
    {
       pc = (pc->*searchFnPtr)(scope);      // in either direction while
+      //LOG_FMT(LSETTYP, "%s(%d): origLine is %zu, origCol is %zu, type is %s\n",
+      //        __func__, __LINE__, pc->GetOrigLine(), pc->GetOrigCol(), get_token_name(pc->GetType()));
    } while (  pc->IsNotNullChunk()         // the end of the list was not reached yet
            && ((pc->*checkFn)() != cond)); // and the demanded chunk was not found either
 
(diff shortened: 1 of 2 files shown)


clalancette:
> fixes #4250

I retested this with our problematic file (https://github.com/ros2/rclcpp/blob/1981e1f12acfcbdc5d9202cd3e636cb692621971/rclcpp/include/rclcpp/any_subscription_callback.hpp), and this does indeed fix the problem in my testing.  Thank you!

gmaurel (approved):

PR #4259, opened 2024-03-14
Fix out-of-bounds memory access in tokenize

adamjernst:
We compile uncrustify with ASAN, which uncovered that certain strange macros can trigger an out-of-bounds read here due to the use of `memcmp` without a bounds check on `pc->Text()`. Just use `strncmp` which gives us the semantics we want and won't read off the end of a string.


micheleCTDE (approved):
Looks good

Issue #4261, opened 2024-03-21
--tracking cannot be used with a file including a colon


JEAYNE:
Hi,
Here a bug preventing to use `--tracking <art>:<file>` with a file including a `:` (colon)
A colon is mainly used by Windows paths that include a drive.
Example  `--tracking space:C:\path\to\file.html`

The root cause of this bug is this short piece of code in the function main() in uncrustify.cpp
```
      // Tokenize and extract key and value
      const char *tracking_art = strtok(buffer, ":");
      const char *html_file    = strtok(nullptr, ":");    // <-- bug
```
The second call to strtok() captures the substring between the first `:` and the second `:`
As a result html_file points to "`C`".

The fix is very simple:
```
      // Tokenize and extract key and value
      const char *tracking_art = strtok(buffer, ":");
      const char *html_file    = strtok(nullptr, "\0");   // <-- fix
```
Now, the second call to strtok() captures the substring after the first `:` up to the end of the string.
As expected Html_file points to "`C:\path\to\file.html`"

Regards

micheleCTDEAdmin:
Solved by PR #4262

PR #4262, opened 2024-03-21
Fix a bug preventing to use --tracking with a path including a colon

JEAYNE:
The way uncrustify.cpp parses `--tracking <art>:<path>` prevents to use a path including a colon.
This PR fixes this bug described in issue #4261


micheleCTDEAdmin (approved):
Looks good

PR #4271, opened 2024-04-08
The option sp_balance_nested_parens never works with the value "false".

### LINKED ISSUE #4270: the option sp_balance_nested_parens doesn't work if the value is "false"

guy-maurel:
It will be replaced by the option _sp_paren_paren_
Fixes #4270


guy-maurel:
replaced by #4273

PR #4281, opened 2024-04-19
remove infinit loop

### LINKED ISSUE #4274: Uncrustify hangs until the system runs out of memory
Using Uncrustify-0.78.1_f, with with the default configuration on Windows 10:

`uncrustify.exe -c - file.cpp`

Use the following code example (the error here is that the user made a typo: there is a bogus `0` leading the method name).

```
int foo( const Object& object )
{
    return object.0Method();
}
```

The process hangs until the system runs out of memory.

guy-maurel:
fixes #4274

### DIFF
--- a/tests/c-sharp.test
+++ b/tests/c-sharp.test
@@ -122,7 +122,7 @@
 60015  cs/UNI-14131.cfg                             cs/UNI-14131.cs
 60016  cs/UNI-11662.cfg                             cs/UNI-11662.cs
 60018  common/UNI-18777.cfg                         cs/UNI-18777.cs
-60019  common/empty.cfg                             cs/UNI-18780.cs
+##60019  common/empty.cfg                             cs/UNI-18780.cs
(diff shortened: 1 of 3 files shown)


tests/c-sharp.test
```diff
@@ -122,7 +122,7 @@
 60015  cs/UNI-14131.cfg                             cs/UNI-14131.cs
 60016  cs/UNI-11662.cfg                             cs/UNI-11662.cs
 60018  common/UNI-18777.cfg                         cs/UNI-18777.cs
-60019  common/empty.cfg                             cs/UNI-18780.cs
+##60019  common/empty.cfg                             cs/UNI-18780.cs
```
micheleCTDEAdmin, line 125: @guy-maurel What is the reason to disable this test? If the code changes make the test fail, we should either adjust the code or the test, depending on what is applicable

gmaurel: Yes, look at #4282

gmaurel (approved):

PR #4294, opened 2024-05-07
Add support for new lining the rest of OC messages

PoeticPete:
Often times, we'll want every OC message (function) to be on a separate line. It can be tedious to do the line by line.

```
// Before
[obj param1:nil param2:nil param3:nil param4:nil param5:nil];

// After
[object param1:nil
        param2:nil
        param3:nil
        param4:nil
        param5:nil];
```

This change adds support for newlining the entire OC message if the message is already on multiple lines. This way we can just newline the 2nd parameter to newline the entire message.


guy-maurel (approved):
fine work, thanks for it.

gmaurel (approved):
thanks

PR #4300, opened 2024-05-23
Add C++20 support by updating strings

PoeticPete:
It would be nice to add support for the latest version of C++ and eventually upgrade. This PR adds support for C++20.

Currently the build fails when using C++20:

```
uncrustify/src/option.cpp:31:20: error: cannot initialize a variable of type 'const char *' with an lvalue of type 'const char8_t[1639]'
static const char *DOC_TEXT_END = u8R"___(
                   ^              ~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/uncrustify.dir/src/option.cpp.o] Error 1
make[1]: *** [CMakeFiles/uncrustify.dir/all] Error 2
make: *** [all] Error 2
```

The "u8" prefix of the literal will interpret strings as UTF-8. C++20 added a distinct type, char8_t, and along a distinct u8string type for this. However since there are only ansi characters within these literals, the "u8" prefix isn't necessary and we can keep using the same type.

Dropping the prefix is the simplest way to make the code compatible with C++20.

After this change, the building with C++20 works.

### DIFF
--- a/scripts/make_options.py
+++ b/scripts/make_options.py
@@ -49,7 +49,7 @@ class Option(object):
     def write_declaration(self, out):
         out.write(u'{} {} = {{\n'.format(self.decl, self.name))
         out.write(u'  "{}",\n'.format(self.name))
-        out.write(u'  u8R"__(\n{}\n)__"'.format(self.desc))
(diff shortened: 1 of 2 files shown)


micheleCTDEAdmin (approved):
Looks good

PR #4301, opened 2024-05-23
Add const to inline operators

PoeticPete:
Just add const to inline operator overloads to ensure the member variables (ie. `m_i`) can't be modified


micheleCTDEAdmin (approved):
Looks good

PR #4303, opened 2024-05-24
Remove unused insert_blank_lines

PoeticPete:
`newlines_insert_blank_lines` already exists in `blank_line.cpp`. The same function in `insert_blank_lines.cpp` isn't used anywhere, so this PR deletes it


micheleCTDEAdmin (approved):
Good catch on the duplicated code, it looks good.

Issue #4304, opened 2024-05-27
Objc msg colon incorrectly labeled as conditional colon


PoeticPete:
On current master https://github.com/uncrustify/uncrustify/commit/d8c58ecf68a4699e10e3bd664c767bb91a055399, if we try to format the following code:

```
static void func() {
    [NSObject doStuff:true ? 1 : 0
               param2:@"test"
               param3:@"test2"];
}
```
It incorrectly turns into:
```
static void func() {
    [NSObject doStuff:true ? 1 : 0
     param2 : @"test"
               param3:@"test2"];
}
```
Using the following cfg:
```
sp_cond_colon                   = force
align_oc_msg_colon_span         = 1
indent_columns                  = 4
indent_with_tabs                = 0
indent_ternary_operator         = 2
```

I will publish a fix soon

PR #4305, opened 2024-05-27
Fix objc msg colon mislabeled as conditional colon

PoeticPete:
Fix for  #4304

On current master https://github.com/uncrustify/uncrustify/commit/d8c58ecf68a4699e10e3bd664c767bb91a055399, if we try to format the following code:
```
static void func() {
    [NSObject doStuff:true ? 1 : 0
               param2:@"test"
               param3:@"test2"];
}
```
It incorrectly turns into:
```
static void func() {
    [NSObject doStuff:true ? 1 : 0
     param2 : @"test"
               param3:@"test2"];
}
```
Using the following cfg:
```
sp_cond_colon                   = force
align_oc_msg_colon_span         = 1
indent_ternary_operator         = 2
```

This is because the colon for the 2nd param is incorrectly labeled as a conditional colon


gmaurel (changes requested):
I missed some test(s)

PoeticPete:
Hi @gmaurel, what changes are you looking for?

gmaurel (approved):

PR #4318, opened 2024-06-05
C comment is too big

guy-maurel:


micheleCTDEAdmin:
Looks like indentation/alignment of text in the comment is a bit messed up. This may be due to tab size. Probably better to revisit before merging?

gmaurel (approved):

PR #4348, opened 2024-06-29
add cmake option FORCE_GNUINSTALLDIRS for MSVC

rkitover:
Add cmake option FORCE_GNUINSTALLDIRS to force using the UNIX GNUInstallDirs installation method under MSVC. This can be helpful for projects using uncrustify as an external project in their build systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -613,7 +613,10 @@ include(CPack)
 #
 # Install
 #
-if(MSVC)
+
+option(FORCE_GNUINSTALLDIRS "Force UNIX installation layout under MSVC" OFF)
+
+if(MSVC AND NOT FORCE_GNUINSTALLDIRS)
   install(TARGETS uncrustify DESTINATION ".")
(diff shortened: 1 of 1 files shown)


PoeticPete (approved):
Looks good to me. For context, it looks like this MSVC check was originally added for better windows support in https://github.com/uncrustify/uncrustify/commit/6244a8d1b23443d75d2847a150088839bf6abd02

PoeticPeteAdmin (approved):

dundargoc:
I don't wamna block this PR but I don't see a problem wirh always enabling gnuinstalldirs for all platforms. The one saving grace cmake has going for it is platform independence, and this goes against that by behaving differently for different platforms.

rkitover:
I disagree, I think this layout is more reasonable for installations on Windows. It doesn't matter that much if the layout under the program directory in program files uses GNUInstallDirs, but this layout is closer to how Windows programs are usually installed.

PR #4349, opened 2024-07-05
support for _Pragma

### LINKED ISSUE #3386: Parse error on trival, valid C source code
$ cat p.c
int f (void)
{
    _Pragma("P2");}

$ gcc -c p.c
$ cat p.cfg 
newlines                        = auto
$ uncrustify -c p.cfg p.c
Output suffix: .uncrustify
Parsing: p.c as language C
indent_text(4060): size is 2
indent_text(4062): File: p.c, open_line is 2, parent is NONE: Unmatched BRACE_OPEN
indent_text(4060): size is 2
indent_text(4062): File: p.c, open_line is 2, parent is NONE: Unmatched BRACE_OPEN
1|$ uncrustify --version
Uncrustify_d-86eb8205
$

guy-maurel:
fixes #3386

### DIFF
--- a/src/chunk.h
+++ b/src/chunk.h
@@ -1713,6 +1713,7 @@ inline bool Chunk::IsParenOpen() const
 {
    return(  Is(CT_PAREN_OPEN)
          || Is(CT_FPAREN_OPEN)
+         || Is(CT_GPAREN_OPEN)
          || Is(CT_LPAREN_OPEN)
          || Is(CT_PPAREN_OPEN)
          || Is(CT_RPAREN_OPEN)
@@ -1725,6 +1726,7 @@ inline bool Chunk::IsParenClose() const
 {
    return(  Is(CT_PAREN_CLOSE)
          || Is(CT_FPAREN_CLOSE)
+         || Is(CT_GPAREN_CLOSE)
          || Is(CT_LPAREN_CLOSE)
          || Is(CT_PPAREN_CLOSE)
          || Is(CT_RPAREN_CLOSE)
--- a/src/token_enum.h
+++ b/src/token_enum.h
@@ -330,7 +330,7 @@ enum E_Token
    CT_PP_UNDEF,
    CT_PP_PROPERTY,
 
-   // the next two PP types MUSST be BEFORE CT_PP_BODYCHUNK or AFTER CT_PP_OTHER
+   // the next two PP types MUST be BEFORE CT_PP_BODYCHUNK or AFTER CT_PP_OTHER
    CT_PP_F_PRAGMA,                                                                         // Issue #3386
    // Removes the L prefix (if any), the outer quotes, and leading/trailing whitespace
    // from string-literal, replaces each \" with " and each \\ with \, then tokenizes
--- a/src/token_enum.h
+++ b/src/token_enum.h
@@ -7,7 +7,6 @@
  *          October 2015
  * @license GPL v2+
  */
-
 #ifndef TOKEN_ENUM_H_INCLUDED
 #define TOKEN_ENUM_H_INCLUDED
 
@@ -21,6 +20,10 @@
  * This is an enum of all the different chunks/tokens/elements that the
  * program can work with. The parser and scanner assigns one of these to
  * each chunk/token.
+ */
+/*
+ * a function to get the text of a token
+ * const char *get_token_name(E_Token token)
  */
 enum E_Token
 {
@@ -215,6 +218,9 @@ enum E_Token
    CT_FPAREN_OPEN,      // 'function' parenthetis after fcn/macro fcn
    CT_FPAREN_CLOSE,
 
+   CT_GPAREN_OPEN,      // parenthetis after _Pragma                                     Issue #2286
+   CT_GPAREN_CLOSE,
+
    CT_LPAREN_OPEN,      // lambda-declarator parenthetis
    CT_LPAREN_CLOSE,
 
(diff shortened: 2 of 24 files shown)


src/tokenizer/tokenize.cpp
```diff
@@ -1656,7 +1656,20 @@ static bool parse_word(TokenContext &ctx, Chunk &pc, bool skipcheck)
       {
          // Turn it into a keyword now
          // Issue #1460 will return "COMMENT_CPP"
+#ifdef DEBUG
```
micheleCTDEAdmin, line 1659: What is the purpose of this debug code? It does exactly the same thing as the `else` branch and does not print any debug info. Am I missing something?

src/tokenizer/combine.cpp
```diff
@@ -1904,6 +1903,34 @@ void do_symbol_check(Chunk *prev, Chunk *pc, Chunk *next)
          pc->SetType(CT_INCDEC_BEFORE);
       }
    }
+
+   // Issue #3386
+   if (pc->Is(CT_PP_F_PRAGMA))
+   {
+      // such as  _Pragma("P2")
+      LOG_FMT(LFCNR, "%s(%d): GGG orig line is %zu, orig col is %zu, Text() '%s', type is %s\n",
```
micheleCTDEAdmin, line 1911: Is `GGG` in the log message something temporary? If so, we should remove. If not, maybe it would be better to use something a bit more explanatory

src/token_enum.h
```diff
@@ -324,6 +330,13 @@ enum E_Token
    CT_PP_UNDEF,
    CT_PP_PROPERTY,
 
+   // the next two PP types MUSST be BEFORE CT_PP_BODYCHUNK or AFTER CT_PP_OTHER
```
micheleCTDEAdmin, line 333: Typo: `MUSST ` --> `MUST`

src/chunk.h
```diff
@@ -1713,6 +1713,7 @@ inline bool Chunk::IsParenOpen() const
 {
    return(  Is(CT_PAREN_OPEN)
          || Is(CT_FPAREN_OPEN)
+         || Is(CT_GPAREN_OPEN)
```
micheleCTDEAdmin, line 1716: There may be other places in the code where there is a similar check for open or closed parenthesis, without the use of `IsParenOpen`/`IsParenClose`. Worth doing a search in the code, IMO.

micheleCTDEAdmin (commented):
A few comments to review

src/tokenizer/tokenize.cpp
guy-maurel: It is a better way for me to debug the statements. YES it is the same code.

src/tokenizer/combine.cpp
guy-maurel: Yes, thanks you for the catch!

src/token_enum.h
guy-maurel: Thanks

src/chunk.h
guy-maurel: I don't understand what you wish. Please make a proposal.

src/tokenizer/tokenize.cpp
micheleCTDEAdmin: Yes, I figured it was for debugging purposes. But once we have prepared a fix for the problem, I think it does not make much sense to include the debug code in the final code that will be committed, since it does nothing different

micheleCTDEAdmin (approved):
It looks good. Please see the latest comment I added in `tokenizer.cpp` for the debug code. Will leave the final choice to you.

src/chunk.h
micheleCTDEAdmin: Look for example at https://github.com/uncrustify/uncrustify/blob/master/src/indent.cpp#L2549-L2555: here there is a check for paren open, but it does not use `IsParenOpen`. So the new `CT_GPAREN_OPEN` is not checked. By memory, there are sure other places similar to this.
Anyway this could be solved in a different PR if you wish.

guy-maurel:
Yes, I want to have a good basis to debug.

etrian-dev:
Hi. I tested the latest version in this PR with the original code from #3386 .
While the program indeed does not crash, the output does not look right, as it causes line splicing.
In this case this does not cause any difference in the semantics of the code, but I'd say it's unintended.

original source:

int f (void)
{
_Pragma("P2");}

command:
`./uncrustify -c p.cfg p.c`

p.cfg:
`newlines=auto`

output:

int f (void)
{
	_Pragma("P2"); \\
}

guy-maurel:
@etrian-dev Thanks for the message

etrian-dev:
Hi @guy-maurel , this last commit looks good on the testcase.

gmaurel (approved):

micheleCTDE:
This caused a regression described in issue #4401. Commit 0271c39e9 is good and commit 53d53a55b (where this PR got merged) shows the problem.

micheleCTDE:
This PR is not good. It only works if a brace is the last token of a file.
For example this code works fine
```
int f (void)
{
_Pragma("P2");}
```

But as soon as we add a new tolek after the brace it will no longer work correctly.
```
int f (void)
{
_Pragma("P2");}

int i;
```

This is also confirm by looking into the [code](https://github.com/uncrustify/uncrustify/blob/master/src/newlines/brace_pair.cpp#L451)

@gmaurel: I propose we revert this PR because it is not fixing the intended issue #3386 correctly, while at the same time causing an important regression (new lines missing in front of the last brace of a file).
We should also consider whether a 0.80.1 release is needed.

clason:
It _does_ block updating uncrustify in homebrew, though.

clason:
No, it's just that one of the "very limited" cases happens to be the test case homebrew checks.

guy-maurel:
Yes!
Do, please, a revert.

micheleCTDE:
PR #4405 reverts this PR and also add the simple 'homebrew' test, so in future we can make sure we don't break it.

@guy-maurel: should I create a 0.80.1 release for this? 0.80.0 breaks homebrew test case, so they are not able to adopt it and they would be stuck with 0.79.0 for the next 6 months.

micheleCTDE:
I got confirmation from Guy through email to create a 0.80.1 release.

micheleCTDE:
Release 0.80.1 is out :slightly_smiling_face:

clason:
already cooking: https://github.com/Homebrew/homebrew-core/pull/197991 :)

PR #4373, opened 2024-09-18
test the nl_ options

guy-maurel:
This is the begin of extended tests. Some more will follow.


micheleCTDEAdmin (approved):
Nice idea. It seems good

Issue #4354, opened 2024-07-18
Add option for space after embedded comment followed by close parenthesis


micheleCTDE:
PR #4328 modified the behavior of spaces after an embedded comment. It would be good to add an option to allow specifying different spacing after an embedded comment for the case when it is and it is not followed by a closed parenthesis.
For example:
```
// Here a space after an embedded comment is nice to have
if (!line.isEmpty() /* && !stream.eof() */ && !isComment(line))

// Here if preferrable not to have a space after an embedded comment
if (!line.isEmpty() /* && !stream.eof() */)
```
Currently the same value is used for both options, so it is not possible to achieve the above formatting (unless I am missing something).

micheleCTDE:
Solved by PR #4380.

PR #4328, opened 2024-06-20
provide test data for the option sp_num_after_emb_cmt

guy-maurel:
fixes #4327

Issue #4327, opened 2024-06-19
stmt never reached


guy-maurel:
The case a config file with
```
sp_after_emb_cmt      = force
sp_num_after_emb_cmt  = 3
```

was never reached
I:
FooBar::CFooBar(CWnd* pParent/*=NULL*/);
E:
CFooBar::CFooBar(CWnd* pParent /*=NULL*/   );

micheleCTDE:
New solution in PR #4380 has replaced the original solution. The new option `sp_emb_cmt_priority` can now be used to control between old and new behavior. By default, the old behavior is preserved. If the option is set to `true`, then embedded comment spacing options will have priority over other spacing options (parenthesis, commas, braces, ...) providing the beharior required for this issue.

PR #4380, opened 2024-10-11
Fix/issue/4354

micheleCTDE:
The solution for issue #4327 provided by PR #4328 affected too many places without providing existing users an option to keep the old behavior. This was likely to raise user complains after the coming release.
The new solution achieves the following things:
1. replace the previous solution for issue #4327, providing users an option to enable the old (default) or new behavior
2. solve the regression explained in issue #4354
3. extend the logic of #4327 to each side of an embedded comment (before and after)

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -1,4 +1,4 @@
-# Uncrustify-0.79.0
+# Uncrustify-0.79.0-dev
 
 #
 # General options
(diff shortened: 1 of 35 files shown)


micheleCTDE:
@guy-maurel let me know if you are ok with this solution instead of #4328.

gmaurel (approved):
It looks fine!
Thanks you
I let you merge.

PR #4406, opened 2024-11-16
Update minimum cmake version to 3.10

dundargoc:
This is to prevent the following warning:

CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.10)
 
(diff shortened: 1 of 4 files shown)


micheleCTDEAdmin (commented):
Looks good, thanks @dundargoc

micheleCTDEAdmin (approved):

PR #4409, opened 2024-11-17
next step D for too big

guy-maurel:


micheleCTDE:
In `test_cli_options.py` we rnu a number of tests and look if anything fails using the `return_flag` bool. 
So I wonder what is the idea behind a different logic only for the "too long" tests: why do we want to exit immediately in such case? What is the problem we are trying to address?

guy-maurel:
Yes, I change a little bit.
Tests with a for-loop should stop early, I think, the bug should be solved
as soon as possible.
Many tests are commented out. I want to re active them.

micheleCTDE:
Ok I understand, It is a matter of pros and cons. Stop early means you can only see one failing test per run, but you stop early. Non stopping early means you can see all failing tests at once, but it may take longer to run. I am ok either way.

micheleCTDE (approved):

PR #4415, opened 2024-11-20
fix tab indentation for blank lines

olemayu:
Blank lines were indented using spaces even though `indent_with_tabs` specifies that 1 and 2 indent with tabs.
```cfg
indent_single_newlines = true
indent_with_tabs = 1
```

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -849,11 +849,11 @@ void output_text(FILE *pfile)
 
                if (pc->IsPreproc())
                {
-                  output_to_column(pc->GetNlColumn(), (pp_indent_with_tabs == 2));
+                  output_to_column(pc->GetNlColumn(), (pp_indent_with_tabs >= 1));
                }
                else
                {
-                  output_to_column(pc->GetNlColumn(), (options::indent_with_tabs() == 2));
+                  output_to_column(pc->GetNlColumn(), (options::indent_with_tabs() >= 1));
                }
             }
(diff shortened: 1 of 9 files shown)


micheleCTDE (changes requested):
You need to add test cases showing the original problem and the expected fixed version.

micheleCTDE (changes requested):
The config files should be in the `c` folder and not in the `common` folder. A file is supposed to be in `common` only if it is used in tests by multiple languages.
You can also squash the commit into a single one, since logically they are one entity.
Other than that it looks good.

olemayu:
There we go! I have never squashed commits before so I hope I did it correctly.

micheleCTDE:
> There we go! I have never squashed commits before so I hope I did it correctly.

Thanks :-)
Squash was good, although the commit message will need to be adjusted before merging. Currently as result of the squash it has a lot of duplication.
In any case I started the automated tests, so they can all pass first.

micheleCTDE:
All the automated tests have passed. There is jsut onle thing to do before merging, which is to amend the current commit message from:
```
use tabs for newline column output
add tests

use tabs for newline column output

add tests

move tests
```
to something like:
```
use tabs for newline column output
```

micheleCTDEAdmin (approved):
It looks good

micheleCTDE (approved):

Issue #575, opened 2016-07-21
Wrong newlines after = assignments


Massimo-B:
``` cpp
$ (cfg="https://bitbucket.org/massimo/configs/raw/72329a67bd4fa2f321faf3e6df1ab6e7dd90f262/uncrustify.cfg" ; input="https://bitbucket.org/massimo/test/raw/c7fec4fca601bec29cb35946ab92edd642bda02b/crusty-test_001.cpp" ; uncrustify -v ; curl -s "$input" ; curl -s "$input" | uncrustify -l cpp -c <(curl -s "$cfg") )

uncrustify 0.63
void Foo::doo()
{
    m_stackCache[m_currentStackNr]->operator [](0) = new QStandardItem(QString::number(m_currentStackNr));
    m_stackCache[m_currentStackNr]->operator [](1) = new QStandardItem(tr("OK"));
    m_stackCache[m_currentStackNr]->operator [](2) = new QStandardItem("0");
    m_stackCache[m_currentStackNr]->operator [](3) = new QStandardItem("0");
    m_stackCache[m_currentStackNr]->operator [](4) = new QStandardItem();
}                                                                                                   
Parsing: 536 bytes (536 chars) from stdin as language CPP
void                                         Foo::doo()
{
    m_stackCache[m_currentStackNr]->operator [](0) = new QStandardItem(QString::number(m_currentStackNr));
    m_stackCache[m_currentStackNr]->operator [](1) = new QStandardItem(tr("OK"));

    m_stackCache[m_currentStackNr]->operator [](2) = new QStandardItem("0");

    m_stackCache[m_currentStackNr]->operator [](3) = new QStandardItem("0");

    m_stackCache[m_currentStackNr]->operator [](4) = new QStandardItem();
}
```

There are too many additional newlines that are only added for some of those lines.

Beside that, the function name is aligned to operator argument of the body which does not make sense at all. Is that possible to configure or a bug?

mihaipopescu:
Can you edit your issue (add some markup https://guides.github.com/features/mastering-markdown/) to show where the problem is ? You can even provide a diff between the expected and the output.

Massimo-B:
Original:
```cpp
void Foo::doo()
{
    m_stackCache[m_currentStackNr]->operator [](0) = new QStandardItem(QString::number(m_currentStackNr));
    m_stackCache[m_currentStackNr]->operator [](1) = new QStandardItem(tr("OK"));
[...]

PR #4417, opened 2024-11-22
Do not mistake explicit operator== call for function prototype

ltomovic-ibl:
This code
```
bool operator ==(const Derived& item) const
{
	return property == item.property
		&& Base::operator==(item);
}
```
is not properly tokenized. `Base::operator==(btt)` is recognized as a function prototype, but it should be actually function call. I attached tokens dump. [These lines of code](https://github.com/uncrustify/uncrustify/blob/28d4229f3c17a1371abe37bd07a5e8d6168805c5/src/tokenizer/combine_fix_mark.cpp#L1252-L1269) should actually solve this. However, this code is unreachable because of [this](https://github.com/uncrustify/uncrustify/blob/28d4229f3c17a1371abe37bd07a5e8d6168805c5/src/tokenizer/combine_fix_mark.cpp#L1219), which was meant to solve issue [575](https://github.com/uncrustify/uncrustify/issues/575). Uncrustify had probably went a long way since than and now removing this solves my problem and the problem in [575](https://github.com/uncrustify/uncrustify/issues/575) isn't broken by this anymore. 

[example.txt](https://github.com/user-attachments/files/17866064/example.txt)

### DIFF
--- a/src/tokenizer/combine_fix_mark.cpp
+++ b/src/tokenizer/combine_fix_mark.cpp
@@ -1216,7 +1216,6 @@ void mark_function(Chunk *pc)
          while ((tmp = tmp->GetPrevNcNnlNi())->IsNotNullChunk()) // Issue #2279
          {
             if (  tmp->Is(CT_BRACE_CLOSE)
-               || tmp->Is(CT_BRACE_OPEN)             // Issue 575
                || tmp->Is(CT_SEMICOLON))
             {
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin (approved):
It looks good, actually solves a few issues with operator= too. Though there are still some extra work to do since I found other corner cases.

Issue #4410, opened 2024-11-18
Wrong indentation of function call arguments when combined with "<<"


mmichalek-ibl:
Hello,
arguments inside function call are indented wrongly when `<<` operator is used as well. 2 tabs are used instead of 1 tab. It seems this issue happens when both `indent_continue` and `indent_shift` rules are used at once.

current version: Uncrustify-0.79.0_f

Test file:
```
void foo()
{
	THROW_EXCEPTION(
		"exception reason: %1"
	) << getLastErrorMsg();
}
```

Generated Output:
```
void foo()
{
	THROW_EXCEPTION(
			"exception reason: %1"
		) << getLastErrorMsg();
}
```
Line with `exception reason: %1` is indented wrongly with 2 tabs. It should be indented only with 1 tab. Also line `) << getLastErrorMsg();` is indented wrongly - rule `indent_paren_close` should be applied here.

Expected results:
```
void foo()
{
	THROW_EXCEPTION(
		"exception reason: %1"
	) << getLastErrorMsg();
}
```

Config file:
```
input_tab_size = 4
indent_columns = 4
indent_with_tabs = 2
output_tab_size = 4
indent_paren_close = 2

indent_shift = 1
align_left_shift = false

indent_continue = indent_columns
```

[parsed.txt](https://github.com/user-attachments/files/17798432/parsed.txt)

ltomovic-ibl:
It looks like indent_shift = 1 is causing this. It is interesting that there is a unit test with expected file `30291-indent_shift.cpp` that contains among other things also this
```
    return theAddr.addrN().family() == AF_INET6 ?
        (theAddr.octet(idx * 2) << 8) + theAddr.octet(idx * 2 + 1) :
        theAddr.octet(idx);}
```
which I believe to be wrong. If we replace the shift by multiplication, i.e.
```
    return theAddr.addrN().family() == AF_INET6 ?
              (theAddr.octet(idx * 2) * 8) + theAddr.octet(idx * 2 + 1) :
              theAddr.octet(idx);
```
then it looks like this, it is indented according to the return keyword. However, in case there is a shift, it is all considered as shift expression and then it is indented as the first example. But the shift expression is only this `(theAddr.octet(idx * 2) << 8)`.

Documentation regarding `indent_shift` rule says this
```
# How to indent continued shift expressions ('<<' and '>>').
# Set align_left_shift=false when using this.
#  0: Align shift operators instead of indenting them (default)
#  1: Indent by one level
[...]

PR #4418, opened 2024-11-22
Fix indent_shift = 1 behaviour to not indent irrelevant lines

ltomovic-ibl:
Related to #4410

This fixes problem mentioned in #4410 . It also makes more sense this way, since only truly continued shift expressions are indented, as the rule says.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -3337,65 +3337,14 @@ void indent_text()
          bool in_shift    = false;
          bool is_operator = false;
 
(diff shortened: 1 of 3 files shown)


micheleCTDE:
I will try to test this tomorrow if I can

micheleCTDEAdmin (approved):
It looks good. The changes in the expected tests are correct.
Well done @ltomovic-ibl!

PR #4419, opened 2024-11-23
Remove 'using namespace' from header file

PoeticPete:
`using namespace` in headers files can unexpectedly cause underlying files to use the namespace. Clang even has a warning for this. Here's a [link](https://bugzilla.mozilla.org/show_bug.cgi?id=1443444) to Firefox enabling the warning.

This PR just removes the single instances of `using namespace std;` in a header file

### DIFF
--- a/src/pcf_flags.h
+++ b/src/pcf_flags.h
@@ -16,9 +16,6 @@
 //#define ARRAY_SIZE(x)    (sizeof(x) / sizeof((x)[0]))
 //#endif
 
-using namespace std;
-
-
 constexpr auto pcf_bit(size_t b) -> decltype(0ULL)
 {
    return(1ULL << b);
--- a/src/pcf_flags.h
+++ b/src/pcf_flags.h
@@ -11,10 +11,18 @@
 #include "enum_flags.h"
 #include "logger.h"
 
+
 //// and the ever-so-important array size macro
+
+
 //#ifndef ARRAY_SIZE
+
+
 //#define ARRAY_SIZE(x)    (sizeof(x) / sizeof((x)[0]))
+
+
 //#endif
+
 
 constexpr auto pcf_bit(size_t b) -> decltype(0ULL)
 {
--- a/src/pcf_flags.h
+++ b/src/pcf_flags.h
@@ -12,18 +12,6 @@
 #include "logger.h"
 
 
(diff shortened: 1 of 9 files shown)


micheleCTDE:
There should at least be an option to enable/disable this feature, default being that it is disabled.
Also why only the `std` namespace but not others? Perhaps we can use `std` as default but allow the user to specify a list of namespace names through the config.
Or alternatively we don't remove but simply provide a warning to the user, since it is quite possible that removing a namespace from a header file without any further changes will let to broken code.

PoeticPeteAdmin:
>There should at least be an option to enable/disable this feature, default being that it is disabled.

As in, add a build flag to fail builds that use namespace in header files? Or gate the 'using namespace' call site? It's generally agreed this is a bad pattern, so I don't think we should allow it at all.

>Also why only the std namespace but not others?

std is the only namespace currently being used in a header file 

>Perhaps we can use std as default but allow the user to specify a list of namespace names through the config.

Are you referring to other developers who might have compiled uncrustify into their own repo, and are importing the header files? Allowing other developers to specify a namespace would be the most configurable option. I don't think making header files configurable adds much value.

Happy to hear examples where configurability might be useful though :)

micheleCTDE:
Ah sorry Pete, I was in a bit of a hurry and thought this was a PR to allow to remove namespaces from processed files. Instead now that I looked at the code in details, I see it removes the namespace from uncrustify code itself. Then my comments/suggestions above don't make any sense and you can simply ignore them.
Should never comment in a hurry. I know that but I still fall trap of this mistake sometimes :cold_sweat:

PoeticPeteAdmin:
Oh no worries at all, I appreciate the quick turnaround Michele!

src/pcf_flags.h
```diff
@@ -11,12 +11,17 @@
...
+
 //#ifndef ARRAY_SIZE
+
+
 //#define ARRAY_SIZE(x)    (sizeof(x) / sizeof((x)[0]))
-//#endif
 
-using namespace std;
+
+//#endif
```
PoeticPete, line 24: new lines need to be added to make the format test pass. Will remove in another PR.

micheleCTDEAdmin (approved):
Looks good

PR #4422, opened 2024-11-23
a better cli for p-test

guy-maurel:


PoeticPeteAdmin (approved):
Are these early exits to make the tests fail faster, if there is no output?

If so, makes sense to me 👍

Issue #4394, opened 2024-11-07
Formatting of complex boolean expressions.


Moroxus:
Is it possible to get a formatting of boolean expressions like this?
```
			if (((i_type != i_typeFIR) || !b_useFIR)
				&& ((i_type != i_typeUIR) || !b_useUIR)
				&& ((i_type != i_typeCTA) || !b_useCTA)
				&& ((i_type != i_typeFIR_P) || !b_useFIR_P)
				&& !b_isIWXXM
			)
```

I mean, I would like it to fit on the line, but if its too long, then every operand should be on a separate line. And then recursively. In this case, operands of && can fit on a line so there is no need to put also nested || on new line.

I was able to get something similar with function calls:
```
				query.query(
					collection,
					i_type,
					spatial,
					bcp::OPMETQueryTimeHistorical(
						bcp::OPMETQueryTimeHistorical::fdReceptionTime,
						stamp,
						anotherArgument
					),
					bcp::OPMETEnums::asDisabled
				);
```

Here if the function call does not fit on a line, every argument is on a separate line. In case an argument is too long, apply the same rule again, in this case call of OPMETQueryTimeHistorical.

I tried 

`pos_bool                        = lead_break   # ignore/break/force/lead/trail/join/lead_break/lead_force/trail_break/trail_force`

And I got to the point were it is broken on every single operator even in case it would fit on a line. Then it looks like this

```
			if (((i_type != i_typeFIR)
				 || !b_useFIR)
				&& ((i_type != i_typeUIR)
					|| !b_useUIR)
				&& ((i_type != i_typeCTA)
					|| !b_useCTA)
				&& ((i_type != i_typeFIR_P)
					|| !b_useFIR_P)
				&& !b_isIWXXM
			)
```

 I tried all possible options of `pos_bool`. Maybe I should combine it with some other setting? I do not know. Is it even possible to achieve?

[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[...]

PR #4426, opened 2024-11-25
Add nl_bool_expression_multiline option

ltomovic-ibl:
It looks like functionality requested in #4394 was not possible. So I implemented new option nl_bool_expression_multiline.

### DIFF
--- a/documentation/htdocs/config.txt
+++ b/documentation/htdocs/config.txt
@@ -2189,6 +2189,12 @@ nl_after_vbrace_close           = false    # true/false
 # structures. If set to ignore, uses nl_after_brace_close.
 nl_brace_struct_var             = ignore   # ignore/add/remove/force
 
+# Whether to add a newline before/after each '&&' or `||` on the same
+# nesting level in a boolean expression if boolean expression will not
+# fit on a line. code_width needs to be positive and pos_bool needs
+# to be 'lead' or 'trail' for this option to have any effect.
+nl_bool_expr_hierarchical       = false    # true/false
+
 # Whether to alter newlines in '#define' macros.
 nl_define_macro                 = false    # true/false
 
--- a/documentation/htdocs/default.cfg
+++ b/documentation/htdocs/default.cfg
@@ -2189,6 +2189,12 @@ nl_after_vbrace_close           = false    # true/false
 # structures. If set to ignore, uses nl_after_brace_close.
 nl_brace_struct_var             = ignore   # ignore/add/remove/force
 
+# Whether to add a newline before/after each '&&' or `||` on the same
+# nesting level in a boolean expression if boolean expression will not
+# fit on a line. code_width needs to be positive and pos_bool needs
+# to be 'lead' or 'trail' for this option to have any effect.
+nl_bool_expr_hierarchical       = false    # true/false
+
 # Whether to alter newlines in '#define' macros.
 nl_define_macro                 = false    # true/false
 
--- a/etc/defaults.cfg
+++ b/etc/defaults.cfg
@@ -2189,6 +2189,12 @@ nl_after_vbrace_close           = false    # true/false
 # structures. If set to ignore, uses nl_after_brace_close.
 nl_brace_struct_var             = ignore   # ignore/add/remove/force
 
+# Whether to add a newline before/after each '&&' or `||` on the same
+# nesting level in a boolean expression if boolean expression will not
+# fit on a line. code_width needs to be positive and pos_bool needs
+# to be 'lead' or 'trail' for this option to have any effect.
+nl_bool_expr_hierarchical       = false    # true/false
+
 # Whether to alter newlines in '#define' macros.
 nl_define_macro                 = false    # true/false
 
--- a/etc/uigui_uncrustify.ini
+++ b/etc/uigui_uncrustify.ini
@@ -5447,6 +5447,15 @@ ChoicesRegex=nl_brace_struct_var\s*=\s*ignore|nl_brace_struct_var\s*=\s*add|nl_b
 ChoicesReadable="Ignore Nl Brace Struct Var|Add Nl Brace Struct Var|Remove Nl Brace Struct Var|Force Nl Brace Struct Var"
 ValueDefault=ignore
 
+[Nl Bool Expr Hierarchical]
+Category=3
+Description="<html>Whether to add a newline before/after each '&amp;&amp;' or `||` on the same<br/>nesting level in a boolean expression if boolean expression will not<br/>fit on a line. code_width needs to be positive and pos_bool needs<br/>to be 'lead' or 'trail' for this option to have any effect.</html>"
+Enabled=false
+EditorType=boolean
+TrueFalse=nl_bool_expr_hierarchical=true|nl_bool_expr_hierarchical=false
+TrueFalseRegex=nl_bool_expr_hierarchical\s*=\s*true|nl_bool_expr_hierarchical\s*=\s*false
+ValueDefault=false
+
 [Nl Define Macro]
 Category=3
 Description="<html>Whether to alter newlines in '#define' macros.</html>"
--- a/src/newlines/chunk_pos.cpp
+++ b/src/newlines/chunk_pos.cpp
@@ -25,6 +25,15 @@ void newlines_chunk_pos(E_Token chunk_type, uncrustify::token_pos_e mode)
 {
    LOG_FUNC_ENTRY();
 
+   if (  chunk_type == CT_BOOL
+      && options::code_width() > 0
+      && options::nl_bool_expr_hierarchical()
+      && (  options::pos_bool() == TP_LEAD
+         || options::pos_bool() == TP_TRAIL))
+   {
+      // here treat as join and let do_code_width do the heavy lifting
+      mode = TP_JOIN;
+   }
    LOG_FMT(LNEWLINE, "%s(%d): mode is %s\n",
            __func__, __LINE__, to_string(mode));
 
--- a/src/options.h
+++ b/src/options.h
@@ -2679,6 +2679,13 @@ nl_after_vbrace_close;
 extern Option<iarf_e>
 nl_brace_struct_var;
 
+// Whether to add a newline before/after each '&&' or `||` on the same
+// nesting level in a boolean expression if boolean expression will not
+// fit on a line. code_width needs to be positive and pos_bool needs
+// to be 'lead' or 'trail' for this option to have any effect.
+extern Option<bool>
+nl_bool_expr_hierarchical; // = false
+
 // Whether to alter newlines in '#define' macros.
 extern Option<bool>
 nl_define_macro;
--- a/src/width.cpp
+++ b/src/width.cpp
@@ -102,6 +102,14 @@ static void split_fcn_params_full(Chunk *start);
 
 
 /**
+ * Splits line on every boolean expression at the top level and then
+ * recursively on nested levels in case splitted lines are still too long.
+ * @param start
+ */
+static void split_bool_expr(Chunk *start);
+
+
+/**
  * A for statement is too long.
  * Step backwards and forwards to find the semicolons
  * Try splitting at the semicolons first.
@@ -336,6 +344,30 @@ static void try_split_here(SplitEntry &ent, Chunk *pc)
       }
    }
 
+   if (  options::nl_bool_expr_hierarchical()
+      && (  options::pos_bool() == TP_LEAD
+         || options::pos_bool() == TP_TRAIL))
+   {
+      // comma has a higher priority than bool, but in case nl_bool_expr_hierarchical is turned on
+      // we want to break on bool if it is less nested than comma, so here the priority inversion is handled
+      if (  pc->Is(CT_COMMA)
+         && ent.pc->Is(CT_BOOL)
+         && ent.pc->GetLevel() < pc->GetLevel())
+      {
+         LOG_FMT(LSPLIT, "%s(%d): don't break after comma if there is a bool on less nested level\n", __func__, __LINE__);
+         return;
+      }
+
+      if (  pc->Is(CT_BOOL)
+         && ent.pc->Is(CT_COMMA)
+         && ent.pc->GetLevel() > pc->GetLevel())
+      {
+         LOG_FMT(LSPLIT, "%s(%d): found possible split on bool, because comma is on more nested level\n",
+                 __func__, __LINE__);
+         return;
+      }
+   }
+
    if (  ent.pc->IsNullChunk()
       || pc_pri < ent.pri
       || (  pc_pri == ent.pri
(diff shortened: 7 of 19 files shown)


micheleCTDE:
I will have a look at this in the next few days.

src/uncrustify.cpp
```diff
@@ -2227,7 +2227,7 @@ void uncrustify_file(const file_mem &fm, FILE *pfout, const char *parsed_file,
       }
       newlines_insert_blank_lines();
 
-      if (options::pos_bool() != TP_IGNORE)
+      if (options::pos_bool() != TP_IGNORE || (options::code_width() > 0 && options::nl_bool_expression_multiline()))
```
micheleCTDE, line 2230: I like the idea, but it needs more work. `nl_bool_expression_multiline` cannot override `pos_bool`, it has to work with it.
The implementation in this PR is basically forcing a `trail` split on long lines. This may not be what the user expects based on the `pos_bool` setting. For example if `pos_bool` is `ignore`, the user would expect no changes at all. If `pos_bool` is `lead`, the user would expect the logical operator to be at the beginning of the next line, not at the end of the current long line.
`nl_bool_expression_multiline` should be active if `pos_bool` is different from `ignore`,  `join` or `break` and then it should place the logical operator based on the `pos_bool` value (`lead` or `trail`).

@guy-maurel what do you think about this?

micheleCTDE (changes requested):

micheleCTDE:
> This is actually not true. The implementation respects both lead and trail. Logical operators are either at the beginning or at the end of a line. I added unit test with only one configuration file. Maybe I should add also other with different pos_bool so it would be more obvious.

Ah great. I thought about this last night and forgot to comment that I may have mislooked at this part. So if it is already following lead/trail, this part is fine. Would be good to add a test case for trail config too.

>  I will also add check so this option is used only in case pos_bool is either lead or trail

Sounds like a plan :-) and a good addition to uncrustify once merged btw

ltomovic-ibl:
Ok, so I have made more unit tests, also this option is now used only if `pos_bool` is `lead` or `trail`.

micheleCTDE:
Great, I will have a look at the PR this week (sorry, very busy week till Friday)

micheleCTDE (changes requested):
There is at least one logical mistake to fix (see comment). Then I will do some regression tests on a big code base.
On a side note, there are a few other files to update when you add an option. You can compare with other options and add the missing chagnes.

micheleCTDE:
I have done a test on about 3200 files. In many places, this PR provides improvement to the splitting of boolean expressions, but I also found a couple or regressions at least.

1. some lines with comments at the end of the line get split in a wrong way, resulting in the comments at the end of the line being pass the code width mark. For example I get
```
	if (fieldname == TQString::fromLatin1("o") || fieldname == TQString::fromLatin1("organization") || // Exchange
	        fieldname == TQString::fromLatin1("organizationname"))                                     // Exchange
```
while before I had
```
	if (fieldname == TQString::fromLatin1("o") ||
	        fieldname == TQString::fromLatin1("organization") ||     // Exchange
	        fieldname == TQString::fromLatin1("organizationname"))   // Exchange
```
The comment of the first line exceeds the 100 mark. I think the boolean splitting logic does not consider the comment length. This could be a separate issue just highlighted by this PR, but disabling the ne option does not show this problem, so I consider this a regression and therefore should be investigated.

2. This is a bit of a weird one and maybe related to another PR we recently merged. This code gets formatted as below when the new multiline option is true
```
	return TQString::localeAwareCompare(a1.givenName(), a2.givenName()) == 0 &&
	               TQString::localeAwareCompare(a1.familyName(), a2.familyName()) == 0;
```
and as below when the option is false:
```
	return TQString::localeAwareCompare(a1.givenName(), a2.givenName()) == 0 &&
	       TQString::localeAwareCompare(a1.familyName(), a2.familyName()) == 0;
```
Note the additional unnecessary indentation of the second line. Again this is a regression.
If you need I can share an example of config and code used for this testing.

Overall I see a lot of improvements in splitting of long bool expressions. We just need to understand what is causing these regressions and fix them.
On a positive note, with the option set to false I see no changes/regression.

ltomovic-ibl:
Yeah it would be great if you could attach config and a piece of code right here. I can investigate it and try to fix both issues.

micheleCTDE:
Sure, will put it together this weekend and share with you.

micheleCTDE:
Attached a zip file showing the problems mentioned above. Configuration, input and output files are included.
I feel like the regressions are actually existing issues that are exposed by the use of `join` when `nl_bool_expression_multiline` is true, although I could be mistaken of course.
Once you have taken a look, happy to discuss together on how to proceed with this.

[regression_pr_4426.zip](https://github.com/user-attachments/files/18135137/regression_pr_4426.zip)

micheleCTDE:
> I feel like the regressions are actually existing issues that are exposed by the use of join

Just done a quite test. I set `nl_bool_expression_multiline` to `false` and set `pos_bool` to `join` and the same issues appeared. So I guess we have two choices:

1. merge the PR as is and work on those issues separately
2. consider whether it makes sense to force `join` when `nl_bool_expression_multiline` is true. To be honest, I can't see how this PR would work if `nl_bool_expression_multiline` is true and we don't force `join`, since it would defy the whole logic of it.

What is your opinion about it? If we go for option 1., perhaps it would make sense to rename `nl_bool_expression_multiline` to something like `nl_bool_at_same_level`. I feel the original name does not really convey a good idea of what the option is doing.

ltomovic-ibl:
Hi, thank you very much for files, analysis and insight you provided.

Regarding the name, I must admit, it was hard to find the right name, I agree that `nl_bool_expression_multiline` is not perfect. I was kind of trying to be consistent with other options that are doing similar thing for definitions, like here
```
nl_func_def_start_multi_line     = true
nl_func_def_args_multi_line     = true
nl_func_def_end_multi_line      = true
``` 
. However, these options are three and for bool I made just one. I mean, I am happy to change the name for something better, but I do not really understand `nl_bool_at_same_level` name. What is meant by level here? I mean, I had similar name chosen for the other pull request that is trying to implement some indentation. In my head, word level is often associated with indent level in formatting context.

Regarding forcing of `join` in `pos_bool`, the current implementation works in a way the same code is used as if `pos_bool` would be `join`, however, it demands `pos_bool` to be either `lead` or `trail`, which is necessary to decide whether the boolean operators should be at the beginning or end of a line.

I will try to look at the input, output and configuration files you provided these days and I will try to debug it a little bit and examine what exactly is happening. Hope I will find enough time for it before Christmas, it is really busy time now. I will try to be in touch. Thank you again for your effort.

micheleCTDE:
> but I do not really understand nl_bool_at_same_level name. What is meant by level here? 

The idea behind it is that if the line is too long, it will be split at the same boolean level. Let me make an example:
```
if ((true && false) ||
    (hello && world))
```
rather than
```
if ((true &&
     false) ||
    (hello && world))
```
The `&&` is at a inner level than the `||`, so for long expression we split at the same level, if I understood the code correctly. But if the name seems inappropriate we can consider something else, mine was just a possible suggestion.

> Regarding forcing of join in pos_bool,

Unless I misunderstand the code, without `join` it would not work correctly, because long lines could be splitted in awkward ways initially. So the extra `join` in the code when `lead` or `trail` is used seems necessary to me.

> I will try to look at the input, output and configuration files you provided these days 

Sure, there is no rush. When you have more info or feedback or questions, just comment here and we will continue the discussion.

ltomovic-ibl:
Hi,

finally I found some time to play with this pull request a little. And I found something. Let's first talk about the additional indentation level in return statement, like this:
```
	return TQString::localeAwareCompare(a1.givenName(), a2.givenName()) == 0 &&
	               TQString::localeAwareCompare(a1.familyName(), a2.familyName()) == 0;
```
I am not sure what is the correct behaviour, but the additional indentation level is here because it is considered continuation line. The logic that does splitting on boolean operators uses alredy existing function `split_before_chunk` which I did not modify in any way. The function calls also this `pc->SetFlagBits(PCF_CONT_LINE);` - it marks the line as continuation.

When `nl_bool_expression_multiline = false`, this just isn't called and by default the next line is not considered as continuation line. I am still not sure whether this is correct or no. Anyway, I modified my function to find out whether the boolean operator is part of the return statement and whether it is on the same level as return statement and if yes, I clear the `PCF_CONT_LINE` so it would not be considered a continuation line.

The second issue with comment exceeding the line length turned out to be much easier to analyze. There is this code
https://github.com/uncrustify/uncrustify/blob/36708116db2e195398b2d1a262ca9e06836d3199/src/width.cpp#L219-L221

which makes sure that comments are not taken into account when checking whether line is too long. However, I did not modify this code in any way so, I am not sure whether this is intentional or not. It looks like intentional and If I would change that I think there would be more tests that would fail.

ltomovic-ibl:
Regarding the continuation lines on return statement, I think the best would be to make an option so a user can choose. However, I agree that this should be addressed in a separate PR. The current behaviour I believe is this:

> disregard the extra continuation indentation for contination lines on return

and I modified the code I wrote so it would be consistent with the rest of uncrustify. I think that should be enough for this PR.

> If we are splitting a line, I think we should mark it as continuation line, regardless of the nl_bool_expression_multiline value, because it is about long lines being splitted, not about how and where we splitted them.

I meant something else here. If the file on the input is already formatted the way that there is multiline return statement, those next lines are not considered continuation lines. No splitting happened. File is just parsed and when indentation is happening, next lines on return statement are not considered as continuation lines. I guess there is somewhere logic that recognizes that those lines are a part of return statement. However, in case `nl_bool_expression_multiline = true`, those lines are joined (in case there are boolean operators) and then split again on bool operators and this splitting marks them as continuation lines. I did not modify `split_before_chunk` in any way. It still marks it as continuation line and I agree that it should be kept that way. Splitting a line should mark next line as continuation. However, `split_bool_expr` is a new function that I wrote and in this function I am handling the return statement so it would be consistent with what would happen in case `nl_bool_expression_multiline = false`. I mean, indentation of multiline return statement should be the same no matter the `nl_bool_expression_multiline`.

Is there something else that should be addressed in this PR? I guess that you would like to run regression tests on bigger data to see whether my last modification did not break anything. We were also talking about renaming the option. I was thinking about possible names. Reality is that it is too complicated idea to be conveyed just by short option name, but anyway, two suggestions that came to my mind:

- nl_bool_expr_levels - suggests that breaking is going by levels
- nl_bool_expr_hierarchical - resulted formatting mirrors the statement hierarchy

I noticed that there are already options that have `expr` in them, like `nl_return_expr` and `nl_throw_expr`. My original name was `nl_bool_expression_multiline` so I guess I should at least shorten `expression` to `expr`. Let me know what you think about those names. I think I like the most `nl_bool_expr_hierarchical` for now, but I would like to hear another opinion.

micheleCTDE:
>>   disregard the extra continuation indentation for contination lines on return
> and I modified the code I wrote so it would be consistent with the rest of uncrustify. I think that should be enough for this PR.

Yes, if that is the case now, it should be good. 

> I mean, indentation of multiline return statement should be the same no matter the nl_bool_expression_multiline.

Thanks for the detailed explanation and yes, I also agree it should be the same.

>  Let me know what you think about those names.

Both `nl_bool_expr_hierarchical` and `nl_bool_expr_levels` are fine with me, so I will leave the final choice to you.

I will run some regression tests on a big code base during the weekend with more time avaialable. Will feedback here after that.

tests/input/c/nl_bool_expression_multiline.c
```diff
@@ -0,0 +1,31 @@
...
+bool complex_nested_conditions(int x, int y, int z) {
+	if ((x == y || (y < z && z > 20)) && ((x + y > z * 2) || (x - z < y * 3 && x > 100))) {
+		return true;
+	}
+	return false;
+}
+
+int main() {
+	int a = 15, b = 3, c = 8, d = 2;
+	int x = 5, y = 5, z = 25;
```
gmaurel, line 20: We like to use only ONE variable pro <type>line.
Please, make a little change

tests/input/c/nl_bool_expression_multiline.c
```diff
@@ -0,0 +1,31 @@
...
+
+	if (nested_boolean_example(a, b, c, d) && complex_nested_conditions(x, y, z)) {
+		printf("Both nested conditions are true.\n");
+	}
+
+	if ((a < b && (b == c || d > 100)) || ((a * b - d < c * c + b) && (d != 1))) {
```
gmaurel, line 26: We preffer only ONE logical statement pro line
```
if ((a < b
&& (b == c
 || d > 100)) 
|| ((a * b - d < c * c + b) 
&& (d != 1))) {
```

gmaurel (changes requested):
fine work, thanks

micheleCTDE:
I have done a big test on ~3200 files, it looks good, even the `return` statements. Once you update the code as per @guy-maurel's comment, this PR seems ready for merging.

tests/input/c/nl_bool_expression_multiline.c
ltomovic-ibl: Well, here I can make that happen, no problem, I will make the change.

tests/input/c/nl_bool_expression_multiline.c
ltomovic-ibl: I wanted to say this cannot be done, but then I realized this is the input file, not the output, so, it actually can be done. I am going to change it now.

ltomovic-ibl:
Well, I tried to rename the branch, but that effectively closed the PR... I renamed the option to nl_bool_exper_hierarchical and I also wanted the branch name to mirror this. But it looks like I cannot change the branch in this pull request. Can you? I made `nl_bool_expr_hierarchical` branch, but it would be great if this discussion to the PR would not be lost. In case the branch cannot be changed, the best way is to live with different branch name I guess.

gmaurel (approved):

micheleCTDE (approved):
Looks good

micheleCTDEAdmin:
> In case the branch cannot be changed, the best way is to live with different branch name I guess.

Renaming branches usually doesn't go well with an existing PR, so it is safer to simply keep the existing name, so all discussion remains. Alternatively, you would have to create a new branch and PR and put a comment in that PR referring to this one for comments and discussion. IMO, not worth it.

@ltomovic-ibl if you are ok with the current branch name, we can merge the PR.

ltomovic-ibl:
I am ok with that, if you are ok with the fact that branch name and option name are not consistent now. I read your guideline about contributing and this was one of the points mentioned there.

Thank you for your effort during this PR, from my point of view, it can be merged.

micheleCTDEAdmin:
It is of course preferrable to have the option and the branch with same/similar name, but in this case given the history of the PR it is acceptable to have a small difference.
Thanks for the good work and patience in updating it based on our feedback.

PR #4430, opened 2024-11-26
Fix indent_shift = 1 behaviour with multiline operands

ltomovic-ibl:
#4410 reported incorrect behaviour of indent_shift = 1 in certain cases. In PR #4418 I tried to fix this and I actually did, but my implementation did not handled correctly a corner case where the right operand of shift spreads across multiple lines. Something like this:
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
			reinterpret_cast<const char*>(error.getMessage().data()),
			error.getMessage().size()
		)
```
Such a use case was not covered by unit tests so this all passed. Now I changed the implementation to also cover this use case and I also added unit test that checks for this.

I am sorry that I did not find the corner case before PR #4418 was merged, I could save you some time, now you have to review twice.

EDIT:

Actual wrong output:
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
		reinterpret_cast<const char*>(error.getMessage().data()),
		error.getMessage().size()
		)
```

Expected correct output:
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
			reinterpret_cast<const char*>(error.getMessage().data()),
			error.getMessage().size()
		)
```

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -570,8 +570,10 @@ void indent_text()
    ParsingFrame      frm;
 
 
-   Chunk *pc        = Chunk::GetHead();
-   bool  classFound = false;                                 // Issue #672
+   Chunk *pc            = Chunk::GetHead();
+   bool  classFound     = false;                             // Issue #672
+   bool  in_shift       = false;
+   Chunk *current_shift = Chunk::NullChunkPtr;
 
    while (pc->IsNotNullChunk())
    {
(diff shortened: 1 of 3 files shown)


PoeticPete:
Thank you for coming back and updating this! To clarify, is this wrong?

```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
			reinterpret_cast<const char*>(error.getMessage().data()),
			error.getMessage().size()
		)
```

And is this right?
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
		reinterpret_cast<const char*>(error.getMessage().data()),
		error.getMessage().size()
		)
```

I believe `std::string(...` by itself will format into

```
std::string(
	reinterpret_cast<const char*>(error.getMessage().data()),
	error.getMessage().size()
)
```

Would this mean the top (indented) block is correct?

ltomovic-ibl:
The other way around. This is correct:
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
			reinterpret_cast<const char*>(error.getMessage().data()),
			error.getMessage().size()
		)
```
And this is wrong:
```
	logError("Exception occurred while checking other exception %1\n%2")
		<< std::string(
		reinterpret_cast<const char*>(error.getMessage().data()),
		error.getMessage().size()
		)
```
And this wrong piece of code is what actually happens after #4418

This is of course correct as you said:
```
std::string(
	reinterpret_cast<const char*>(error.getMessage().data()),
	error.getMessage().size()
)
```

micheleCTDEAdmin (approved):
It looks good

PR #4452, opened 2025-01-30
correct the test data

guy-maurel:
when cannot be used for a variable name.
Please to any C# programer: have a look to the changes. Thanks.

### DIFF
--- a/src/tokenizer/combine.cpp
+++ b/src/tokenizer/combine.cpp
@@ -610,14 +610,14 @@ void do_symbol_check(Chunk *prev, Chunk *pc, Chunk *next)
          return;
       }
 
-      if (  language_is_set(lang_flag_e::LANG_CS)
-         && pc->Is(CT_WHEN)
-         && pc->GetNext()->IsNotNullChunk()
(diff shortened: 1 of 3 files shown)


src/tokenizer/combine.cpp
```diff
@@ -610,14 +610,14 @@ void do_symbol_check(Chunk *prev, Chunk *pc, Chunk *next)
...
 
-      if (  language_is_set(lang_flag_e::LANG_CS)
-         && pc->Is(CT_WHEN)
-         && pc->GetNext()->IsNotNullChunk()
-         && pc->GetNext()->IsNot(CT_SPAREN_OPEN))
-      {
-         pc->SetType(CT_WORD);
-         return;
-      }
+      //if (  language_is_set(lang_flag_e::LANG_CS)
```
micheleCTDEAdmin, line 613: If this code is not used, we can simply remove it instead of commenting it out.

micheleCTDEAdmin: With the TODO, it looks good.

micheleCTDEAdmin (approved):
It looks good.

Issue #4459, opened 2025-02-15
create an option


guy-maurel:
for the cast 
(   *a_new_type    )

PR #4460, opened 2025-02-15
remove the code and

guy-maurel:
TODO: create a new option for the cast:
(   *a_new_type    )
see #4459


micheleCTDEAdmin (approved):
It makes sense to add an option for that.

PR #4115, opened 2023-10-06
add handling for templated lambdas and lambda specifiers

rrsettgast:
When formatting templated lambdas and/or lambdas with specifiers the formatting would be "incorrect" to my eyes. For instance:

```
namespace ns1
{
namespace ns2
{
namespace ns3
{
namespace ns4
{

  template< int BF_INDEX >
  constexpr static REAL_TYPE value( REAL_TYPE const & coord )
  {
    return executeSequence< numSupportPoints >( [&] <int ... a > () constexpr
        {
          return ( valueFactor< BF_INDEX, a >( coord ) * ... );
        } );
  }

}
}
}
}
```

would lead to:
```
namespace ns1
{
namespace ns2
{
namespace ns3
{
namespace ns4
{

template< int BF_INDEX >
constexpr static REAL_TYPE value( REAL_TYPE const & coord )
{
  return executeSequence< numSupportPoints >( [&] <int ... a > () constexpr
          {
            return ( valueFactor< BF_INDEX, a >( coord ) * ... );
          } );
}

}
}
}
}
```

which is the same error I tried to fix for execution contexts in:
https://github.com/uncrustify/uncrustify/pull/2716

The attempt is to recognize the template and specifiers as valid in the context of the lambda.
The result is this:
```
namespace ns1
{
namespace ns2
{
namespace ns3
{
namespace ns4
{

template< int BF_INDEX >
constexpr static REAL_TYPE value( REAL_TYPE const & coord )
{
  return executeSequence< numSupportPoints >( [&]< int ... a > () constexpr
  {
    return ( valueFactor< BF_INDEX, a >( coord ) * ... );
  } );
}

}
}
}
}
```

As of the initial writing, I couldn't figure how to "properly" account for the new items, so I hacked it in a bit. Please let me know if you have suggestions to account for the new items in a manner more in line with the rest of the code.

for instance, additional `<>` can exist inside of the outer `<>` and the hard coded manner in which I set the chunks will break.

PR #4466, opened 2025-02-22
Support templated lambda

lceWolf:
This MR is pickung up the work from #4115.
Additionally to the original MR this one also allows a template specifier to appear after `CT_TSQUARE`.

The support for additional keywords besides `mutable` is expanded as well to support `noexcept(...)`.

Further, I added a test for the new template parsing.

Lastly, existing tests are adapted. Before this change, a lambda containing any keyword other than `mutable` was not formatted correctly (this was part of the expected output). After this change, the lambda is formatted correctly and the expected output had to be changed accordingly

### DIFF
--- a/src/tokenizer/check_template.cpp
+++ b/src/tokenizer/check_template.cpp
@@ -31,11 +31,14 @@ bool invalid_open_angle_template(Chunk *prev)
       return(false);
    }
    // A template requires a word/type right before the open angle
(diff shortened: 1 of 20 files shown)


micheleCTDEAdmin (approved):
It looks good

lceWolf:
@micheleCTDEAdmin I think you can close PR #4115

micheleCTDE:
Great, will do

PR #4467, opened 2025-02-22
allow arithmetic symbols to appear before lambda

lceWolf:
An immediately invoked lambda can be used in arithmetic expressions. Also preceding a lambda with an unary plus operator triggers an conversion to a function pointer for non-capturing lambdas. Hence both CT_ARITH and CT_POS should be allowed to appear in front of a lambda


micheleCTDE (approved):
Looks good.

PR #4474, opened 2025-03-09
Update kr-indent.cfg

MauMendes:
I am not sure if you are accepting PR for this cfg files. and I feel that with this small change the code looks better.

Comparison bellow with the standard kr-indent.cfg and after updates

standard:
```c
/* Sample.c */
#include <stdio.h>
int main(void)
{
	int n = 9; printf("Multiplication table for %d:\n", n); for (int i = 1; i <= 10; i++) {
		printf("%d x %d = %d\n", n, i, n * i);
	}
	return 0;
}
```

after:
```c
/* Sample.c */
#include <stdio.h>
int main(void)
{
	int n = 9;
	printf("Multiplication table for %d:\n", n);

	for (int i = 1; i <= 10; i++) {
		printf("%d x %d = %d\n", n, i, n * i);
	}
	return 0;
}
```

### DIFF
--- a/etc/kr-indent.cfg
+++ b/etc/kr-indent.cfg
@@ -27,8 +27,8 @@ nl_fdef_brace			= add		# "int foo() {" vs "int foo()\n{"
 nl_brace_while			= remove
 nl_brace_else			= remove
 nl_squeeze_ifdef		= TRUE
-nl_after_semicolon  = TRUE
-nl_before_for  = TRUE
+nl_after_semicolon		= TRUE
+nl_before_for			= TRUE
 
 # mod_paren_on_return		= add		# "return 1;" vs "return (1);"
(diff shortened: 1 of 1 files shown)


etc/kr-indent.cfg
```diff
@@ -27,6 +27,8 @@ nl_fdef_brace			= add		# "int foo() {" vs "int foo()\n{"
 nl_brace_while			= remove
 nl_brace_else			= remove
 nl_squeeze_ifdef		= TRUE
+nl_after_semicolon  = TRUE
```
gmaurel, line 30: please align at the = sign

gmaurel (changes requested):

gmaurel (approved):

PR #4481, opened 2025-03-16
the code line will never been reached

guy-maurel:


src/language_names.cpp
```diff
@@ -71,7 +71,7 @@ size_t language_flags_from_name(const char *name)
       }
    }
 
-   return(0);
+   return(0); // will never been reached
```
micheleCTDEAdmin, line 74: It could be reached if we call the function with an invalid/incorrect name. Not sure we should add this comment., although it should never be reached when things are right.

gmaurel (approved):

PR #4487, opened 2025-04-02
provide more test data for the options pos_xyz

guy-maurel:


tests/input/cpp/pos_assign.cpp
```diff
@@ -1,7 +1,19 @@
-static const unsigned char radiooff_light_bits[] =
+
+void foo()
 {
-	0x00, 0x00, 0x00, 0x00, 0x00, 0x08, 0x00, 0x08, 0x00, 0x10, 0x00, 0x10,
```
micheleCTDEAdmin, line 3: Why removing this part of the test? it may be useful to check this kind of code.

gmaurel (approved):

Issue #4346, opened 2024-06-27
Constructor colon issue with indent_namespace_single_indent option


erkanvatan:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"
Hello,
While I was experimenting  with the `indent_namespace_single_indent` option, I found out a bug that causes class constructor inside namespace to be seen as regular function. In the debug file it says `CLASS_COLON` instead of `CONSTR_COLON` for the colon of constructor and my constructor settings do not apply. 

- source file: [issue_constructor_colon_src.txt](https://github.com/user-attachments/files/16021497/issue_constructor_colon_src.txt)
- generated result: [issue_constructor_colon_out.txt](https://github.com/user-attachments/files/16021498/issue_constructor_colon_out.txt)
- expected result: [issue_constructor_colon_exp.txt](https://github.com/user-attachments/files/16021499/issue_constructor_colon_exp.txt)
- debug file: [issue_constructor_colon_dbg.txt](https://github.com/user-attachments/files/16021500/issue_constructor_colon_dbg.txt)

Uncrustify version: 0.79.0_f

PR #4492, opened 2025-04-04
Fix constructors not being detected as such in nested namespaces if indent_namespace_single_indent = true

marco-stephan:
Fixes https://github.com/uncrustify/uncrustify/issues/4346

Previously, for a constructor to be detected as such one of the conditions was that the brace level of the namespace plus 1 had to be equal to the level of the constructor definition (note brace level vs. level). Normally the brace level and the level of such chunks are equal so it does not matter that you compare these two different types of levels. For nested namespaces and if `indent_namespace` and `indent_namespace_single_indent` are both set to true, this does not hold up anymore because of https://github.com/uncrustify/uncrustify/commit/896c25688a24b14e30eaaa11f3635ef8105c66c3#diff-e040a4103d2245520694a6804cb2e3de5f1f00d5c78a8bd9f1167eb8394f0290R625.
As a result, the "brace level" is not really the brace level anymore but acts more as an indention indicator.

To fix the behavior, the comparison now uses the brace level for both chunks.

### DIFF
--- a/src/tokenizer/EnumStructUnionParser.cpp
+++ b/src/tokenizer/EnumStructUnionParser.cpp
@@ -1600,16 +1600,16 @@ void EnumStructUnionParser::mark_constructors()
       /**
        * get the name of the type
        */
-      auto *body_end   = get_body_end();
-      auto *body_start = get_body_start();
-      auto *name       = m_type->Text();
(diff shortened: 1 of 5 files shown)


gmaurel (approved):
Fine, thanks

Issue #4438, opened 2024-12-12
Spacing is not correct for references to pointers(*&)


MaximKolesnik:
[//]: # "  If the issue is connected to existing Uncrustify options please, if possible, add the"
[//]: # "  following information to ease up the process:"
[//]: # "  • a link to a debug file:"
[//]: # "      generated with: 'uncrustify -p debug.txt -c pathToUsedConfig.cfg -f toBeFormatedFile.cpp' "
[//]: # "      Example: [debug.txt](https://linkToTheFile)"
[//]: # "      Example hosters for debug files: pastebin.com, gist.github.com, ..."
[//]: # "      The used config file is included in the debug file and does not need to be included here."
[//]: # 
[//]: # "  • include a small but complete test file that will be uncrustifyed"
[//]: # "  • include the generated results"
[//]: # "  • include the expected results"
[//]: # 
[//]: # "  ✋ please add a line containing ``` above and below of each of those three code sections"
[//]: # 
[//]: # "  • include the current version of your Uncrustify executable"
[//]: # "      printout via 'uncrustify -v'"
[//]: # "      Example: current version: uncrustify 0.63"
[//]: # "      or if possible additionally with the git sha of the commit"
[//]: # "                    current version: uncrustify 0.63 dc7b412"
[//]: # 
[//]: # "  • if possible include a version that worked"
[//]: # "      Example: working version: uncrustify 0.63"
[//]: # "      or"
[//]: # "                    working version: uncrustify 0.63 2a5e88f"

Hello, seems like formatter does not handle reference to pointer correctly
Binary version 0.80.1_f

Expected:
```
void funcPtr(int *param);
void funcRef(int &param);
void funcPtrRef(int *&param);
```
Result:
```
void funcPtr(int *param);
void funcRef(int &param);
void funcPtrRef(int*&param);
```
[debug.txt](https://github.com/user-attachments/files/18116452/debug.txt)

PR #4493, opened 2025-04-04
Fix spacings for pointer reference parameters

marco-stephan:
Fixes two bugs:
1. Spacing before pointer characters for named pointer reference parameter using `unnamed` option (fixes https://github.com/uncrustify/uncrustify/issues/4438)
2. Spacing between pointer characters and reference character for unnamed reference parameter using `sp_before_unnamed_byref` option


gmaurel (approved):
thanks

PR #4494, opened 2025-04-04
Qual: Add pre-commit configuration & ci

mdeweerd:
# Qual: Add pre-commit configuration & ci

Add initial pre-commit configuration also run in ci.


guy-maurel:
Well, first thanks for the PR.
I am the maintainer of uncrustify.
Avez vous une adresse en France?

These PR are too big for us to accept.
I prefer some little step.
Let us try at new!

Some of the files you have tested are
too old or never used by me before,
some have been included by error.
I/we need to remove some of them before.

Thanks, à bientôt.

mdeweerd:
Oui, j'ai une adresse en France.

On peut désactiver des hooks (=hook-stage manual) et activer 1 par 1 chacun avec ses résulutions.
Ou exclure des fichiers/répertoires...

Je peux écrire sur votre adresse (mél. git).

mdeweerd:
I think I understand what was meant now: this PR included some changes of the other PRs to check which changes were still recommended by the tools.

This PR is now reduced to the pre-commit configuration.

A limited list of the notices from the tools can be checked in the files tab.  You need to go to the end of the contents to see some:
https://github.com/uncrustify/uncrustify/pull/4494/files

mdeweerd:
The cppcheck and cpplint messages are now down to the next few.

I'ld prefer feedback on the preferred approches for the following:

```plaintext
# Probably fixed by passing a const reference.
src/option.h(195): performance (passedByValue): Function parameter 'val' should be passed by const reference.

# Due to "complex" macro, not clear if this should be ignored or if a fix is possible.
# Ok to add `NOLINT()` or `cppcheck suppress` directives?
src/pcf_flags.h(85): error (returnDanglingLifetime): Returning object that points to local variable 'f2' that will be invalid when returning.

# Instead of using a "const" for the length of the buffer we could use a '#define ..' and assign that to the const.?
src/uncrustify.cpp:704:  Do not use variable-length arrays.  Use an appropriately named ('k' followed by CamelCase) compile-time constant for the size.  [runtime/arrays] [1]
src/uncrustify.cpp:834:  Do not use variable-length arrays.  Use an appropriately named ('k' followed by CamelCase) compile-time constant for the size.  [runtime/arrays] [1]

# Due to the '#if .. else' around the 'if'.  Personnally I fix this with
#    if(
#    #if MACRO_CONDITION
#       <if_condition1>
#    #else
#       <if_condition2>
#    #endif
#    ) {
#      <if_body>
#    }
src/uncrustify.cpp:1663:  If/else bodies with multiple statements require braces  [readability/braces] [4]
src/uncrustify.cpp:1665:  If/else bodies with multiple statements require braces  [readability/braces] [4]
```

I can split the other changes to the code into several PR's or the commits of this PR can be navigated in the github interface.

Once everything is fixed (or ignored), the pre-commit flow can be merged.

mdeweerd:
@guy-maurel A quick hint on how to run just that test?  I ran tests locally, but apparently there are many more to run.

What about the remaining notices from the lint tools?

mdeweerd:
I reorganized the commit to trigger anotherappveyor run.  If still in error, then gcc's snprintf and MSVC's v19's snprintf are different.

I must admit that I find the ElidedText function a bit odd.
It either produces strings of length <= truncate_value and also string of length `truncate_value + 30`.

- Suppose "test_it" is 41 characters long, with a truncate value of 40.
  Then `test_it_length <  40 + 30` and we truncate test_it to 40 - 30 = 10 characters.  And we append '... <the string is truncated>', ending up with 40 characters and and EOS token.

- Now suppose "test_it" is 80 characters long, with a truncate value of 40.
  Then `test_it_length >  40 + 30` and we truncate test_it to 30  characters.  And we append '... <the string is truncated>', ending up with 60 characters and and EOS token.

IMHO, in these cases the code should always take 10 characters from the source string, and append the ' ... <>' string, always limiting to 40 characters.


It is possible that that this code intended to manage another corner case: when 'truncate_value <30' in which case the string we append would not fit in the maximum line size.

mdeweerd:
Thes are the remaining notices with "Cppcheck 2.17.1"

```log
src\log_rules.cpp(72): warning (nullPointerOutOfMemory): If memory allocation fails, then there is a possible null pointer dereference: r
src\log_rules.cpp(99): warning (nullPointerOutOfMemory): If memory allocation fails, then there is a possible null pointer dereference: r
src\log_rules.cpp(126): warning (nullPointerOutOfMemory): If memory allocation fails, then there is a possible null pointer dereference: r
src\keywords.cpp(217): error (uninitvar): Uninitialized variables: &key.type, &key.lang_flags
```

The notice
```log
src/pcf_flags.h(85): error (returnDanglingLifetime): Returning object that points to local variable 'f2' that will be invalid when returning.
```
is gone with that cppcheck version (I fixed it) and seems to be a false positive at this point in the github runner probably using a different cppcheck version.

guy-maurel:
prends contact avec moi: guy.j@maurel.de je ne comprends pas tout ce que tu as écrit. Merci.

mdeweerd:
Some true issues:

Two tests on `if (pc->Is(CT_VBRACE_OPEN))` with two different actions, the second one never done.
Two examples, I am not fixing them:

https://github.com/uncrustify/uncrustify/blob/57a57677557eaf357099d2a539090889584ce178/src/unc_tools.cpp#L97-L111

https://github.com/uncrustify/uncrustify/blob/57a57677557eaf357099d2a539090889584ce178/src/unc_tools.cpp#L347-L361

The 'IsNotNull' test in the else part is clearly false because it is the same condition as for the 'if' part - not fixing (and reported by cppcheck):
https://github.com/uncrustify/uncrustify/blob/57a57677557eaf357099d2a539090889584ce178/src/tokenizer/combine_fix_mark.cpp#L1405-L1421

mdeweerd:
I am labeling this "Ready for review":

- The remaining notices from cppcheck are better fixed by a maintainer;
- `cppcheck` is locally built - with cache - avoiding false positives that are reported in the default version of ubuntu-latest.

Currently visible notices from cppcheck:

```
src/uncrustify.cpp(936): style (knownConditionTrueFalse): Condition 'error=redir_stdout(output_file)' is always false
src/uncrustify.cpp(948): style (knownConditionTrueFalse): Condition 'error=redir_stdout(output_file)' is always false
src/uncrustify.cpp(1027): style (knownConditionTrueFalse): Condition 'error=redir_stdout(output_file)' is always false
src/uncrustify.cpp(1858): warning (nullPointerRedundantCheck): Either the condition 'ref=ref->GetPrev()' is redundant or there is possible null pointer dereference: ref.
src/tokenizer/combine_fix_mark.cpp(1420): style (knownConditionTrueFalse): Condition 'tmp2->IsNotNullChunk()' is always false
src/unc_tools.cpp(109): style (multiCondition): Expression is always false because 'else if' condition matches previous condition at line 97.
src/unc_tools.cpp(361): style (multiCondition): Expression is always false because 'else if' condition matches previous condition at line 347.
src/unc_tools.cpp(446): style (multiCondition): Expression is always false because 'else if' condition matches previous condition at line 432.
src/tokenizer/combine_labels.cpp(285): style (knownConditionTrueFalse): Condition 'tmp->IsNullChunk()' is always false
```

EDIT: I updated the list after fixing some more - apparently the number of notifications shown is limited and as more are fixed the analysis takes longer so I'll probably add a caching mechanism for cppcheck.

mdeweerd:
Creating/Saving and Restoring cppcheck cache was succesful: gained about 20 minutes, pre-commit workflow finished in 2m 22s .

guy-maurel:
fine work, thanks

micheleCTDE:
Can we refactor this PR? there are way too many commits and several of them are for similar things, so they could be grouped together.
It would als probably be better to split this PR in smaller chunks to make review and discussion easier.

mdeweerd:
@guy-maurel 
The code changes in this PR are now split in multiple smaller PRs.
I will rebase this PR on the master branch in order to ensure that this PR is limited to the flow additions.

mdeweerd:
Almost only the update to the flow is remaining here and the following notices.

Only the first notice is reported as an error.  It is true that `ref=ref->GetPrev()` is never false because if it can be false, then the next if condition uses a null pointer.


```
src/uncrustify.cpp(1862): warning (nullPointerRedundantCheck): Either the condition 'ref=ref->GetPrev()' is redundant or there is possible null pointer dereference: ref.
src/tokenizer/combine_fix_mark.cpp(1420): style (knownConditionTrueFalse): Condition 'tmp2->IsNotNullChunk()' is always false
src/align/var_def_brace.cpp(129): style (knownConditionTrueFalse): Condition 'pc->IsPreproc()' is always false
src/tokenizer/combine_labels.cpp(285): style (knownConditionTrueFalse): Condition 'tmp->IsNullChunk()' is always false
```

gmaurel (changes requested):
are we ready with this PR ?

mdeweerd:
This PR is ok and ignores some notices that are enabled in #4579 .
So this one can be integrated and later #4579 when the remaining notices are somehow decided on.

gmaurel (approved):
we try it

PR #4495, opened 2025-04-05
Qual: Fix spelling

mdeweerd:
# Qual fix spelling

Fix spelling suggested by codespell.


commit.log
```diff
@@ -2126,7 +2126,7 @@ Date:   Mon Oct 28 16:26:50 2013 -0700
     Differentiate between normal ternary (a ? b : c) and short ternary, (b ?: c) with option sp_cond_ternary_short
 
 commit 4b83d06c86ab949d1bc71d8eaa2cee9e9c7332bb
-Author: Tim Hütz <tim@huetz.biz>
+Author: Time Hütz <time@huetz.biz>
```
micheleCTDE, line 2129: I think this fix is incorrect.

commit.log
```diff
@@ -7657,7 +7657,7 @@ commit 23632dae52e5a30a8ff595d9bd9d6c6d2eede07a
 Author: Ben Gardner <gardner.ben@gmail.com>
 Date:   Sat Oct 17 14:02:39 2009 -0500
 
-    Spelling fix: preceeded -> preceded
+    Spelling fix: prece\eded -> preceded
```
micheleCTDE, line 7660: This seems wrong.

micheleCTDE (changes requested):
Some adjustments required

commit.log
mdeweerd: If I leave preceeded then codespell is reporting it, so I inserted this '\' on purpose to avoid the match.
The other way is to add a separate file with lines to ignore.

commit.log
mdeweerd: Changed.

commit.log
micheleCTDE: The line in the log is clearly reporting a fix of the incorrect word into the correct word, it makes no sense to change it to avoid codespell to report it, it would invalidate the purpose of the log message in first place.

commit.log
micheleCTDE: The email is now ok but the same still says `Time Hütz`, which is wrong

mdeweerd: Sorry, corrected that as well.

commit.log
mdeweerd: OK, I'll revert and add a file for the spelling exceptions.  Preceeded was still misspelled quite often so I'ld avoid it as a word to ignore in the spelling check.

micheleCTDE: You are right, "preceeded" is a common mistake. Is there a way to exclude speficic entries in specific files instead of excluding the complete word? Otherwise an option could be for "preceeded" to still be checked when you run codespell and simply ignore that specific change in commit.log (perhaps you can a script to locally modify commit.log prior to spell check and put it back after the spell check)

micheleCTDE (approved):

PR #4496, opened 2025-04-05
Fix shellcheck notices

mdeweerd:
# Fix shellcheck notices

This fixes notices from shellcheck - there was one bug, a redirect that was on a line by itself.

### DIFF
--- a/coverity.sh
+++ b/coverity.sh
@@ -6,7 +6,7 @@
 BRANCH=coverity_scan
 
 if [ -z "$1" ] ; then
-	cat <<EOF
+  cat <<EOF
 Usage: $0 REMOTE
 
 This script triggers a coverity build by pushing the current code to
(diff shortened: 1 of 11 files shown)


micheleCTDE (commented):
I would suggest indentation to e left unchanged in this PR. Different people have different preferences, so @guy-maurel (main maintainer) should be the one making a call. Without a strong reason for it, there is no need to change from 4 to 2 spaces.

mdeweerd:
The principle is the same as uncrustify: apply the same formatting to the entire project.

The number of spaces for indenting is an option and can also be set to tab.

Automatic indenting sometimes help identify the statements that are (not) under an if clause (or the opposite).

If the decision is to not format the shell files, then beautysh needs to be disabled/removed.

gmaurel (changes requested):
some of the files are not used any more.
I will remove after the check.
Please, wait a little bit.

gmaurel (approved):

PR #4497, opened 2025-04-05
Fix: Repeated author key in package.json, add contributors

mdeweerd:
Repeated keys are not allowed in json objects.  The contributors key is more appropriate in this case

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,12 +15,23 @@
     "uncrustify",
     "source code beautifier"
   ],
+  "author": "Ben Gardner",
   "contributors": [
     {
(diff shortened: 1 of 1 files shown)


micheleCTDE (commented):
Ben Gardner is the original author of Uncrustify, so using the `author` field seems appropriate. Then in `contributors` we could list the current code owners (4 people)

mdeweerd:
I am assuming that package.json is npm compatible.

micheleCTDEAdmin (approved):
Looks good.

PR #4499, opened 2025-04-05
Doc: Update code type suggestions in README

mdeweerd:
Some replacements/additions with 'bash'/'console' code formatting hint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,10 @@ Pre compiled binaries for Windows can be downloaded from the [Releases](https://
 To generate a build system for Uncrustify using CMake on UNIX-like systems, create a
 build folder and run CMake from it, making sure to specify Release mode:
 
-```console
-$ mkdir build
-$ cd build
-$ cmake -DCMAKE_BUILD_TYPE=Release ..
+```bash
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=Release ..
 ```
 Other systems may require other flags (e.g. `cmake -G Xcode ..` for Xcode).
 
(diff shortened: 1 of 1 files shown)


micheleCTDE (commented):
In some places we are replacing `bash` with `console` and in others `txt` with `bash`. Seems a bit counter intuitive. Why not only `bash` or only `console`?

mdeweerd:
This follows the original way this was documented.

I used console where the documentation was showing '$ command' as that can only be in a console,  and bash where it is likely a bash command.  'console' is for the former, 'bash' can be used for the latter.

micheleCTDEAdmin:
For example, line 225 uses `bash` to run `uncrustify -c ...`, line 313 uses `console` ` to run `uncrustify -c ...`, which is the exact same command (different arguments only).
I think we can just use the same for all, either `console` or `bash`

mdeweerd:
@micheleCTDE 

We can do the same everywhere if we either have the `$` before the commands everywhere (`console`)or no `$` everywhere (`bash`).

The lines you referred to have different prefixes.

I did not modify the content itself because I do not know the reason why these prefixes are different.

Personally I would not add the `$` because that favors the copy/paste feature.
Adding the `$` is more useful to differentiate between the command and its output, like this:

```console
$ uncrustify -c uncrustify.cfg simcom_messages.c
Output suffix: .uncrustify
do_source_file: Parsing: simcom_messages.c as language C
```

These were the lines (225 and 313) that were referred to:

https://github.com/uncrustify/uncrustify/pull/4499/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R225-R227

https://github.com/uncrustify/uncrustify/pull/4499/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L313-L316

micheleCTDE:
> I did not modify the content itself because I do not know the reason why these prefixes are different.

Being the author of the commit that added those `.txt` sections, I can say for sure that it was a mistake on my side. It should have been a console/bash block and should have added a '$' sign, to be conform to the rest of the README file.

mdeweerd:
I just committed the opposite option: removing all the leading `$` tokens (which improves the copy/paste experience as the copied text is executabe on the CLI when pasted).

micheleCTDE:
It's fine, I agree removing '$' makes copy-n-paste easier. One final note: lines 226 and 233 are using an indentation in the command that other blocks don't have. This is also visible when the page is rendered in hmtl. Would be good to remove those spaces so bash blocks look consistent.

mdeweerd:
I've update the indendations with regards to the location of the backticks.

micheleCTDEAdmin (approved):
Now it looks good

PR #4501, opened 2025-04-06
correct the description of the option nl_comment_func_def

guy-maurel:
provide more test data


micheleCTDEAdmin (changes requested):
Comment and functionality don't match, they are off by 1
If I use 1, in `/* comment */\n <here> void foo() {...}` I would expect an additional new line resulting in an empty line between comment and function.
The correct comment should be `/* comment */ <here> void foo() {...}`, without `\n`. Then description and functinoality would match

gmaurel:
Thanks for the comment

micheleCTDEAdmin (approved):
Now it looks good

PR #4509, opened 2025-04-26
Qual: Fix executable bits, spelling, whitespace

mdeweerd:
This fixes some more notices in relation with setting up checks run in pre-commit #4494 .


micheleCTDEAdmin (approved):
It looks good

Issue #4502, opened 2025-04-07
`&` in bitwise AND against scoped enum variant treated as byref


briggsywxf:
Encountering the following issue in v0.80.1

In the minimal example:

```
enum MyEnum
{
	Null = 0,
	FlagA = 1,
	FlagB = 2,
};

int main()
{
	MyEnum flags = MyEnum::Null;

	if ((flags & MyEnum::FlagA) != MyEnum::Null)
	{
		return 1;
	}

	return 0;
}
```

With cfg
```
sp_before_byref = remove
sp_after_byref = force
sp_arith = force
sp_bool = force
```

The line

`if ((flags & MyEnum::FlagA) != MyEnum::Null)`

Is formatted as

`if ((flags& MyEnum::FlagA) != MyEnum::Null)`

If I were to add subsequent bitwise AND's within the parentheses, they are also aligned incorrectly

I can see in the dump output that the parentheses are mistakenly parsed as a c-style cast, and `flags` as a type, hence the ampersand being considered a 'byref' and thus moved.

```
#  12>                 IF|               NONE|     PARENT_NOT_SET[  9/  9/ 11/  0][1/1/0][0-1]         if
#  12>        SPAREN_OPEN|                 IF|     PARENT_NOT_SET[ 12/ 12/ 13/  1][1/1/0][0-0]            (
#  12>         PAREN_OPEN|             C_CAST|     PARENT_NOT_SET[ 13/ 13/ 14/  0][1/2/0][0-0]             (
#  12>               TYPE|             C_CAST|     PARENT_NOT_SET[ 14/ 14/ 19/  0][1/3/0][0-0]              flags
#  12>              BYREF|             C_CAST|     PARENT_NOT_SET[ 20/ 20/ 21/  1][1/3/0][0-0]                    &
#  12>               TYPE|             C_CAST|     PARENT_NOT_SET[ 22/ 22/ 28/  1][1/3/0][0-0]                      MyEnum
#  12>          DC_MEMBER|             C_CAST|     PARENT_NOT_SET[ 28/ 28/ 30/  0][1/3/0][0-0]                            ::
#  12>               TYPE|             C_CAST|     PARENT_NOT_SET[ 30/ 30/ 35/  0][1/3/0][0-0]                              FlagA
#  12>        PAREN_CLOSE|             C_CAST|     PARENT_NOT_SET[ 35/ 35/ 36/  0][1/2/0][0-0]                                   )
[...]

PR #4510, opened 2025-04-28
introduce CT_BITWISE_AND as CT_ARITH

guy-maurel:


briggsywxf:
Unfortunately this seems to not fix the issue in #4502 . The line in question is still formatted as `if ((flags& MyEnum::FlagA) != MyEnum::Null)` .
The parsing dump output also shows that the `&` is still parsed as a byref inside a c_cast, so it doesn't seem to be an issue with the config. Perhaps due to `flags` being parsed as `TYPE` even when it isn't and then we hit the assignment at line 1337 in these changes so the `&` is still set to `CT_ADDR`.

Is it worth adding the example given in #4502 as a test case? I can write it up properly as an additional test if needed.

guy-maurel:
I look...
Some more test data is always welcome!
Thanks

briggsywxf:
> I look... Some more test data is always welcome! Thanks

I'll add a test and open it as a PR into this one

maxbachmann:
Any chance t get this fixes? Or is there a deeper problem?

PR #4511, opened 2025-04-28
ci: parallelize build and test scripts

mdeweerd:
- Add commands to get the number of CPU cores available
- Update build and test scripts to use parallel execution based on the number of cores


mdeweerd:
I can't see any speedup so I guess that Appveyor provides only 1 core in this run.
Anyway, it might help sometimes.

PR #4512, opened 2025-04-28
ci: update GitHub Actions workflow for uncrustify tests

mdeweerd:
- Add additional CMake flags for sanitizers (disabled)
- Enable parallel build and test execution
- Update actions/checkout to v4

I have tested the test execution with the sanitizers - it was considerably slower so that's something that you may only want to do for a release/nightly/... .  No issue was found with the selected sanitizers active.

Regarding the parallel execution: no considerable speedup was noted.


micheleCTDEAdmin (approved):
It looks good. Hopefully it will also speed up building thanks for parallel option.

PR #4513, opened 2025-04-30
Add test for enum arithmetic in PR #4510 / Issue #4502

briggsywxf:
This is a test for https://github.com/uncrustify/uncrustify/pull/4510 and as such is blocked on it.

This is currently expected to fail on master, but should pass once #4510 is correct.

The final three `if` statements format correctly thanks to the extra parentheses. The first 3 currently do not due to the reasons mentioned in #4502 and #4510


gmaurel:
The test data is too old.
Please, make a little change to get it starts again.

gmaurel:
@briggsywxf 
Please have a look to my comment and make a change!

briggsywxf:
@gmaurel I've recently returned from holiday and will revisit this soon.
Thanks

gmaurel:
fine... I wait!

briggsywxf:
Hi @gmaurel - I've updated this branch onto the latest `master` revision - though I don't see any changes that needed to be made beyond this. Were you referring to something else before? It's expected that the test I've added would fail until the fix in #4510 is correct. 
Thanks for taking another look :)

gmaurel:
All this are very old.
Can you please close the call and create a new Issue.

Issue #4515, opened 2025-05-10
bug and workaround fix: bad size_t assignement


c2main:
Reduced case:

```C
void recursive_bug(){if (foo())
	bar();
}
```

```
# Uncrustify-0.80.1-146-bb68ec18b
eat_blanks_after_open_brace     = true
eat_blanks_before_close_brace   = true
nl_remove_extra_newlines        = 2
# option(s) with 'not default' value: 118
#
```

Lead to an abnormal value in `GetColumn()`.

This patch fix *this* bug but is probably not the right fix, I am facing some similar behavior with another case which I didn't reduced to minimal yet but I suspect is similar:

```
From 62acd653ed1aee03234ca6cf364c383e7d9e231c Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?C=C3=A9dric=20Villemain?= <Cedric.Villemain@Data-Bene.io>
Date: Sat, 10 May 2025 09:23:40 +0200
Subject: [PATCH] Fix negative size_t

In reindent, we call SetColum(), using max(), but the size_t used for
one of type is built from ints and be can be negative.
Naive test: if negative, set value to 0
---
 src/reindent_line.cpp | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)

diff --git a/src/reindent_line.cpp b/src/reindent_line.cpp
index 5d9b5750b..2cf7faed6 100644
--- a/src/reindent_line.cpp
+++ b/src/reindent_line.cpp
@@ -102,7 +102,10 @@ void reindent_line(Chunk *pc, size_t column)
       }
       else
       {
-         pc->SetColumn(max(pc->GetColumn() + col_delta, min_col));
+         if (static_cast<int>(pc->GetColumn() + col_delta) < 0)
+            pc->SetColumn(min_col);
+         else
+            pc->SetColumn(max(pc->GetColumn() + col_delta, min_col));
 
          LOG_FMT(LINDLINED, "%s(%d): set column of ", __func__, __LINE__);
 
-- 
2.47.2
```

micheleCTDEAdmin:
> Lead to an abnormal value in GetColumn().

I didn't do a test/analysis yet, but I am curious about the abnormal value returned by `GetColumn()`. What do you get?

c2main:
I have reduced the other case: infinite loop due to line length, "abcdef" makes the line 79 chars.

```
static void infinite_loop(){for (int i = 0; i < 10; i++){if (abcdef) continue;
[...]

PR #4521, opened 2025-05-28
bug fix for #4515

guy-maurel:
Thanks to @c2main who solve the bug #4515

### DIFF
--- a/src/reindent_line.cpp
+++ b/src/reindent_line.cpp
@@ -102,8 +102,14 @@ void reindent_line(Chunk *pc, size_t column)
       }
       else
       {
-         pc->SetColumn(max(pc->GetColumn() + col_delta, min_col));
-
+         if (static_cast<int>(pc->GetColumn() + col_delta) < 0)            // Issue #4515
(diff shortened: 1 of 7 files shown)


c2main:
I have spotted some 6 other places with similar risk, and 1 where it's one "on purpose" to get the max value as a result.

Maybe it's best to merge this one, and revisit with a more global approach...

micheleCTDEAdmin:
Based on comment on issue #4515, we should try to fix the abnormal value of `GetColumn()` rather than merging a fix that we already suspect is not right in first place, based on comments from @c2main in #4515.

c2main:
> We get the value `18446744073709551612` for the `bar` column, which is `-4` when converted to `size_t`. This PR is definitely not correct. We should instead find out why we get the wrong column and fix that instead.

FYI I have enough for my case to fix all, though there is something pending. I think that some code path is/are missing a `SetNlCount()`, thus `col_delta` is not reset to 0, leading to unexpected delta greatest than position.

PR #4523, opened 2025-05-30
Use MARK_CHANGE macro instead of changes++

c2main:
There were still few places not using the macro but directly updating the counter, assuming those were just missed.


gmaurel (approved):
thanks

Issue #4522, opened 2025-05-30
infinite loop with combination of long line and removing extra new lines fully.


c2main:
I have reduced another case: infinite loop due to line length, "abcdef" makes the line 79 chars. I though it was related the #4515 but ...

```
static void infinite_loop(){for (int i = 0; i < 10; i++){if (abcdef) continue;
                if (a) continue;
        }
}
```

With config:

```
nl_after_semicolon              = true    # true/false
nl_remove_extra_newlines        = 2       # unsigned number
code_width                      = 78      # unsigned number
#debug_max_number_of_loops       = 20        # number
```

I have not found yet the root cause but it's probably somewhere under "code_width", here we probably have an out of order increment of "changes" (after the first loops, when going "inside the second loop"). My dump files from "-ds" are all the same once entering the infinite loop. See an example:

```
STEP: Inside second while loop
--------------
# -=====-
# number of loops               = 21
# -=====-
# language                      = C
# -=====-
# Line                Tag         Parent_type  Type of the parent         Columns Br/Lvl/pp             Flags   Nl  Text
#   1>          QUALIFIER|           FUNC_DEF|     PARENT_NOT_SET[  1/  1/  7/  0][0/0/0][             e 0000][0-0] static
#   1>               TYPE|           FUNC_DEF|     PARENT_NOT_SET[  8/  8/ 12/  1][0/0/0][             2 0000][0-0]        void
#   1>           FUNC_DEF|               NONE|     PARENT_NOT_SET[ 13/ 13/ 26/  1][0/0/0][                  0][0-0]             infinite_loop
#   1>        FPAREN_OPEN|           FUNC_DEF|     PARENT_NOT_SET[ 26/ 26/ 27/  0][0/0/0][        2 0000 0000][0-0]                          (
#   1>       FPAREN_CLOSE|           FUNC_DEF|     PARENT_NOT_SET[ 27/ 27/ 28/  0][0/0/0][        2 0000 0008][0-0]                           )
#   1>         BRACE_OPEN|           FUNC_DEF|     PARENT_NOT_SET[ 28/ 28/ 29/  0][0/0/0][        2 0000 0000][0-0]                            {
#   1>            NEWLINE|               NONE|     PARENT_NOT_SET[ 29/ 29/  0/  0][0/0/0][                  0][1-0]
#   1>                FOR|               NONE|     PARENT_NOT_SET[  9/ 29/ 32/  0][1/1/0][             c 0000][0-0]         for
[...]

PR #4524, opened 2025-05-30
WIP: Revert partially commit 4f60ddd1e and alternative fix for #1349

c2main:
All tests are passing, and it also fixes the infinite loop outlined in #4522

WIP: I will add tests for my infinite loop

### DIFF
--- a/src/uncrustify.cpp
+++ b/src/uncrustify.cpp
@@ -2356,7 +2356,7 @@ void uncrustify_file(const file_mem &fm, FILE *pfout, const char *parsed_file,
       add_long_preprocessor_conditional_block_comment();
    }
    // Align everything else, reindent and break at code_width
-   cpd.pass_count = 3;
+   cpd.pass_count = 1;
    first          = true;
 
    dump_step(dump_file, "Before second while loop");


gmaurel:
I miss the test(s) you write about

gmaurel:
@c2main 
I am waiting for the test you spoke about

c2main:
> @c2main I am waiting for the test you spoke about

hey @gmaurel ! 

It went off the radar.

I have core dump with the following, I had a more reduced case, but cannot find it, so at least:

```
void
get_psi_filepath(const char *cgroup, const char *resource, char *filepath,
				 size_t filepath_size)
{
	/*
	 * check resource is managed, we may not restrict but as we're building a
	 * path it's safer
	 */
	if (strcmp(resource, "cpu") != 0 && strcmp(resource,
											   "io") != 0 && strcmp(resource,
																	"memory")
		!= 0)
		ereport(ERROR,
				(errmsg(
					 "Invalid argument. Expected 'io', 'memory', or 'cpu'.")));
	/* Construct the file path depending on whether a cgroup is provided */
	/* FIXME document this case */
	if (strcmp(cgroup, "proc") == 0)
	{
		snprintf(filepath, filepath_size, "/proc/pressure/%s", resource);
	}
	else
	{
		/*
		 * we need the root cgroup of postmaster, we derive it t get the real
		 * postgres root cgroup
		 */
		PidGetCgroup(PostmasterPid, &cgroup_root_path);
		if (strcmp(cgroup, "") == 0)
			snprintf(filepath, filepath_size, "%s/../../%s.pressure",
					 cgroup_root_path, resource);
		else
			/*  TODO check the cgroup input */
			snprintf(filepath, filepath_size, "%s/../../%s/%s.pressure",
					 cgroup_root_path, cgroup, resource);
	}
}
```

With cfg:

```
# Uncrustify-0.80.1_f
nl_after_semicolon              = true    # true/false
nl_remove_extra_newlines        = 2        # unsigned number
code_width                      = 78        # unsigned number
```


A run: 

```
./build/uncrustify -c bug_uncrustify/bug2.cfg bug_uncrustify/test.c
Output suffix: .uncrustify
do_source_file: Parsing: bug_uncrustify/test.c as language C
CopyAndAdd(645): no column number
CopyAndAdd(645): no column number
CopyAndAdd(645): no column number
uncrustify: /home/cedric/src/uncrustify/src/indent.cpp:243: void align_to_column(Chunk*, size_t): Assertion `column >= pc->GetColumn()' failed.
Abandon (core dumped)
```

guy-maurel:
I take a look

micheleCTDE:
This is possibly similar or related to issue #4548 and #4537 and probably stems from the wrong handling of vbrace column values. We need to tackle that before further addressing this.
Btw I have reduced the test case to
```
void get_psi_filepath()
{
		if (true)
			sprintf(filepath, "%s/../../%s.pressure", cgroup_root_path, resource);
		else
			sprintf(filepath2, "%s/../../%s.pressure", cgroup_root_path2, resource2);
}
```

If you add real braces the problem does not show up, again an indication that the use of vbrace is the likely root cause the problem
```
void get_psi_filepath()
{
		if (true){
			sprintf(filepath, "%s/../../%s.pressure", cgroup_root_path, resource);}
		else{
			sprintf(filepath2, "%s/../../%s.pressure", cgroup_root_path2, resource2);}
}
```

PR #4542, opened 2025-09-10
Control the use of min and max

guy-maurel:
Control and make a documentation of the use of the functions min and max


guy-maurel:
Thanks to you for the comment.

PR #4547, opened 2025-09-21
solve overflow size_t to int

### LINKED ISSUE #4537: Infinite loop on long condition
I hit an infinite loop trying to reformat the Opus codebase. I managed to reduce it down to this piece of code:

```
void test_enc_api(int i, int err)
{
   if(err!=OPUS_OK || (i!=OPUS_BANDWIDTH_NARROWBAND&&
      i!=OPUS_BANDWIDTH_MEDIUMBAND&&i!=OPUS_BANDWIDTH_WIDEBAND&&
      i!=OPUS_BANDWIDTH_FULLBAND))test_failed();
}
```

and the following configuration:

```
sp_before_sparen = add
indent_columns = 4
pos_bool = lead
code_width = 120
```

Version is Uncrustify-0.78.1_f

guy-maurel:
fixes #4537

### DIFF
--- a/src/reindent_line.cpp
+++ b/src/reindent_line.cpp
@@ -102,7 +102,22 @@ void reindent_line(Chunk *pc, size_t column)
       }
       else
       {
-         pc->SetColumn(max(pc->GetColumn() + col_delta, min_col));
+         LOG_FMT(LOUTIND, "%s(%d): pc->column %zu\n",
+                 __func__, __LINE__, pc->GetColumn());             // size_t
+         LOG_FMT(LOUTIND, "%s(%d): col_delta %d\n",
+                 __func__, __LINE__, col_delta);                   // int
+         LOG_FMT(LOUTIND, "%s(%d): min_col %zu\n",
+                 __func__, __LINE__, min_col);                     // size_t
+         int imin_col = min_col;
+         LOG_FMT(LOUTIND, "%s(%d): imin_col %d\n",
+                 __func__, __LINE__, imin_col);                    // int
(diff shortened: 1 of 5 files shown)


micheleCTDEAdmin (commented):
UHm, I will test this on 3200 files and report back. But:
1) there is way too much logging, we could combine then in one or two LOG_FMT max
2) the fix seems more like a trick. We are forcing a `size_t` into an `int` and magically everything works. This smells like the real problem is somewhere else and we are just hiding it.
3) the same could have been coded in a single line as
```
pc->SetColumn(max((int)(pc->GetColumn() + col_delta), (int)min_col));
```

Will report back after testing on more files. I suggest we hold this PR and see if we can find the real problem.

guy-maurel:
Your proposal is OK for me, BUT
both my proposal and yours get an **int** value to be set .
But the function is defined as:
_inline void Chunk::SetColumn(size_t col)_
The parameter should be **size_t**.
Is it OK?

micheleCTDE:
The single line of code in my previous comment was simply implementing the exact code implemented by this PR. I didn't mean to say that it is the correct solution to the problem. I think we need to look at at least two points:
1. the incorrect columns reported for the vbrace
2. the line reindentation
There may be more to it, but I need to have a detailed looked. I am planning to do some work on this during the weekend, I will come back with my findings.

micheleCTDE:
I tested this on 3200 files and didn't see any regression.

guy-maurel:
have a look to PR #4551

gmaurel:
see PR #4551

PR #4559, opened 2025-10-31
do not use a size_t function as a bool function

guy-maurel:
but make a complete statement, it is better to read


guy-maurel:
don't run properly

PR #4567, opened 2025-11-18
Qual: Apply include-what-you-use recommendations

mdeweerd:
Add includes to files where items provided by these includes are used. This follows recommendations by the static analysis tools and is good practice.


mdeweerd:
This PR is the group of changes from #4494 that only add includes to files.
When this is merged, I'll look for other groups of changes for PRs.

PR #4577, opened 2025-11-21
refactor(chunk/indent): String truncation and logging functionality

mdeweerd:
- Fixed potential buffer overflow in string truncation
- Refactored string truncation logic in ElidedText function for better readability and safety
- Added cstdio header include for printf functions (include-what-you-use)
- Improved logging functionality by making chunk_log functions const-correct
- Updated various functions to use const-correct Chunk pointers where appropriate

### DIFF
--- a/src/chunk.cpp
+++ b/src/chunk.cpp
@@ -6,6 +6,8 @@
  * @license GPL v2+
  */
 
+#include <cstdio>
+
 #include "chunk.h"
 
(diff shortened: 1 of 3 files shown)


mdeweerd:
@gmaurel Apparently this PR was not submitted in the "batch" I did before.

gmaurel (approved):

PR #4578, opened 2025-11-24
suppress the underrun

### LINKED ISSUE #4556: Deadlock with indent_ignore_first_continue = true; indent_paren_open_brace = true; indent_align_paren = false parameter set
```
[user@archlinux ~/p/uncrustify-bug]$ cat cfg.cfg
indent_ignore_first_continue = true
indent_paren_open_brace = true
indent_align_paren = false
```
```
[user@archlinux ~/p/uncrustify-bug]$ cat sample.cpp
void func() {
    call({ x, y, z });
}

void func() {
    call({
        x, y, z });
}

void func() {
    call({
        x, y, z
    }, w);
}

void func() {
    call({
        x, y, z
    });
}
```
```
[user@archlinux ~/p/uncrustify-bug]$ uncrustify -c ./cfg.cfg -f sample.cpp
do_source_file: Parsing: sample.cpp as language CPP
void func() {
	call({ x, y, z });
}

void func() {
	call({
x, y, z });
}

void func() {
	call({
x, y, z
^C⏎
```

In last sample, uncrustify never finishes. Removing any of 3 options resolves the issue, but I believe that deadlock is still unintended behavior.

guy-maurel:
fixes #4556

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1946,10 +1946,13 @@ void indent_text()
                log_rule_B("indent_align_paren");
                // We are inside ({ ... }) -- where { and ( are on the same line, avoiding double indentations.
                // only to help the vim command }
-               frm.top().SetBraceIndent(frm.prev().GetIndent() - indent_size);
-               indent_column_set(frm.top().GetBraceIndent());
-               frm.top().SetIndent(frm.prev().GetIndentTmp());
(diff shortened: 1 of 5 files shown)


guy-maurel:
a new version at #4578

PR #4586, opened 2025-12-12
Add 3 filtering capacities to span parameter `align_var_xxx_span` for {struct,class,def}

ericch1:
Hi Uncrustify team,

We work with an old codebase ([MEF++](https://fr.wikipedia.org/wiki/MEF%2B%2B), almost 29 years old!) that needs careful formatting. We tried hard to use `clang-format`, but it changed too much the formatting of our old code and caused ugly results (even if [we asked for modifications](https://github.com/llvm/llvm-project/issues/72897)). We found that only Uncrustify gives us the control we need. It's not aggressive and lets us "do nothing" when we want to keep things as they are. This is very helpful for our long-time project.

So, I'm happy to share a new feature that makes Uncrustify even better:

I've added 9 new options `align_var_{def,struct,class}_span_no_{empty,pp,cmt}_lines` numbers) **specifically to enhance the 'span' alignment settings.**

These new options let Uncrustify ignore (skip) a certain number of empty lines, preprocessor lines, or lines with only comments when it tries to align code sections, helping the alignment stay together and making the code look cleaner.

By default, these new options are set to '0', so they don't change how things work unless you turn them on.

**Example**

When `align_var_def_span = 1` and `align_var_def_span_no_cmt_lines = 2` and `align_var_def_span_no_pp_lines = 1` are active:

**Input:**
```cpp
int var1; //!< Some var/attributes
//! have a lot 
//! of comments
#ifdef DEBUG
float var2;
#endif
char var3;
```

**Output:**
```cpp
int   var1; //!< Some var/attributes
//! have a lot 
//! of comments
#ifdef DEBUG
float var2;
#endif
char  var3;
```
Here, the comment and preprocessor lines between `var1` and `var2` are skipped, allowing `float var2` and `char var3` to align with `int var1`.

About this work: The first ideas came from the Claude Opus 4.5 AI and Gemini (2.5-flash). But I spent a lot of time checking, cleaning, and making it much better before sharing it.

P.S. If you like this, I want to add this skip feature to other alignment options later.

### DIFF
--- a/tests/config/cpp/align_var_class_span_num_mixed_mix_var_proto.cfg
+++ b/tests/config/cpp/align_var_class_span_num_mixed_mix_var_proto.cfg
@@ -1,0 +1,8 @@
+# Enable mixing variable and prototype alignment
+align_mix_var_proto = true
+# Set spans to allow alignment across 1 non-skippable line
+align_var_class_span = 1
+# Allow skipping empty lines, preprocessor lines, and comment lines for variable definitions
+align_var_class_span_num_empty_lines = 1
+align_var_class_span_num_pp_lines = 1
+align_var_class_span_num_cmt_lines = 1
--- a/tests/cpp.test
+++ b/tests/cpp.test
@@ -1251,6 +1251,9 @@
 # Test for recursive align_var_def_brace with nested braces, comments and preprocessor directives
 34754  cpp/nested_brace_alignment.cfg                       cpp/nested_brace_alignment.cpp
 
+# Test for align_mix_var_proto option
+34755  cpp/align_var_class_span_num_mixed_mix_var_proto.cfg  cpp/align_var_class_span_num_mixed_mix_var_proto.cpp
+
 # Adopt some UT tests
 10000  common/empty.cfg                                     cpp/621_this-spacing.cpp
 10001  common/empty.cfg                                     cpp/622_ifdef-indentation.cpp
--- a/tests/expected/cpp/34755-align_var_class_span_num_mixed_mix_var_proto.cpp
+++ b/tests/expected/cpp/34755-align_var_class_span_num_mixed_mix_var_proto.cpp
@@ -1,0 +1,36 @@
+// Test: align_var_class_span with mixed empty/comment/PP lines
+// Tests interleaved PP, comments, and empty lines mixed with function prototypes.
+
+class TestClass
+{
+public:
+int               x;
+double            func1();
+unsigned long int y;
+#ifdef A
+//!< 1 PP + 1 cmt between y and z
+double            z;
+
+//!< 1 empty + 1 cmt between z and w
+float             w;
+
+//!< 1 empty + 1 cmt between w and func2
+double            func2();
+#endif
+#ifdef B
+//!< 2 PP + 1 cmt between w and s
+long s;
+//!< 1 cmt between s and t
+//!< 2 cmt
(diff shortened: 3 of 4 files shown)


guy-maurel:
laissez-moi un moment pour évaluer votre travail, qui semble fort intéressant.
merci et à bientôt.

ericch1:
Parfait, merci @guy-maurel .  J'ai un petit doute sur le nom des options que j'ai choisis, n'hésitez pas à me faire changer ça ou tout autre chose pour que ça soit mieux, ça me fera plaisir!

micheleCTDEAdmin:
Great addition @ericch1, the functionality for the empty lines is something I had been wanting to add at some point too.

micheleCTDEAdmin:
`documentation/htdocs/default.cfg` should also be updated with the new options

micheleCTDEAdmin:
In some countries/cultures/contexts, `no` is used as a short form for `number` but in others it is not and people may interpret the name `no_empty_lines` as meaning `there are no empty lines` rather than `number of empty lines`.
May I suggest we use `num` instead of `no` in the new option names?

ericch1:
> `documentation/htdocs/default.cfg` should also be updated with the new options

oups, ok, it's done now.

The files `documentation/htdocs/default.cfg` and `etc/defaults.cfg` seems identical.  

Should one be a symbolic link to the other?

(I can add this change in a separate commit).

ericch1:
> In some countries/cultures/contexts, `no` is used as a short form for `number` but in others it is not and people may interpret the name `no_empty_lines` as meaning `there are no empty lines` rather than `number of empty lines`. May I suggest we use `num` instead of `no` in the new option names?

Yes, me and a colleague were bugged by my initial choice after I created the pull-request...  I am willing to change this so it will be clear in the mind of everybody.

I renamed them all to `xxx_span_num_yyy_lines` with git-filter-repo:
```
rm -rf .git/filter-repo

git-filter-repo --force \
  --filename-callback '
    if b"_span_no_" in filename:
        return filename.replace(b"_span_no_", b"_span_num_")
    if b"_span_0_no_" in filename:
        return filename.replace(b"_span_0_no_", b"_span_0_num_")
    return filename
  ' \
  --commit-callback '
    if b"span_no_" in commit.message:
        commit.message = commit.message.replace(b"span_no_", b"span_num_")
    if b"span_0_no_" in commit.message:
        commit.message = commit.message.replace(b"span_0_no_", b"span_0_num_")
  ' \
  --blob-callback '
    if b"span_no_" in blob.data:
        blob.data = blob.data.replace(b"span_no_", b"span_num_")
    if b"span_0_no_" in blob.data:
        blob.data = blob.data.replace(b"span_0_no_", b"span_0_num_")  ' \
  --refs HEAD~9..HEAD
```

If you think a better choice exists, don't hesitate, it is now easy to re-run this command to rewrite the commits.

ericch1:
oups, forgot a few `_no_` in files and commit messages... re-done it...

ericch1:
@micheleCTDEAdmin I changed the PR description too to reflect the new names.

ericch1:
Thanks for accepting these additions!  We are very happy!  I am now working to add the same possibilities to `align_func_proto_span`...

micheleCTDE:
@ericch1 thanks for the adjustments after my comments and for the good addition to uncrustify code!

Issue #4589, opened 2025-12-22
Bug: `decltype`'d-return function's argument on new line confuses `sp_arith`


notgapriel:
Issue exists to demonstrate bug found within 2168d13962c10faafbda24493591bdacd4ea61ba, marked as version `Uncrustify-0.82.0-37-2168d1396`, compiled from source.

## Input file

```cpp
using defined_type = int;

decltype(auto) fn(
	defined_type& value_on_new_line
	) {
}
```

Output is identical to input if empty config is used to format it.

## Output of `uncrustify -c <config file> <input file>`

Config file contains only `sp_arith = add`.

```cpp
using defined_type = int;

decltype(auto) fn(
	defined_type & value_on_new_line
	) {
}
```

Noting there is now a space between `defined_type` and `&` despite it being a reference type.

## Expectation

If `defined_type` is substituted out for the keyword `int` in the input, no space is added, line reads as `int& value_on_new_line`. 

Noting that adding `sp_before_byref=remove` to the config does not change the output in any case.

notgapriel:
Noting that `decltype(anything)` triggers this issue as well, not just `decltype(auto)`

maxbachmann:
Using c++20 concepts on a function returning a pointer leads to similar confusion:
```cpp
template<typename T>
requires(!std::is_const_v<T>)
void* func(T* obj, void* &ptr)
{
	return nullptr;
}
```
is currently formatted as:
```cpp
template<typename T>
requires(!std::is_const_v<T>)
void* func(T * obj, void* & ptr)
{
	return nullptr;
}
```

with `sp_arith = add`

maxbachmann:
I do run into a similar issue for:
```cpp
#define TEST void fn(MyType& right) {}
```
which is formatted as:
```cpp
#define TEST void fn(MyType & right) {}
```

as well. For basic types like `int` instead of `MyType` the same issue doesn't occur.

maxbachmann:
```cpp
class MyClass
{
public:
    template<typename... T>
    void isOfType() {}

    template<typename T>
    const T &getValue(T* buffer)
    {
        return *static_cast<const T*>(buffer);
    }
};
```
The reuse of the template type T after the variadic template also confuses `sp_arith`.

This leads to the following formatting:
```cpp
class MyClass
{
public:
template<typename ... T>
void isOfType() {
}

template<typename T>
const T & getValue(T * buffer)
{
	return *static_cast<const T *>(buffer);
}
};
```

[...]

PR #4599, opened 2025-12-26
Fix: decltype-returning functions

notgapriel:
Please see https://github.com/uncrustify/uncrustify/issues/4589 for details about what this PR aims to fix.

This also fixes a previous test case where a one-liner was used in the expected output even though the test case did not configure "allow one-liners" to be on.

### DIFF
--- a/tests/config/cpp/bug_4589.cfg
+++ b/tests/config/cpp/bug_4589.cfg
@@ -1,0 +1,2 @@
+sp_before_byref=remove
+sp_arith=add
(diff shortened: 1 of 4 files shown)


notgapriel:
Just simplified some logic from an old commit, b5c4e19ab7167566210068f4fc7d771ad3121842, and rebased onto `master`.

notgapriel:
For posterity, head was previously at bb8a3361934028afe237114fd50036f2f29b0b50

notgapriel:
@gmaurel hello! You triggered a few tests on this PR but do not seem to have looked at it any further. I just want to make sure: is there anything preventing this PR from being reviewed that I can alleviate? I expect it's just a matter of time constraints but if not please let me know. If it is just time constraints, no rush, obviously, take your time. Thank you!

PR #4612, opened 2026-01-29
Fix case colon misclassification after elvis operator

PoeticPete:
When an elvis operator (?:) is encountered, mark_question_colon marks the colon as CT_COND_COLON. However, combine_labels pushes the ? onto a stack expecting a CT_COLON to pop it. Since the colon is already CT_COND_COLON, it never pops, causing subsequent case label colons to be incorrectly classified as conditional colons and moved to new lines with pos_conditional = lead.

The fix adds:
1. A condition in combine_labels to pop the question from the stack when we encounter a CT_COND_COLON that matches a question at the same level.
2. Elvis operator detection in mark_question_colon to properly mark the elvis colon without affecting the outer ternary's colon search.

Added tests 51005 and 51011 to verify the fix.

### DIFF
--- a/src/tokenizer/combine_labels.cpp
+++ b/src/tokenizer/combine_labels.cpp
@@ -155,6 +155,13 @@ void combine_labels()
             hit_case = true;
          }
       }
+      else if (  next->Is(CT_COND_COLON)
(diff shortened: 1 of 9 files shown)


gmaurel:
Thanks

PR #4613, opened 2026-01-29
Fix dictionary colon misclassified as ternary colon

PoeticPete:
## Summary

  When a ternary expression contains an OC dictionary literal (`@{...}`) in the true branch, the dictionary colon was incorrectly marked as `CT_COND_COLON`, causing the real ternary colon to be
   classified as `OC_COLON` and lose its spacing.

  For example:
  ```objc
  // Before fix - space before outer `:` was removed
  id result = [obj method:handler != nil ? @{kKey : handler, }: @{}];

  // After fix - correct spacing preserved
  id result = [obj method:handler != nil ? @{kKey : handler, } : @{}];
```

  Changes

  - Added brace_depth tracking in search_for_colon() for OC dictionary literals
  - Colons inside @{...} dictionaries are now correctly skipped when searching for the ternary colon
  - Added test 51012 to verify the fix

  Test plan

  - Test 51012 passes
  - All Objective-C tests pass
  - All C/C++ tests pass

  Stack

  This PR is stacked on fix-case-colon-after-ternary

### DIFF
--- a/src/tokenizer/mark_question_colon.cpp
+++ b/src/tokenizer/mark_question_colon.cpp
@@ -26,8 +26,9 @@ Chunk *search_for_colon(Chunk *pc_question, int depth, bool is_sibling_ternary =
 {
    Chunk *pc2                    = pc_question->GetNextNcNnl();
    bool  colon_found             = false;
-   bool  colon_after_colon_found = false;    // Test #51007: Track if we see a colon after our ternary colon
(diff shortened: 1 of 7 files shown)


micheleCTDEAdmin (approved):
LGTM, but need to solve the conflicts first

gmaurel (approved):

PR #4624, opened 2026-02-05
a solution for the warning nullPointerRedundantCheck

guy-maurel:
the next proposals from cppcheck will follow.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,123 +3,128 @@ exclude:
   (?x)^(
   __NONE__)$
 repos:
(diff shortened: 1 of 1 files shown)


mdeweerd:
@guy-maurel FYI - to skip a pre-commit hook (in ci), instead of commenting everything, you can just add `stages: [manual]  # temporary disabled`.  It's also possible to updat the "SKIP" variable set in the workflow action file.

guy-maurel:
@mdeweerd Thanks for the hint.
I only use the wrong file.

PR #4625, opened 2026-02-06
Fix mod_enum_last_comma adding comma to #define inside enum

PoeticPete:
When `mod_enum_last_comma = force` is enabled, uncrustify incorrectly adds a trailing comma to `#define` statements inside enum declarations. This happens because the code that adds trailing commas doesn't skip past preprocessor directives when looking for the last enum value.

The fix checks if the previous chunk is inside a preprocessor directive (`PCF_IN_PREPROC` flag) and skips past it, similar to the existing handling for `CT_PP_ENDIF`.

Added test case 51013 to verify the fix.


micheleCTDEAdmin (approved):
Looks good

PR #4585, opened 2025-12-10
Better OC support: new macro configs and auto-detect OC in .h headers

NSProgrammer:
This diff adds three significant improvements to uncrustify for handling Objective-C code:

## `macro-no-format-args`
For macros like `NS_SWIFT_NAME` where the content inside the parentheses should not be reformatted (e.g., Swift method signatures with colons like `getter:description()`).

## `macro-no-indent`
For single-line attribute macros like `NS_ASSUME_NONNULL_BEGIN` that should not affect the indentation of subsequent code.

## Auto-detect Objective-C content in `.h` headers
Previously, unless explicitly specified as Objective-C,`.h` files were parsed as C-Header which caused Objective-C spacing rules (like `sp_after_oc_property`) to not apply. Now, uncrustify scans `.h` files for OC-specific keywords (`@import`, `@interface`, `@implementation`, `@property`, `@protocol`, `@end`, `@class`, `@selector`, `@synthesize`, `@dynamic`) and automatically upgrades the language to Objective-C when found. This happens before keyword initialization, ensuring all OC-specific formatting rules apply correctly.

## Implementation Details

**New token types:**
- `CT_MACRO_NO_FORMAT_ARGS` - Tokens marked with this type have their paren contents preserved exactly
- `CT_MACRO_NO_INDENT` - Tokens marked with this type don't push/modify the indent stack

**New flag:**
- `PCF_IN_MACRO_NO_FMT_ARGS` - Flag set on tokens inside macro-no-format-args parens

**Key code changes:**
- `uncrustify.cpp`: Added `detect_oc_content()` function to scan raw file bytes for OC keywords; upgrades language before keyword initialization
- `option.cpp`: Parse the new config directives
- `keywords.cpp`: Output the new macro types when dumping config
- `tokenize_cleanup.cpp`: Mark tokens inside macro-no-format-args with the flag
- `space.cpp`: Early-exit spacing logic for flagged tokens; add `CT_MACRO_NO_INDENT` to spacing ignore list
- `indent.cpp`: Skip indent stack modification for `CT_MACRO_NO_INDENT` parens
- `combine.cpp`: Properly flag parens for macro types; fix `@protocol` angle bracket handling
- `combine_fix_mark.cpp`: Prevent function misclassification when preceded by these macro types

This fixes issues where uncrustify was incorrectly:
1. Adding/removing spaces inside Swift interop macro arguments
2. Causing subsequent code to be indented after single-line macros
[...]

PR #4627, opened 2026-02-06
[v2] Better OC support: new macro configs and auto-detect OC in .h headers

PoeticPete:
This is an iteration of https://github.com/uncrustify/uncrustify/pull/4585, so we can get it landed

It fixes the following error by renaming `MACRO_NO_FORMAT_ARGS ` to `CT_MACRO_NO_FMT_ARGS`:

```
main(472): The token name 'MACRO_NO_FORMAT_ARGS' is too long (20)
```


PoeticPeteAdmin:
> I didn't do a test, but it seems good. Does this PR replace PR #4585?

yep! I'll merge this and close #4585.

micheleCTDE:
@PoeticPete the call to `detect_oc_content(fm.raw))` can unnecessarily slow down processing since it will have to go thorugh all the input file. I think it may be better to add an option to enable/disable the OC detection upgrade rather than doing it automatically.
I should have catched this earlier....

PoeticPeteAdmin:
> @PoeticPete the call to `detect_oc_content(fm.raw))` can unnecessarily slow down processing since it will have to go thorugh all the input file. I think it may be better to add an option to enable/disable the OC detection upgrade rather than doing it automatically. I should have catched this earlier....

Good point! I'll add a flag for OC detection upgrade this week.

PR #4639, opened 2026-02-17
Fix issue 4545

fjmolinas:
Cast detection was skipped inside typedef blocks, causing inconsistent spacing:

```c
struct bare {
    uint8_t arr[(unsigned int)8];  // No space after cast
};

typedef struct typed {
    uint8_t arr[(unsigned int) 8];  // Unwanted space after cast
} typed_t;
```

## Changes

- **src/tokenizer/combine.cpp**: Removed `!pc->TestFlags(PCF_IN_TYPEDEF)` condition from cast detection logic. FPAREN/SPAREN parens are already marked before this check, so function pointer typedefs remain unaffected.
- **tests/**: Added test case `Issue_4545.c` validating consistent cast spacing in both contexts.

After fix, both contexts format identically: `(unsigned int)8`


fjmolinas:
Sorry for the noise, with the double PR

Issue #4636, opened 2026-02-13
Crash: stack-buffer-overflow in check_template() triggered by malformed template tokens


Criticayon:
## Summary

While fuzzing uncrustify with AFL++ and AddressSanitizer enabled, I discovered a reproducible crash caused by a stack buffer overflow in `check_template()` inside the tokenizer stage. The issue can be triggered by a crafted input file and results in an immediate abort under ASAN.

This appears to be a bounds error when writing past a fixed-size stack array (`tokens`) during template parsing.

---

## Environment

* uncrustify version: Uncrustify_d-0.82.0-132-bcc41cbdc (built from source)
* Compiler: clang
* Instrumentation: AFL++ + ASAN
* OS: Ubuntu 24.04.3 LTS

---

## Build Configuration

```
CC=afl-clang-fast
CXX=afl-clang-fast++
AFL_USE_ASAN=1


cmake ..
make -j
```

---

## Reproduction Steps
The configuration file is provided below.
```
# config.cfg used for reproduction
indent_columns=4
```

Run:

```
./uncrustify -c config.cfg -f poc_min.c
```

The program crashes reliably.

I verified reproducibility across multiple runs .

---

## PoC

See attached file: [poc_min.c](https://github.com/user-attachments/files/25282011/poc_min.c)

(Originally generated via AFL++ fuzzing and minimized with afl-tmin.)

---

## ASAN Report

```
do_source_file: Parsing: /home/fuzz_uncrustify/poc_min.c as language C
=================================================================
==4119354==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7f4c37a01020 at pc 0x5583da13a666 bp 0x7fffbc196290 sp 0x7fffbc196288
WRITE of size 4 at 0x7f4c37a01020 thread T0
    #0 0x5583da13a665 in check_template(Chunk*, bool) /home/new_fuzz_uncrustify/uncrustify/src/tokenizer/check_template.cpp:305:35
    #1 0x5583da2014fd in tokenize_cleanup() /home/new_fuzz_uncrustify/uncrustify/src/tokenizer/tokenize_cleanup.cpp:468:13
    #2 0x5583da251ef5 in uncrustify_start(std::deque<int, std::allocator<int>> const&) /home/new_fuzz_uncrustify/uncrustify/src/uncrustify.cpp:2013:4
[...]

PR #4650, opened 2026-02-21
Fix buffer overflow in tokenizer. This resolves issue #4636.

micheleCTDE:
This correctly checks for access to the `tokens` array and prevents the buffer overflow described in issue #4636. It does that even without increasing the size of `tokens`.
This PR replaces PR #4641.


micheleCTDE:
@guy-maurel let me know if it is ok to merge this one.

PR #4662, opened 2026-03-04
Make full qualified reference to E_Token values and remove the 'CT_' prefix.

micheleCTDE:
This is in preparation for changing it into an `enum class`, which will require some extra work to handle matching parenthesis-like tokens.

@guy-maurel please let me know if you are ok to merge this. I will work on the `enum class` changes after that. It is mostly a renaming + scoping of all the `CT_*` values. Using an `enum class` will ensure we don't allow invalid assignment/comparison by mistake.


micheleCTDE:
Looks like there is some extra work to do to pass the tests :smile:

micheleCTDE:
I am closing this and I will create a new one when I have fixed some of the problems. Looks like in appveyors there are errors that do not show up in my linux box, will need to spend more time on this.

PR #4610, opened 2026-01-23
Add 3 filtering capacities to span parameter `align_assign_span` and `align_enum_equ_span`

ericch1:
Hi again Uncrustify team,

   As the continuation of #4587, this PR adds 6 new options to control line skipping in assignment alignment:

     - align_assign_span_num_{empty,pp,cmt}_lines
     - align_enum_equ_span_num_{empty,pp,cmt}_lines

   These new options let Uncrustify ignore (skip) a certain number of empty lines, preprocessor lines, or comment-only lines when it tries to align assignments and enum values, helping the alignment stay together and making the code look cleaner.

   By default, these new options are set to '0', so they don't change how things work unless you turn them on.

   **Example**

   When `align_assign_span = 1` and `align_assign_span_num_cmt_lines = 2` are active:

   Input:
```
     int var1 = 100;
     // comment line 1
     // comment line 2
     double var2 = 200;

     unsigned short var3 = 300;
```

   Output:

```
     int var1    = 100;
     // comment line 1
     // comment line 2
     double var2 = 200;

     unsigned short var3 = 300;
```

   The `=` of `var1` and `var2` are aligned because the 2 comment lines are skipped. But `var3` is not aligned because the empty line breaks the span.

   **Implementation details:**

   - Use GetNlCountFiltered() instead of GetNlCount() for newline handling (with an exception)
   - Maintain separate budget config for assign vs enum contexts
   - Reset budget after each alignment addition and when starting a new alignment group
   - Budget is context-aware: switches between assign and enum configs when recursing into enum braces

   **Tests:**

   - 14 new C tests covering all 6 options
   - 33 new C++ tests covering recursive contexts, mixed line types, and interleaved prototypes with defaults

   About this work: The first ideas came from the Claude Opus 4.5 AI. But I spent a lot of time checking, cleaning, fixing bugs, and making it much better before sharing it.

   I hope this helps! Please tell me if anything needs changes. Thank you!

ps: There are "holes" into tests numbers because I removed some duplicates after (having a hard time) debugging more complex cases which I added after the simpler ones and decided to remove the simpler ones.

PR #4664, opened 2026-03-04
Add 3 filtering capacities to span parameter align func params span

ericch1:
Hi again Uncrustify team,

   As the continuation of #4610, this PR adds 6 new options to control line skipping in function parameter and typedef alignment:

     - align_func_params_span_num_{empty,pp,cmt}_lines
     - align_typedef_span_num_{empty,pp,cmt}_lines

   These new options let Uncrustify ignore (skip) a certain number of empty lines, preprocessor lines, or comment-only lines when it tries to align function parameters and typedef statements, helping the alignment stay together and making the code look cleaner.

   By default, these new options are set to '0', so they don't change how things work unless you turn them on.

   **Example (function parameters)**

   When `align_func_params_span = 1` and `align_func_params_span_num_pp_lines = 1` are active:

   Input:
```
     void process(int            param1,
                  #ifdef FEATURE_A
                  double         param2,
                  #endif
                  unsigned long  param3
                 );
```

   Output: param1, param2 and param3 all align because between each pair of consecutive parameters, only 1 PP line needs to be skipped.

   **Example (typedef)**

   When `align_typedef_span = 1` and `align_typedef_span_num_pp_lines = 1` are active:

   Input:
```
     typedef VeryLongTypeName TypeAlias1;
     #ifdef FEATURE_A
     typedef int              TypeAlias2;
     #endif

     typedef double           TypeAlias3;
```

   Output: TypeAlias1 and TypeAlias2 align because the #ifdef line is skipped. TypeAlias3 does not align because the empty line after #endif breaks the span.

   **Bug fixes included:**

   This PR also fixes 4 pre-existing parser/alignment bugs that were exposed when PP directives appear inside function parameters at a nested level (inside a struct, class, or namespace):

   1. `fix_fcn_def_params()` would bail out when encountering a `#define` directive at a lower nesting level. Fixed by skipping PP directives instead of bailing.
   2. `mark_function()` used `GetNextNcNnl()` which would traverse into PP directive tokens, causing `can_be_full_param()` to return false and FUNC_PROTO to be wrongly reclassified as FUNC_CTOR_VAR. Fixed by using `GetNextNcNnlNpp()`.
   3. `can_be_full_param()` did not handle macro-wrapped parameters like `int _MACRO(param)`. Fixed by adding a fallback case for CT_PAREN_OPEN.
   4. `align_func_param()` would break its alignment loop when encountering PP tokens at a lower nesting level. Fixed by adding a PCF_IN_PREPROC check to the break condition.

  (Just for the story: I started to add these features on Dec. 18,  but was blocked by these bugs which caused me a surprise and a lot of tests to add...)

   **Implementation details:**

   - Use GetNlCountFiltered() instead of GetNlCount() for newline handling in align_func_param
   - Add PCF_IN_PREPROC check to the alignment break condition so PP tokens don't prematurely stop the alignment loop
   - Keep diagnostic LOG_FMT statements (LAS severity) for future debugging

   **Tests:**

   - 9 new C tests for align_func_params_span_num_*_lines
   - 24 new C++ tests for align_func_params_span_num_*_lines (including nested structures, mixed line types, and macro calls)
   - 9 new C tests for align_typedef_span_num_*_lines
   - 9 new C++ tests for align_typedef_span_num_*_lines

   About this work: The first ideas came from the Claude Opus 4.5/4.6 AI. But I spent a lot of time checking, cleaning, fixing bugs, and making it much better before sharing it.

   I hope this helps! Please tell me if anything needs changes, especially for the 4 "Fix" commits. Thank you!

### DIFF
--- a/tests/c.test
+++ b/tests/c.test
@@ -689,3 +689,18 @@
 10206  common//align_func_params_span_num_pp_lines-1.cfg       c/align_func_params_span_num_pp_lines.c
 10207  common//align_func_params_span_num_pp_lines-2.cfg       c/align_func_params_span_num_pp_lines.c
 10208  common//align_func_params_span_num_pp_lines-3.cfg       c/align_func_params_span_num_pp_lines.c
+
+# align_typedef_span_num_* tests - Empty lines
+10209  common/align_typedef_span_num_empty_lines-1.cfg         c/align_typedef_span_num_empty_lines.c
+10210  common/align_typedef_span_num_empty_lines-2.cfg         c/align_typedef_span_num_empty_lines.c
+10211  common/align_typedef_span_num_empty_lines-3.cfg         c/align_typedef_span_num_empty_lines.c
+
+# align_typedef_span_num_* tests - Comment lines
+10212  common/align_typedef_span_num_cmt_lines-1.cfg           c/align_typedef_span_num_cmt_lines.c
+10213  common/align_typedef_span_num_cmt_lines-2.cfg           c/align_typedef_span_num_cmt_lines.c
+10214  common/align_typedef_span_num_cmt_lines-3.cfg           c/align_typedef_span_num_cmt_lines.c
+
+# align_typedef_span_num_* tests - Preprocessor lines
+10215  common/align_typedef_span_num_pp_lines-1.cfg            c/align_typedef_span_num_pp_lines.c
+10216  common/align_typedef_span_num_pp_lines-2.cfg            c/align_typedef_span_num_pp_lines.c
+10217  common/align_typedef_span_num_pp_lines-3.cfg            c/align_typedef_span_num_pp_lines.c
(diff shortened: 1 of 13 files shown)


micheleCTDEAdmin:
If I see it correctly, the config files for C and CPP are identical. In such case we should have a single copy under the `common` folder instead of duplicating the files.

Note: I noticed that this also happened for PR #4610. It would be good to do the same simplification for the files added there too. You can use a new PR for de-duplicating those files.

ericch1:
> If I see it correctly, the config files for C and CPP are identical. In such case we should have a single copy under the `common` folder instead of duplicating the files.

ok, I didn't knew that was existing.  I did the move for this PR, I push it right now, thanks for telling!

> 
> Note: I noticed that this also happened for PR #4610. It would be good to do the same simplification for the files added there too. You can use a new PR for de-duplicating those files.

Right!  I will have to do this for my other PRs too (#4586 #4587)...

micheleCTDEAdmin:
I will run a test on 3200 source files and see if those "fixes" cause any regression. I will feedback after that.

Issue #4536, opened 2025-08-13
uncrustify --update-config breaks quoted strings


bfredl:
To reproduce, use https://gist.github.com/bfredl/09f7a208507510c6be20e8996d1e55e0 

```
uncrustify -c uncrustify.cfg --update-config-with-doc -o uncrustify_new.cfg
diff -u uncrustify.cfg uncrustify_new.cfg
```

This gives:
```diff
@@ -3575,7 +3575,7 @@
 #
 
 # The regex for include category with priority 0.
-include_category_0              = "(?!.*generated.h\").*"             # string
+include_category_0              = "(?!.*generated.h").*"             # string
 
 # The regex for include category with priority 1.
 include_category_1              = ""         # string
```

the regex in `include_category_0` is broken as removing the backslash before the embedded quote changes the meaning.

PR #4694, opened 2026-04-13
fix: uncrustify --update-config breaks quoted strings

andrey-starodubtsev:
Fix for https://github.com/uncrustify/uncrustify/issues/4536

### DIFF
--- a/src/option.cpp
+++ b/src/option.cpp
@@ -1341,7 +1341,36 @@ void save_option_file(FILE *pfile, bool with_doc, bool minimal)
 
          if (option->type() == OT_STRING)
          {
-            fprintf(pfile, "\"%s\"", val.c_str());
+            size_t to_be_escaped = 0;
+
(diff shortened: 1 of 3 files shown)


src/option.cpp
```diff
@@ -1341,7 +1341,19 @@ void save_option_file(FILE *pfile, bool with_doc, bool minimal)
 
          if (option->type() == OT_STRING)
          {
-            fprintf(pfile, "\"%s\"", val.c_str());
+            std::string escaped;
+            escaped.reserve(val.length());
```
micheleCTDEAdmin: It may be good to reserve a couple of extra characters, to cater for the possibility of some extra backslashes that need to be added.
Other than that it seems good

andrey-starodubtsev: yup, fixed

micheleCTDE (approved):
Looks good.

PR #4697, opened 2026-04-16
Simplify test cpp:33057.

micheleCTDE:
This test was taking 156 seconds in Github actions on Windows - Debug. The test has been simplified to keep the functionality but avoid a lot of the repetition that caused the long execution. It now takes less than 10 seconds.


guy-maurel (approved):
fine

micheleCTDEAdmin (approved):

PR #4698, opened 2026-04-16
feat: add global_span_num_{empty,pp,cmt}_lines options

ericch1:
Hi again Uncrustify team,

   This PR adds 3 global options that serve as fallback defaults for all 24 existing `*_span_num_{empty,pp,cmt}_lines` options, following up on #4675, #4664, #4610, #4587, and #4586.

   **Motivation**

   Currently, a user who wants a uniform line-skipping policy across all alignment families must set 24 individual options. This PR introduces 3 global options that act as defaults, drastically reducing boilerplate in configuration files.

   **New options**

   - `global_span_num_empty_lines` (unsigned, default 0)
   - `global_span_num_pp_lines` (unsigned, default 0)
   - `global_span_num_cmt_lines` (unsigned, default 0)

   **How it works**

   All 24 specific `*_span_num_*_lines` options change from `unsigned` (default 0) to `signed` (default -1):

   - `-1` (default): use the corresponding `global_span_num_*_lines` value.
   - `0`: explicitly disable line skipping for this family (even if global is nonzero).
   - `>0`: use this value, overriding the global.

   This means existing configurations are fully backward compatible: all specific options default to -1, and all global options default to 0 — so the effective behavior is identical to before (no lines skipped).

   **Example**

   Instead of writing:
```
align_assign_span_num_pp_lines     = 2
align_var_def_span_num_pp_lines    = 2
align_func_proto_span_num_pp_lines = 2
align_typedef_span_num_pp_lines    = 2
# ... 4 more families ...
```

   You can simply write:
```
global_span_num_pp_lines = 2
```

   And if one family needs a different value:
```
global_span_num_pp_lines           = 2
align_assign_span_num_pp_lines     = 5   # override for assignments
```

   **Implementation details**

   - New helper `resolve_span_num_config()` in `src/align/span_num_resolve.h` centralizes the -1 → global resolution logic, avoiding repetition across callsites.
   - All 5 alignment source files (`assign.cpp`, `func_params.cpp`, `func_proto.cpp`, `typedefs.cpp`, `var_def_brace.cpp`) updated to use the helper.
   - Documentation updated in `documentation/align-span-line-filters.txt`.

   **Tests**

   - 42 new tests (21 C + 21 C++) across all 8 alignment families, using mixed input files (interleaved empty, comment, and PP lines).
   - 3 scenarios per family:
     1. Global fallback only (specific = -1, global sets values)
     2. Global set, specific overrides to 0 (disables filtering)
[...]

PR #4701, opened 2026-04-20
Add align_constr_value_span_num_{empty,pp,cmt}_lines options

ericch1:
Hi Uncrustify team,

This PR extends the `*_span_num_{empty,pp,cmt}_lines` budget mechanism to constructor value alignment (`align_constr_value_span`), completing the 9th alignment family.

### New options

- `align_constr_value_span_num_empty_lines` (signed, default -1)
- `align_constr_value_span_num_pp_lines` (signed, default -1)
- `align_constr_value_span_num_cmt_lines` (signed, default -1)

### How it works

Same semantics as all other `*_span_num_*_lines` options:
- `-1` (default): use the corresponding `global_span_num_*_lines` value
- `0`: explicitly disable line skipping (even if global is nonzero)
- `>0`: use this value, overriding the global

### Bug fix included

The existing code in `class_colon_pos.cpp` called `constructorValue.NewLines(paren_vor_value->GetNlCount())`, but `GetNlCount()` on `CT_FPAREN_OPEN` chunks always returns 0 (the `m_nlCount` field is only meaningful for `CT_NEWLINE` chunks). This made the span mechanism essentially non-functional for constructor values.

This PR replaces the call with a proper counting loop that iterates between alignment candidates using `GetNlCountFiltered()`, consistent with how other alignment families (e.g., `align_assign`) count newlines.

### Changes

**Code:**
- `src/options.h`: 3 new `BoundedOption<signed, -1, 5000>` options
- `src/newlines/class_colon_pos.cpp`: Budget resolution via `resolve_span_num_config()`, proper NL counting loop at both call sites (colon path and comma path), budget reset after each `Add()` and at `End()`

**Documentation:**
- `documentation/align-span-line-filters.txt`: Added `align_constr_value_span` row
- Auto-generated files updated: `defaults.cfg`, `default.cfg`, `uigui_uncrustify.ini`, CLI test references

**Tests (8 new, 34881-34888):**
- 5 specific budget configurations testing each budget type independently
- 3 global fallback configurations testing -1 fallback, 0 override, and specific override
- Input file with 2 classes: BasicCtorAlign (baseline) and MixedLinesCtorAlign (8 members with interleaved empty lines, PP directives, and comments)

### Backward compatibility

Fully backward compatible. The 3 new options default to -1 (use global), and the global options default to 0 (no line skipping), so the effective behavior is identical to before when no options are explicitly set.

About this work: The first ideas came from the Claude Opus 4.6/4.5 AI. But I spent time checking, cleaning, and making it much better before sharing it.


src/newlines/class_colon_pos.cpp
```diff
@@ -57,8 +58,14 @@ void newlines_class_colon_pos(E_Token tok)
...
-   bool       with_acv = (acv_span > 0) && language_is_set(lang_flag_e::LANG_CPP);
-   AlignStack constructorValue;    // ABC_Member(abc_value)
+   bool with_acv = (tok == E_Token::CT_CONSTR_COLON)
+                   && (acv_span > 0)
+                   && language_is_set(lang_flag_e::LANG_CPP);
+   AlignStack     constructorValue; // ABC_Member(abc_value)
+
+   LineSkipConfig acv_skip_cfg    = {};
+   LineSkipConfig acv_skip_budget = {};
+   Chunk          *prev_acv_chunk = nullptr;
```
micheleCTDEAdmin: You should not use `nullptr` for a chunk but rather `Chunk::NullChunkPtr`. Later you test whether a chunk is the NullChunk or not.

src/newlines/class_colon_pos.cpp
```diff
@@ -183,8 +218,25 @@ void newlines_class_colon_pos(E_Token tok)
                LOG_FMT(LBLANKD, "%s(%d): paren_vor_value orig line is %zu, orig col is %zu, text '%s', type is %s\n",
                        __func__, __LINE__, paren_vor_value->GetOrigLine(), paren_vor_value->GetOrigCol(),
                        paren_vor_value->GetLogText(), get_token_name(paren_vor_value->GetType()));
-               constructorValue.NewLines(paren_vor_value->GetNlCount());
+               size_t nl_count = 0;
+
+               if (prev_acv_chunk != nullptr)
```
micheleCTDEAdmin: You can skip this `if` check if you use `NullChunkPtr` as mentioned in the other comment. `tmp` will be the `NullChunk` and the `for` loop will do nothing.

micheleCTDEAdmin (changes requested):
Some changes required about the usage of `nullptr` rather than `NullChunk`.

src/newlines/class_colon_pos.cpp
EricChamberland: I see.  I'm modifying it.

(Sorry for the delay, I didn't saw the email notifications about your comments...)

micheleCTDEAdmin (approved):
Looks good

ericch1:
Just to keep track: this PR followed #4698.

PR #4706, opened 2026-04-28
Switch github actions to nodejs 24 explicitly, following deprecation of version 20

micheleCTDE:
See [here](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) for info about the deprecation of nodejs 20 on github.


micheleCTDE:
Will replace with another PR since this didn't solve the warning issue.

PR #4708, opened 2026-04-29
Fix "Unmatched BRACE_OPEN" crash with split extern "C" guards.

LucaCiucci:
This fixes the problem described in #4681.

When `{` and `}` of an `extern "C"` block appear in separate `#ifdef __cplusplus` guards containing an `#else` directive, uncrustify would crash with exit code 74.

The Root cause was in `ParsingFrameStack::check()` as described in 0507f862b51ad89bc1196f74e226dee792dd6c8c .

Fixes #4681 .

### DIFF
--- a/src/parsing_frame_stack.cpp
+++ b/src/parsing_frame_stack.cpp
@@ -346,6 +346,7 @@ int ParsingFrameStack::check(ParsingFrame &frm, int &pp_level, Chunk *pc)
              */
             // check if #if block was unbalanced
             size_t brace_level = frm.GetBraceLevel();
+            size_t cur_size    = frm.size();
             pop(frm);
 
             if (brace_level != frm.GetBraceLevel())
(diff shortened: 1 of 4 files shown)


micheleCTDEAdmin (changes requested):
This fix causes regressions in real code. I can prepare a test case if needed, it may be a bit complex.
But the main point I want to make is that as explained in `documentation/preprocessor_indentation.txt`, preprocessor indentation can be extremely tricky and therefore `uncrustify` has taken then approach to discard level changes made inside preprocessor blocks. Warning for unbalanced preprocessor blocks are available if enabled in the configuration file.

LucaCiucci:
Hi @micheleCTDEAdmin , thank you for the review! I'll try to revise the fix keeping that document in mind.
It would be great if you could share a minimal example that reproduces the regression. I think it could be useful since all the other tests are passing.

micheleCTDE:
Ciao Luca,
I have attached a minimal example (config + input file). The problem is actually triggered by the `case` in the preprocessor block, which adds a brace before `#endif`. Without this PR the additional brace is basically ignored, because as explained in the documentation `uncrustify` will restore the same level before `#ifdef` 
It may be a bit of a corner case, but there is no really any 'right" solution for how to handle conditional preprocessor blocks. 
[Archive.zip](https://github.com/user-attachments/files/27317672/Archive.zip)

LucaCiucci:
Sorry for the long silence. After spending some time trying to revise the fix, I couldn't come up with a correct solution that handles both the original crash and the formatting regression, so I'm going to close this PR.

A proper fix would likely require more extensive tracking within the frame stack, which is beyond the scope of what I can address here (also considering `preprocessor_indentation.txt`).

PR #4712, opened 2026-05-06
as shown by cppcheck

guy-maurel:
part 2b


gmaurel (approved):
some more will follow

PR #4714, opened 2026-05-06
Add align_same_func_call_params_span_num_{empty,pp,cmt}_lines options

ericch1:
Hi Uncrustify team,

This PR extends the `*_span_num_{empty,pp,cmt}_lines` budget mechanism to same-function-call parameter alignment (`align_same_func_call_params_span`), following #4587, #4610, #4664, #4675, #4698, and #4701.

### New options

- `align_same_func_call_params_span_num_empty_lines` (signed, default -1)
- `align_same_func_call_params_span_num_pp_lines` (signed, default -1)
- `align_same_func_call_params_span_num_cmt_lines` (signed, default -1)

### How it works

Same semantics as all other `*_span_num_*_lines` options:
- `-1` (default): use the corresponding `align_span_num_*_lines` global fallback value
- `0`: explicitly disable line skipping (even if global is nonzero)
- `>0`: use this value, overriding the global

### Implementation

The implementation uses a deferred NL counting : instead of counting newlines as they are encountered, newlines are counted between consecutive matching `CT_FUNC_CALL` chunks using `GetNlCountFiltered()`.

### Changes

**Code:**
- `src/options.h`: 3 new `BoundedOption<signed, -1, 5000>` options
- `src/align/same_func_call_params.cpp`: Budget resolution via `resolve_span_num_config()`, deferred NL counting loop before each `Add()` for both stacks, budget reset after each counted gap

**Documentation:**
- `documentation/align-span-line-filters.txt`: Added `align_same_func_call_params_span` row
- Auto-generated files updated: `defaults.cfg`, `default.cfg`, `uigui_uncrustify.ini`, CLI test references

**Tests C/C++:**
- 3 empty-line budget configurations
- 3 PP-line budget configurations
- 3 comment-line budget configurations
- 6 mixed configurations (all three budget types combined)
- 3 global fallback configurations in `tests/config/common/`

### Backward compatibility

Fully backward compatible. The 3 new options default to -1 (use global), and the global options default to 0 (no line skipping), so the effective behavior is identical to before when no options are explicitly set.

About this work: The first ideas came from the Claude Sonnet 4.6 AI. But I spent a lot of time checking, cleaning, and making it much better before sharing it.


ericch1:
I used a symlink in the original commit, but it failed on windows... I changed it for a copy.  Don't know if there is a better way to do this?  The idea is to have the same file, but treated as a C++ input... (since in some cases, it may changes the behaviour).

micheleCTDE:
> I used a symlink in the original commit, but it failed on windows...

Can you provide some more info? which file are you referring to?

ericch1:
> > I used a symlink in the original commit, but it failed on windows...
> 
> Can you provide some more info? which file are you referring to?

I'm talking about: `tests/input/cpp/align_sf_call_span_num_mixed_c.cpp` which is now a simple copy of `tests/input/c/align_sf_call_span_num_mixed.c`, but renamed as a C++ file so uncrustify switches to c++ mode.

I once got a bug (see d5eac36b4f80810) specific to C++ (`if (  language_is_set(lang_flag_e::LANG_CPP)`), so I do this to fully test the feature both in C and C++ if the code ever changes to treat C++ specific things.

micheleCTDE:
> I'm talking about: tests/input/cpp/align_sf_call_span_num_mixed_c.cpp which is now a simple copy of tests/input/c/align_sf_call_span_num_mixed.c, but renamed as a C++ file so uncrustify switches to c++ mode.

You can place that file in `tests/input/common` and reference that from the test lists. No need for any symlink or for file duplication.
You can force a specific language in the config by using `file_ext` at the end, so you can run one test as `C` and one as `C++`

ericch1:
> You can place that file in tests/input/common and reference that from the test lists. No need for any symlink or for file duplication.

Ok, and if I do this, will the input be treated as 'C++' if the filename extension is `.c` ?

micheleCTDE:
Yes, as long as you specify that in the config file with something like `file_ext CPP .cpp .c .hpp .h`

ericch1:
ok, since I share the config file between C and C++, I can't modifiy it.  But I can do this in `c.test`:
```
10241  common/align_sf_call_span_num_empty-1.cfg     common/align_sf_call_span_num_mixed.c C
```
and `cpp.test`:
```
34907  common/align_sf_call_span_num_empty-1.cfg     common/align_sf_call_span_num_mixed.c CPP
```
now the funny thing is that expected output ends into `expected/common` directory for both tests...  since both are identical,  it would be "nice" to have the same filename... but id din't changed that as it could also be confusing if an error occurs... or if both tests are launched at the same time, a kind of race-condition may appear...

So I have added another commit with these modifications with these file moved, tell me if it seems ok or not.  Maybe I should just remove the C++ tests with the C file?

Thanks,

Eric

micheleCTDE:
> Maybe I should just remove the C++ tests with the C file?

If you think the tests are meaningful, we should keep them. If you think they are not really essential, perhaps we can drop them and spend a bit less time in each test run. But I leave the choice to you.

> now the funny thing is that expected output ends into expected/common directory

Yeah, it looks a bit funny, perhaps confusing. In hindsight, perhaps the bst solution was to have a single duplicated source file in `c/` and `cpp/` as you did initially. It looks clearer. But I had not thought about all the consequences of putting the file in `common`, sorry.
What do you think?

ericch1:
> > Maybe I should just remove the C++ tests with the C file?
> 
> If you think the tests are meaningful, we should keep them. If you think they are not really essential, perhaps we can drop them and spend a bit less time in each test run. But I leave the choice to you.

I see it as a guarantee that the C++ treatments stay valid in the futur.

> 
> > now the funny thing is that expected output ends into expected/common directory
> 
> Yeah, it looks a bit funny, perhaps confusing. In hindsight, perhaps the bst solution was to have a single duplicated source file in `c/` and `cpp/` as you did initially. It looks clearer. But I had not thought about all the consequences of putting the file in `common`, sorry. What do you think?

Since symbolic links are not supported, I also prefer to put a copy of the source file, but I propose to add a comment into the files *and* into the `c.test` and `cpp.test` files that it is a copy of te same file to remember that choice and this PR # as a reference to keep note of why we decided to do so...

ericch1:
I added the comments into a separated commit.  If all looks good to you, then you can squash the last commit into his parent before merging if you want (or I can do it for you).

Just tell me what you prefer.

Thanks,
Eric

micheleCTDEAdmin (commented):
It looks good.
We can keep the last comment separated, no prob.
I will merge since we are doing a new release tomorrow and it is good to include this PR too.

micheleCTDEAdmin (approved):

PR #4719, opened 2026-05-16
Update github actions to Visual Studio 2026 to avoid warnings

micheleCTDE:
As per title.


micheleCTDE:
This one is a funny failure. The tests success in my own fork where VS 2026 is available, but the same does not here in the main uncrustify repository, where apparently VS 2026 is not yet available. I guess we will need to wait a while till all GH builders are using the latest windows 2025 image (should be default from May 12 but apparently it isn't :-( )

micheleCTDEAdmin (approved):

Issue #4673, opened 2026-03-18
Heap-buffer-overflow in Chunk::IsStringAndLevel triggered by malformed C input


Criticayon:
## Description

The crash happens in: ```Chunk::IsStringAndLevel```
which performs a memcmp on a buffer returned by GetLogText().
Under certain malformed inputs, the internal chunk text buffer length becomes inconsistent with the length returned by Len(). This leads to an out-of-bounds read in memcmp.

ASAN reports indicate that the comparison reads past the allocated buffer.

---

## Environment

* uncrustify version: Uncrustify_d-0.82.0-205-c036e924b (built from source)
* Compiler: clang
* Instrumentation: AFL++ + ASAN
* OS: Ubuntu 24.04.3 LTS

---

## Build Configuration
This is the build function I’m using in my compilation script
```
build() {
  export AFL_USE_ASAN=1
  cd $SRC_DIR
  mkdir -p build-afl
  cd build-afl

  CC=afl-clang-fast \
  CXX=afl-clang-fast++ \
  cmake ..

  make -j$(nproc)

  echo "[+] build done"
}
```

---

## Reproduction Steps
The configuration file is provided below.
```
# config.cfg used for reproduction
indent_columns=4
```

Run:

```
./uncrustify -c config.cfg -f poc.c
```

The program crashes reliably.

---

## PoC

See attached file: [poc.c](https://github.com/user-attachments/files/26085597/poc.c)

---

## ASAN Report

ASAN frame info:

```
Output suffix: .uncrustify
do_source_file: Parsing: poc.c as language C
parse_cleanup(481): pc orig line is 1, orig col is 2, text is ')', type is PAREN_CLOSE
parse_cleanup(481): pc orig line is 1, orig col is 10, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 1, orig col is 11, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 1, orig col is 12, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 2, orig col is 16, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 2, orig col is 17, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 2, orig col is 18, text is '}', type is BRACE_CLOSE
parse_cleanup(481): pc orig line is 2, orig col is 31, text is '}', type is BRACE_CLOSE
[...]

PR #4730, opened 2026-05-24
Fix for read access beyond array boundary

andrey-starodubtsev:
This MR provides the fix for situation described in https://github.com/uncrustify/uncrustify/issues/4682.
Also fixes the issue from https://github.com/uncrustify/uncrustify/issues/4673.

### DIFF
--- a/src/semicolons.cpp
+++ b/src/semicolons.cpp
@@ -42,7 +42,7 @@ void remove_extra_semicolons()
    while (pc->IsNotNullChunk())
    {
       Chunk *next = pc->GetNextNcNnl();
-      Chunk *prev;
+      Chunk *prev = Chunk::NullChunkPtr;
 
       if (  pc->Is(E_Token::CT_SEMICOLON)
          && !pc->TestFlags(PCF_IN_PREPROC)
--- a/tests/cli/test_cli_options.py
+++ b/tests/cli/test_cli_options.py
@@ -475,7 +475,7 @@ def string_replace(string_target, replacement):
 
 
 def s_path_join(path, *paths):
-    """
+    r"""
     Wrapper for the os.path.join function, splits every path component to
     replace it with a system specific path separator. This is for consistent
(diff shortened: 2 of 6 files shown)


src/semicolons.cpp
```diff
@@ -42,7 +42,7 @@ void remove_extra_semicolons()
    while (pc->IsNotNullChunk())
    {
       Chunk *next = pc->GetNextNcNnl();
-      Chunk *prev;
+      Chunk *prev = nullptr;
```
micheleCTDEAdmin: Chunk pointers should never be `nullptr`. They should be `Chunk::NullChunkPtr` in case of null chunks.

tests/cli/test_cli_options.py
```diff
@@ -491,7 +491,7 @@ def s_path_join(path, *paths):
         a joined path, see os.path.join
 
     >>> s_path_join('./z/d/', '../a/b/c/f')
-    r'.\z\a\b\c\f'
+    r'.\\z\\a\\b\\c\\f'
```
micheleCTDEAdmin: It's a raw string, why do we need the double `\\`?

micheleCTDEAdmin (changes requested):
Please take a look at the usage of `nullptr` and adjust as required.

src/semicolons.cpp
andrey-starodubtsev: fixed, thank you

tests/cli/test_cli_options.py
andrey-starodubtsev: `r''` is located in `"""` heredoc string and is not applied

I guess, in this case it'd be better to use `r"""` for whole heredoc.

Fixed

micheleCTDEAdmin: Ah thanks for pointed that out. Yes, r""" seems more readable.

micheleCTDEAdmin (approved):
Looks good, thanks.

PR #4733, opened 2026-06-05
Add span_num_{empty,pp,cmt}_lines budgets to align_right_cmt_span

ericch1:
First, fixes the `align_right_cmt_span` option to behave like other span options when set to `1`.
This fix requires changes to existing configuration files.

Second, adds three new budget options that allow `align_right_cmt_span` to skip
empty lines, preprocessor directives, and comment-only lines when evaluating
whether two trailing comments are close enough to be aligned together.

This follows the exact same pattern already established for other `align_*_span` options.

## New Options

| Option | Default | Description |
|--------|---------|-------------|
| `align_right_cmt_span_num_empty_lines` | -1 | Budget of empty lines to skip |
| `align_right_cmt_span_num_pp_lines`    | -1 | Budget of preprocessor lines to skip |
| `align_right_cmt_span_num_cmt_lines`   | -1 | Budget of comment-only lines to skip |

**Value semantics** (same as all other budget options):
- `-1`: Inherit from `global_span_num_xxx_lines` (default)
- `0`: Disable skipping for this span (overrides global)
- `>0`: Skip up to this many lines (overrides global)

## Motivation

In many C/C++ codebases, trailing comments are frequently separated by
blank lines, `#ifdef`/`#define` blocks, or comment paragraphs:

```c
int a  = 1; /* first item  */
int bb = 2; /* second item */

int ccc = 3; /* third item — separated by blank line */
```

With budget options and `align_right_cmt_span = 1`, one can more easily control
when a line breaks the span. With
`align_right_cmt_span_num_empty_lines = 1`, `a`, `bb`, and `ccc` are aligned.

## Backward Compatibility

All new options default to `-1` (inherit from global). If the global budget
options are at their default of `0`, the new options behave identically to
the pre-existing `align_right_cmt_span` — **no change to existing output**.

About this work: The first ideas came from the Claude Sonnet 4.6 AI. But I spent a lot of time checking, cleaning, and making it much better before sharing it.


micheleCTDEAdmin (approved):
Looks good, thanks.

PR #4735, opened 2026-06-08
untill_sorting

guy-maurel:
as proposed from cppcheck


gmaurel (approved):

micheleCTDE:
`#include "../build/uncrustify_version.h"`

No, we can't hard code build paths! It will break building if a user uses a different path :-(

micheleCTDEAdmin:
> `#include "../build/uncrustify_version.h"`
> 
> No, we can't hard code build paths! It will break building if a user uses a different path :-(

PR #4759 removes the hardcoded path.

PR #4752, opened 2026-07-09
Revise logic to mark constructors in structs

### LINKED ISSUE #4748: Constructor parameters inside struct not aligned in header files
When processing C++ header files, parameters of constructors inside `struct` definitions are not aligned, while those inside `class` definitions are.  See below for a reproducer.

This behavior could be traced down to not labeling constructors inside `struct` at all in case of a header file, see [this code snippet](https://github.com/uncrustify/uncrustify/blob/ac12adc2cf26e663f912b00ca553b0864aae5cb1/src/tokenizer/EnumStructUnionParser.cpp#L1002-L1011). By default, a header is classified as  `C-Header`, corresponding to `e_LANG_C | e_LANG_CPP | e_FLAG_HDR`.  That is, `LANG_C` is true and thus the second part of the `if` condition evaluates to `false`, skipping the call to `mark_constructors`.  But since `LANG_CPP` is also true, this is clearly wrong.

Unfortunately, my initial attempts to fix the condition led to failures in C#/VALA tests, i.e., I'm probably missing some context here.

```shell
$ uncrustify --version
Uncrustify-0.83.0-51-ac12adc2c
```
Input file "foo.h":
```cpp
class Foo
{
public:
Foo(int a,
double b);
};

struct Bar
{
Bar(int a,
double b);
};
```
Configuration file:
```
indent_columns = 4
indent_class = true
align_func_params = true
```
Output:
```cpp
class Foo
{
public:
    Foo(int    a,
        double b);
};

struct Bar
{
    Bar(int a,
        double b);
};
```
Expected output:
```cpp
class Foo
{
public:
    Foo(int    a,
        double b);
};

struct Bar
{
    Bar(int    a,
        double b);
};
```

geimer:
Fixes: #4748

### DIFF
--- a/src/tokenizer/EnumStructUnionParser.cpp
+++ b/src/tokenizer/EnumStructUnionParser.cpp
@@ -1000,8 +1000,9 @@ void EnumStructUnionParser::analyze_identifiers()
    try_post_identify_macro_calls();
 
    if (  m_start->IsClassOrStruct()
-      && (  m_start->IsNot(E_Token::CT_STRUCT)
-         || !language_is_set(lang_flag_e::LANG_C)))
+      && (  language_is_set(lang_flag_e::LANG_CPP)
+         || language_is_set(lang_flag_e::LANG_CS)
+         || language_is_set(lang_flag_e::LANG_VALA)))
    {
       /**
        * if a type has been identified, mark any constructor matching constructor


micheleCTDEAdmin:
The fix looks ok, but how about the D, OC languages? They can also have structs.

geimer:
According to my (very limited) understanding of Objective-C, structs are the same as in C.  Classes and interfaces are defined differently.  Thus the fix doesn't apply, I believe.  I have no idea about D, though.

micheleCTDEAdmin (approved):
Ok, let's merge as is and if some users report issues with other languages, we will amend as needed.

PR #4727, opened 2026-05-20
Add span_num_{empty,pp,cmt}_lines budget options for align_braced_init_list_span

ericch1:
Hi Uncrustify team,

This PR adds three new line-filter ("budget") options to the `align_braced_init_list_span`
alignment.

Follows the same pattern as PR #4675 (global span filters) and the prior
budget-option PRs.

## New options

```
align_braced_init_list_span_num_empty_lines  (signed, default -1)
align_braced_init_list_span_num_pp_lines     (signed, default -1)
align_braced_init_list_span_num_cmt_lines    (signed, default -1)
```

Each follows the same -1/0/≥1 priority logic:
- `-1`: inherit from `global_span_num_empty/pp/cmt_lines`
- ` 0`: disable budget (no lines of this type are skipped)
- `≥1`: skip up to this many lines of that type between candidates

## Example

Before (braced init lists broken by a blank line or PP directive):

```cpp
// With align_braced_init_list_span = 1 only
class Config {
int m_width { 800 };       // <- not aligned with m_height

int m_height { 600 };      // <- blank line breaks the span
};
```

After (with `align_braced_init_list_span_num_empty_lines = 1`):

```cpp
class Config {
int m_width  { 800 };

int m_height { 600 };      // <- aligned: blank line is skipped (budget=1)
};
```

About this work: The first ideas came from the Claude Sonnet 4.6 AI. But I spent time checking, cleaning, and making it much better before sharing it.

I hope this helps! Please tell me if anything needs changes. Thank you!

PR #4754, opened 2026-07-13
Add align_pp_define_span_num_{empty,pp,cmt}_lines options

ericch1:
Hi Uncrustify team,

This PR extends the `*_span_num_{empty,pp,cmt}_lines` budget mechanism to
preprocessor define alignment (`align_pp_define_span`).

This is the last of the series of PRs that have progressively added per-alignment-family line-filter
options to every *existing* span that can benefits from them (#4586, #4587, #4610, #4664, #4698, #4701, #4714, #4727 and #4733)

### New options

- `align_pp_define_span_num_empty_lines` (signed, default -1)
- `align_pp_define_span_num_pp_lines`    (signed, default -1)
- `align_pp_define_span_num_cmt_lines`   (signed, default -1)

### How it works

Same semantics as every other `*_span_num_*_lines` option:

| Value | Effect  | Description                                               |
|-------|---------|-----------------------------------------------------------|
| `-1`  | INHERIT | Use the corresponding `global_span_num_*_lines` value     |
|  `0`  | DISABLE | Explicitly disable filtering (overrides the global)       |
| `>0`  | OVERRIDE| Use this budget, ignoring the global                      |

### Implementation notes

The original `align_preprocessor()` loop called `pc->GetNextNc()` to skip
past comments.  This made it impossible to count comment lines individually.
The loop now advances token-by-token with `pc->GetNext()` and processes
`CT_NEWLINE` and comment chunks through `GetNlCountFiltered()`, exactly as
other recently-updated alignment families do.  The budget is reset
(`skip_budget = skip_cfg`) each time a `#define` candidate is added to the
alignment stack, so each pair of adjacent defines starts with a fresh budget.

Both the value-macro stack (`as`) and the function-macro stack (`asf`) share
the same budget, consistent with the fact that they share the same span.

A new `pp_excludes_define` flag has been added to `LineSkipConfig`.
When set (as it is in the preprocessor aligner), `CT_PP_DEFINE` tokens are
**not** counted as PP lines in `GetNlCountFiltered()`.  This is necessary
because `CT_PP_DEFINE` is also `IsPreproc()`, so without this flag the
newline immediately before any `#define` would consume one unit of the
`pp_lines` budget — making the effective budget for non-`#define` PP
directives (`#ifdef`, `#endif`, `#pragma`, …) one less than the configured
value.  The flag defaults to `false` for all other alignment families, so
their behaviour is unchanged.

### Behavioral change

Switching from `GetNextNc()` to `GetNext()` has a side effect on **multi-line
block comments** (`/* ... */` spanning several lines) that appear between
`#define` groups when `align_pp_define_span > 0`.

Previously only the two surrounding newlines (before and after the comment
block) were counted toward the span; the internal newlines of the comment were
invisible to the span counter.  Now those internal newlines are also counted.

Impact by case:

| Situation | Behavior |
|-----------|----------|
| `align_pp_define_span = 0` (the default) | **No change** — alignment is disabled |
| Single-line comments (`//` or `/* */` on one line) | **No change** — `GetNlCount()` returns 0 for such chunks |
| Multi-line `/* ... */` with `align_pp_define_span > 0` | **Changed** — internal newlines now count toward the span |

Users who relied on multi-line block comments being "transparent" to the span
counter can restore the previous behaviour by setting
`align_pp_define_span_num_cmt_lines` to an appropriate budget, or by raising
the span value.

### Tests

36 new tests: 18 for C, 18 for C++.

Config files (all in `tests/config/common/`):
- `align_pp_define_span_num_empty-{1,2,3}.cfg` — empty-line budget 1, 2, 3
- `align_pp_define_span_num_pp-{1,2,3}.cfg`    — PP-line budget 1, 2, 3
- `align_pp_define_span_num_cmt-{1,2,3}.cfg`   — comment-line budget 1, 2, 3
- `align_pp_define_span_num_mixed-{1..6}.cfg`  — combined budgets
- `global_align_span_pp_define_num_lines-{1,2,3}.cfg` — global fallback,
  global+override-0, and global+specific-override

Input files cover four interruptor scenarios for both value macros and
function-like macros:
1. Empty lines between `#define` groups
2. Comment lines (C-style `/* */` in `.c`, C++ `//` in `.cpp`)
3. Non-`#define` preprocessor lines (`#ifdef`/`#endif`)
4. Mixed interruptors (empty + PP + comment)

### Backward compatibility

**Mostly backward compatible**, with one edge case:

- When `align_pp_define_span = 0` (the default), nothing changes.
- When `align_pp_define_span > 0` and only **single-line comments** appear
  between `#define` groups, nothing changes.
- When `align_pp_define_span > 0` and **multi-line block comments** (`/* ... */`)
  appear between `#define` groups, the effective line count is now higher because
  the comment's internal newlines are counted (they were previously invisible).
  Affected users can use `align_pp_define_span_num_cmt_lines` to absorb those
  lines, or simply raise the span value.

The 3 new options themselves default to `-1` (inherit from global), and the
global options default to `0` (no filtering), so no new filtering is active
unless explicitly configured.

About this work: The first ideas came from the Claude Sonnet 4.6 AI. But I (hopefully) spent a lot of time checking, cleaning, and making it much better before sharing it.


gmaurel (changes requested):
Hello ericch1!

Your explanations are very fine.
It would be sorry to "lost" them as a "comment about 4754".

I wish, you modify the text a little bit to get a pretty documentation,
which should be pushed under <uncrustify>/documentation/some_thing.txt

It is not a "wish" but a request, so we keep a trace of this work.

Thanks
guy

ericch1:
> Hello ericch1!
> 
> Your explanations are very fine. It would be sorry to "lost" them as a "comment about 4754".
> 
> I wish, you modify the text a little bit to get a pretty documentation, which should be pushed under /documentation/some_thing.txt
> 
> It is not a "wish" but a request, so we keep a trace of this work.
> 
Ok, no problem.  Just tell me exactly which explanation above you want me to put into a document since there is already [align-span-line-filters.txt](https://github.com/ericch1/uncrustify/blob/508ccfb5d6d24eb1e724403a352bd84eb2f23144/documentation/align-span-line-filters.txt
) I did and updated into the actual PR.

Maybe you want that I add the behavior change explanations above into that document or you had something else in mind?

Thanks,
Eric

micheleCTDE:
@ericch1 Hi Eric, I think @gmaurel means the part about the mostly compatible behavior of `align_pp_define_span`?

gmaurel:
I mean the part of the conversation which has many details about your work.
It would be fine to save all that.
Please create one/some files for the <uncrustify>documentation

Thanks

micheleCTDEAdmin (approved):
Let's merge this. we can always add more documentation in future as needed.

micheleCTDEAdmin:
@gmaurel need your approval to proceed.

ericch1:
Bonjour, je reviens tout juste de vacances.  Je veux bien faire un fichier avec les commentaires que vous souhaitez conserver.

Je propose d'ajouter les commentaires des sections `Implementation notes` et `Behavioral change` dans le fichier `documentation/align-span-line-filters.txt` comme une note sur ces changements qui deviendront "historiques".  Notez qu'on retrouve pas mal ces commentaires dans les messages de commits aussi.

Si vous préférez un fichier à part ou d'autres choses dedans, n'hésitez pas, je ferai les modifs...

Merci,
Eric

ericch1:
Voilà, c'est une proposition, si ça ne rencontre pas vos attentes, dites-moi et je refais.
a+

ericch1:
I just rebased and renumbered the tests to resolve the conflict...

micheleCTDEAdmin:
I went ahead and merge anyway. The PR is good and passes all the tests.

ericch1:
Thank you :) !

I'm planning the next new features soon...

PR #4755, opened 2026-07-17
Add support to nl_after_c_comment

acassis:
Hi Uncrustify team,

I think uncrustify is the only lint/formatter able to fix almost 95% of source code to fit the NuttX coding style:
https://nuttx.apache.org/docs/latest/components/tools/uncrustify.html

But it doesn't support to add a new line after a comment line:

/* This is a single line comment */
bool code_need_empty_line_after_comment = true;

shold be converted to:

/* This is a single line comment */

bool code_need_empty_line_after_comment = true;

### DIFF
--- a/src/options.h
+++ b/src/options.h
@@ -3129,6 +3129,10 @@ nl_before_c_comment;
 extern BoundedOption<unsigned, 0, 16>
 nl_before_cpp_comment;
 
+// Whether to force a newline after a single-line comment.
+extern Option<bool>
+nl_after_c_comment;
+
 // Whether to force a newline after a multi-line comment.
 extern Option<bool>
 nl_after_multiline_comment;
--- a/src/newlines/after.cpp
+++ b/src/newlines/after.cpp
@@ -28,6 +28,52 @@ void newline_after_label_colon()
    }
 } // newline_after_label_colon
 
+void newline_after_singleline_comment()
+{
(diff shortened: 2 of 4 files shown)


src/options.h
```diff
@@ -3129,6 +3129,11 @@ nl_before_c_comment;
 extern BoundedOption<unsigned, 0, 16>
 nl_before_cpp_comment;
 
+// The minimum number of newlines after a C comment.
+// Doesn't apply if after a brace open or other single-line C comments.
+extern BoundedOption<unsigned, 0, 16>
```
acassis: I added BoundedOption, but I'm not complaint with it, I will change to bool Option

acassis:
ping @gmaurel

micheleCTDEAdmin:
@acassis please fix the failing tests. Also note that when you add a new option, there are a number of files that need to be updated. You can look where other options are used and do the same for the new option. Alternatively you can use  the `optiondocs` of the `release_tool.py` script to do that.

gmaurel:
@acassis 
What about theproposal?

acassis:
> @acassis What about theproposal?

Hi @gmaurel basically this: https://nuttx.apache.org/docs/latest/contributing/coding_style.html#comments

We need a rule to add a NL after a single comment. For multi-line comments there is already a rule to force a NL, but not for single line.

PR #4757, opened 2026-08-06
Fix declaration paren indentation in templates

Linvail:
Do not apply indent_paren_after_func_decl when a function-like parenthesis is nested directly inside a template argument list. This prevents member-function-pointer partial specializations from receiving an extra indentation level on the first formatting pass.

Add a regression case that starts from the single-line specialization syntax and verifies that formatting reaches the stable multiline result in one pass while preserving ordinary declaration indentation.

Change-Id: I33f68fa0e1c7cc77e5e6a35744324e6ae74d9c65


micheleCTDEAdmin (approved):
Looks good

PR #4761, opened 2026-08-20
Speed up two no-op scans during formatting

adamjernst:
Avoid two expensive scans that almost never change the output.

AlignStack::Flush() only uses the preceding typedef lookup when the typedef is on the same source line as the aligned token, but GetPrevType() walks arbitrarily far backward through the token list to find it. Stop the search at the start of the current line instead.

newlines_cleanup_braces() unconditionally calls newline_var_def_blk() over the whole file, but that function only changes anything when one of nl_var_def_blk_end_func_top, nl_var_def_blk_start, nl_var_def_blk_end or nl_var_def_blk_in is non-zero. All four default to 0, so skip the scan when they are all disabled.

Measured on a large file from the Meta monorepo, median runtime dropped from 0.5465s to 0.1770s with identical output. All tests pass.


src/newlines/cleanup.cpp
```diff
@@ -1271,7 +1271,20 @@ void newlines_cleanup_braces(bool first)
       }
    }
 
-   newline_var_def_blk(Chunk::GetHead());
+   // newline_var_def_blk() is a no-op unless one of these is enabled, so skip
+   // the full-file scan when they are all disabled
+   log_rule_B("nl_var_def_blk_end_func_top");
+   log_rule_B("nl_var_def_blk_start");
+   log_rule_B("nl_var_def_blk_end");
+   log_rule_B("nl_var_def_blk_in");
```
micheleCTDEAdmin: Suggest to move this 4 lines inside the following `if` body, so they are only printed if one of them is non zero and we call `newline_var_def_blk`

micheleCTDEAdmin:
What if we have a multiline typedef for any reason? would it still work ok? I am thinking of something like
```
typedef struct foo {
  /* fields */
} bar;
```

src/align/stack.cpp
```diff
@@ -410,15 +410,21 @@ void AlignStack::Flush()
    if (Len() == 1)
```
adamjernst, line 410: Note here that this only applies for single-line typedefs.

adamjernst:
> What if we have a multiline typedef for any reason? 

Nah, note that this code only applies to single-line typedefs. Added a test covering this.

> Suggest to move this 4 lines inside the following if body

Done.

src/align/stack.cpp
micheleCTDEAdmin: great, thanks for pointing this out.

micheleCTDEAdmin (approved):
Looks good.

PR #4762, opened 2026-08-21
Stop recompiling import-sorting regexes for every file and comparison

adamjernst:
Import sorting was the single largest source of std::regex compilation in profiles of an Objective-C workload, in two places:

prepare_categories() built a fresh std::regex for each include_category_N option at the start of every file, and cleanup_categories() destroyed them again at the end. Keep the compiled regexes between files and only rebuild one when the option string it came from has changed.

text_contains_filename_without_ext() compiled two regexes on every call that missed the cache: one to escape the filename, and one for the pattern '\S?<filename>\b.*' that it then ran through std::regex_match(). That pattern can only match with the filename at offset 0, or at offset 1 behind a single non-space character, so test those two positions with a plain string compare plus a word-boundary check.

Neither the sort options nor the configured categories change. A differential test of the old regex against the replacement over 1908 filename/text pairs, covering regex metacharacters in filenames, empty filenames, extensionless paths and leading whitespace, reports no behavior difference, and formatting 941 files in a single process produces byte-identical output.

Measured on a large Objective-C corpus: 93.592s before, 79.209s after replacing the filename regex, 60.667s after also caching the category regexes. std::__detail::__compile_nfa no longer appears in the profile.

### DIFF
--- a/src/sorting.cpp
+++ b/src/sorting.cpp
@@ -13,9 +13,12 @@
 #include "mark_change.h"
 #include "newlines/add.h"
 #include "newlines/double_newline.h"
+#include "unc_ctype.h"
 
 #include <algorithm>
 #include <cstdio>  // to get fprintf
+#include <cstring>
+#include <memory>
 #include <regex>
 #include <string>
 #include <unordered_map>
@@ -42,14 +45,16 @@ constexpr static int kIncludeCategoriesCount = 3;
 struct include_category
 {
    include_category(const std::string &pattern)
-      : regex(pattern)
+      : pattern(pattern)
+      , regex(pattern)
    {
    }
-   std::regex regex;
+   std::string pattern;
+   std::regex  regex;
 };
 
 
-include_category                      *include_categories[kIncludeCategoriesCount];
+std::unique_ptr<include_category>     include_categories[kIncludeCategoriesCount];
 std::unordered_map<Chunk *, int>      chunk_priority_cache;
 std::unordered_map<std::string, bool> filename_without_ext_cache;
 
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin:
please amend in order to pass all actions.

adamjernst:
Fixed the failures, and also refactored to avoid adding globals.

src/sorting.cpp
```diff
@@ -134,39 +137,93 @@ static int get_chunk_priority(Chunk *pc)
...
+//! Matches the character class the regex '\w' stands for.
+static bool is_regex_word_char(char ch)
+{
+   return(  (  ch >= 'a'
+            && ch <= 'z')
+         || (  ch >= 'A'
+            && ch <= 'Z')
+         || (  ch >= '0'
+            && ch <= '9')
+         || ch == '_');
```
micheleCTDEAdmin, line 175: `isalnum(ch) || ch == '_'` looks better :-)

src/sorting.cpp
```diff
@@ -134,39 +137,93 @@ static int get_chunk_priority(Chunk *pc)
...
+}
+
+
+/**
+ * Returns true if the text contains filename without extension.
+ *
+ * This used to build the equivalent regex '\S?<filename>\b.*' and run
+ * std::regex_match() against it, which meant escaping the filename and
+ * compiling two regexes on every call. The pattern only allows the filename at
+ * offset 0, or at offset 1 behind a single non-space character, so match those
```
micheleCTDEAdmin: Since we are reworking this logic, would it make sense to remove the 0 or 1 offset? I am not really sure what the original logic intended to do with that offset. Any idea?

adamjernst: Updated the comment to indicate the reason for the 0-or-1 offset. (It is necessary, it turns out!)

src/sorting.cpp
adamjernst: done!

micheleCTDEAdmin:
> for a preprocessor include, the include target with its opening delimiter still attached and its closing one dropped ('"foo.h', '<foo.h')
 
Thanks for the explanation. In such case I think we should fix `chunk_sort_str()` and strip the leading delimiter, so that it returns 'foo.h' only. Then we don't need to worry about offset 0 or 1 and we can simplify the code.
Would it be ok for you to look into it?

adamjernst:
@micheleCTDEAdmin sure, #4786 implements that option in a separate commit db40b5e7ba79631c61f51ff0937c8493f051f500 and then reimplements this change on top of it. If you merge #4786 you can close this PR. I left it open for now, so you can choose between the two options.

PR #4763, opened 2026-08-21
Skip rule-logging work when the log severity is off

adamjernst:
log_rule_B() is invoked from hot loops throughout the formatter. log_rule3() normalized the function name with get_unqualified_func_name() before handing the message to LOG_FMT(), which is where the severity is actually tested, so ordinary formatting runs paid for debug-only string work on every call.

Test the severity first and return early. Because the check has already been made, the LOG_FMT() calls become direct log_fmt() calls.

Measured on a large Objective-C++ corpus, where log_rule3() accounted for 2.29% of self time before the change, total wall time dropped from 233.500s to 229.602s.


micheleCTDEAdmin (approved):
Looks good, thanks.

PR #4764, opened 2026-08-21
Skip the class-colon newline pass when it cannot change anything

adamjernst:
newlines_class_colon_pos() walks every chunk in the file for both CT_CLASS_COLON and CT_CONSTR_COLON, but each edit it can make is gated on an option: the colon is only repositioned when pos_class_colon or pos_constr_colon selects LEAD or TRAIL, newlines are only added or removed when nl_class_colon/nl_constr_colon or nl_class_init_args/nl_constr_init_args is not IGNORE, and the alignment stack is only fed when align_constr_value_span is enabled for C++. pos_class_comma and pos_constr_comma only choose between branches nl_*_init_args has already enabled, so they cannot enable the pass on their own.

Return before the scan when none of those apply. Configurations that leave the class-colon options alone no longer pay for two full-file walks.

Measured on a large Objective-C++ corpus, where newlines_class_colon_pos() accounted for 2.60% of self time before the change, total wall time dropped from 229.602s to 223.653s.

### DIFF
--- a/src/newlines/class_colon_pos.cpp
+++ b/src/newlines/class_colon_pos.cpp
@@ -67,6 +67,20 @@ void newlines_class_colon_pos(E_Token tok)
    LineSkipConfig acv_skip_budget = {};
    Chunk const    *prev_acv_chunk = Chunk::NullChunkPtr;
 
+   // Every edit the loop below can make is gated on one of these: the colon is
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin (approved):
Nice finding. Thanks.

PR #4766, opened 2026-08-21
Bound the assignment search in mark_function()

adamjernst:
When mark_function() reaches a CT_FUNCTION it has not otherwise classified, it looks ahead for an assignment to decide between CT_FUNC_CALL and CT_FUNC_DEF (Issue #575, `x->operator [](0) = ...`). The lookahead used GetNextType(CT_ASSIGN, level), which has no stopping point, so every unresolved function token could scan through the whole rest of the file looking for an assignment that belongs to some unrelated later statement.

Only an assignment within the current statement is meaningful here, so stop the search at a same-level semicolon or brace. The Issue #575 case is unaffected: its assignment sits between the operator call and the statement's semicolon.

Measured on a large Objective-C++ corpus, where mark_function() accounted for about 8% of self time, total wall time dropped from 252.304s to 234.368s.

### DIFF
--- a/src/tokenizer/combine_fix_mark.cpp
+++ b/src/tokenizer/combine_fix_mark.cpp
@@ -1196,6 +1196,48 @@ void mark_function_return_type(Chunk const *fname, Chunk *start, E_Token parent_
 } // mark_function_return_type
 
 
+/**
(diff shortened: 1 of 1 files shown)


micheleCTDEAdmin (approved):
Good improvement, thanks.

PR #4768, opened 2026-08-23
ci: Fix cache

mdeweerd:
The 'cppcheck' hook of  https://github.com/pocc/pre-commit-hooks is not ok for caching, which I fixed in fork  https://github.com/mdeweerd/pre-commit-hooks .

This PR uses that fork and fixes the cache restoration order in the github action.

With cache:
image

Without cache:
image


micheleCTDEAdmin (approved):
Thanks, this should speed up things quite a bit :-)

Issue #1620, opened 2018-02-15
Array elements don’t get indented when passing as named argument


KalyanaChakravarthi:
```diff --git a/UNI-24090.cs b/test.out.cs
index 1830dea4..f33a2248 100644
--- a/UNI-24090.cs
+++ b/test.out.cs
@@ -1,28 +1,27 @@
 public void GeneratesCorrectVisualStudioProjectFile()
 {
-       GenerateProjectsAndCompareWithTemplates(
-               runInJam: InJamCreateTestProject,
-               generatedPath: TestRoot.Combine("Solution"),
-               templatesPath: "Tools/Unity.BuildSystem/Unity.BuildSystem.VisualStudio.Tests/Templates",
-               templates: new[] {
-                    "TestProjectGeneration_CApplication.sln",
-                    "Projects/TestProjectGeneration_CApplication.vcxproj",
-                    "Projects/TestProjectGeneration_CApplication.vcxproj.filters"
-       });
+    GenerateProjectsAndCompareWithTemplates(
+        runInJam: InJamCreateTestProject,
+        generatedPath: TestRoot.Combine("Solution"),
+        templatesPath: "Tools/Unity.BuildSystem/Unity.BuildSystem.VisualStudio.Tests/Templates",
+        templates: new[] {
+        "TestProjectGeneration_CApplication.sln",
+        "Projects/TestProjectGeneration_CApplication.vcxproj",
+        "Projects/TestProjectGeneration_CApplication.vcxproj.filters"
+    });
 }
```
[UNI-24090.cs](https://github.com/uncrustify/uncrustify/files/1727157/UNI-24090.cs.txt)

satishp87:
In unity fork this issue is addressed with below PR
Unity-Technologies#100

guy-maurel:
Thanks to @KalyanaChakravarthi

gmaurel:
Fixed

PR #4770, opened 2026-08-24
Fix braced initializer indentation without language gating

adamjernst:
Issue #4556 is a C++ reproducer that hangs while formatting a multiline braced initializer passed as a call argument. The affected branch subtracts indent_size from ParenStackEntry::GetIndent(), both of which are unsigned. When the previous indent is zero, the subtraction wraps and formatting produces unbounded indentation output.

PR #4600, commit 3084d0599d1e62e4483c4259c495994b468e0322 (merged as c21b7ff4595cb560cebd3b4003620ec02e5a5f09), avoided the failing C++ path by restricting it to C#. That restriction reflects the origin of the branch in C# issue #1620, but it is not the actual safety requirement. It also changed established C++ and Objective-C++ formatting whenever the subtraction would have been valid.

PR #4580 previously guarded GetIndent() != 0 inside the selected branch. Besides not covering positive values smaller than indent_size, that consumed the branch without allowing the remaining indentation cases to run. PR #4600 instead let C++ fall through, but did so for every C++ and Objective-C++ input.

Gate the branch on GetIndent() >= indent_size. Unsafe states now fall through to the normal indentation logic used by the #4556 fixture, while safe states retain the pre-0.83 anti-double-indent behavior in every applicable language.

Removing the C# gate without this check reproduced the #4556 hang. With this check, the existing #4556 fixture completes and matches its current golden output, while representative C++ and Objective-C++ call-argument initializers retain their prior indentation. Add regression cases for both languages; they fail against unmodified 0.83 with the extra indentation introduced by #4600.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1945,7 +1945,7 @@ void indent_text()
             }
             // Issue # 1620, UNI-24090.cs
             else if (  frm.prev().GetOpenChunk()->IsOnSameLine(frm.top().GetOpenChunk())
-                    && language_is_set(lang_flag_e::LANG_CS)             // Issue 4556
+                    && frm.prev().GetIndent() >= indent_size
                     && !options::indent_align_paren()
                     && frm.prev().GetOpenChunk()->IsParenOpen()
                     && !pc->TestFlags(PCF_ONE_LINER))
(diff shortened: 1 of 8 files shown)


adamjernst:
@gmaurel If this should truly be C# only let me know.

micheleCTDEAdmin (approved):
LGTM

PR #4771, opened 2026-08-25
Keep boolean && before sizeof expressions

adamjernst:
When a parenthesized expression places && after a template-id and before sizeof or decltype, the tokenizer mistakes the operator for an rvalue reference. This causes reference-spacing options to format boolean expressions incorrectly.

Recognize those expression starters as evidence that && remains boolean, and add a regression test for the sizeof case.


micheleCTDEAdmin (approved):
LGTM

PR #4772, opened 2026-08-25
Preserve lambda rvalue-reference parameters

adamjernst:
Objective-C++ lambda parameters whose template type ends before && can be mistaken for boolean expressions. This applies boolean spacing and separates the reference operator from the parameter name.

Recognize the surrounding lambda capture, parameter, and body delimiters while excluding default-argument expressions, so && remains an rvalue reference.


micheleCTDEAdmin:
Please resolve conflicts

micheleCTDEAdmin (approved):
Looks good

PR #4773, opened 2026-08-26
Bound preprocessor-skipping parameter scans

adamjernst:
Constructor-variable detection scans function parameters while skipping preprocessor directives. For function-like macro aliases, that skip can move past the expected closing parenthesis, leaving the loop to advance forever on the null chunk.

Stop the scan at the null chunk and only perform constructor-variable classification when the closing parenthesis was reached. Otherwise, fall through to the existing prototype handling.


micheleCTDEAdmin:
Please resolve conflicts

micheleCTDEAdmin:
New conflicts to resolve

Issue #4767, opened 2026-08-22
[BUG]: uncrustify doesn't align inside typedef struct (and probably other typedefs like enum etc)


levicki:
When Uncrustify encounters a `typedef`, it seems to flag the entire block (whether it encloses a `struct`, `enum`, or `union`) under a generic `typedef` parsing context. Instead of treating the contents inside the braces `{ ... }` as an isolated scope with standard variable, function, or enumerator alignment rules, it treats the entire block as part of a single compound type-definition statement ignoring the scope content except for right side comments.

**Original**
```cpp
typedef struct IMUIShell
{
	/* IMUIObject slots (B2 prefix): */
	uint32_t (*AddRef)(IMUIObject *Self);
	uint32_t (*Release)(IMUIObject *Self);
	uint32_t (*Version)(IMUIObject *Self);
	uint32_t (*IfaceId)(IMUIObject *Self);
	IMUIObject *(*QueryIface)(IMUIObject *Self, uint32_t IFaceId);
	/* shell-specific: */
	HRESULT (*Initialize)(IMUIShell *Self);							/* COM + window + RT; one-way */
	int (*Run)(IMUIShell *Self);								/* blocks; returns exit code  */
	void (*Shutdown)(IMUIShell *Self);							/* idempotent finalize        */
	IMUIVisual *(*Root)(IMUIShell *Self);							/* AddRef'd; caller Releases  */
	HRESULT (*SetThemeMode)(IMUIShell *Self, uint32_t Mode);				/* MUITheme */
	HRESULT (*SetOnDeviceLost)(IMUIShell *Self, MUIFnReason Handler, void *UserCtx);	/* MUIReason payload: ReasonDevice* + Attempt */
} IMUIShell;
```
**Desired format**
```cpp
typedef struct IMUIShell
{
	/* IMUIObject slots (B2 prefix): */
	uint32_t	(*AddRef)(IMUIObject *Self);
	uint32_t	(*Release)(IMUIObject *Self);
	uint32_t	(*Version)(IMUIObject *Self);
	uint32_t	(*IfaceId)(IMUIObject *Self);
	IMUIObject	*(*QueryIface)(IMUIObject *Self, uint32_t IFaceId);
	/* shell-specific: */
	HRESULT		(*Initialize)(IMUIShell *Self);							/* COM + window + RT; one-way */
	int		(*Run)(IMUIShell *Self);							/* blocks; returns exit code  */
[...]

PR #4776, opened 2026-09-01
Fix issue #4767: don't propagate PCF_IN_TYPEDEF into nested struct/union/enum/class bodies

levicki:
For cotnext see [the issue I reported](https://github.com/uncrustify/uncrustify/issues/4767).

This fix has been co-authored with Claude.

`ix_typedef()` stamped `PCF_IN_TYPEDEF` on every chunk inside `typedef struct/union/enum {...} X;`, including member declarations nested one level deeper. `mark_function_type()` used that flag to misclassify function-pointer struct members as type aliases instead of variables, suppressing `PCF_VAR_1ST_DEF` and silently breaking `align_var_def_brace()` column alignment for `TYPE (*NAME)(ARGS);` members -- while plain members and right-side comments kept aligning correctly via unrelated passes, hiding the defect.

The typedef marking loop now skips over a nested class/enum/struct/ union body's member declarations (keeping the flag on the body's own delimiting braces, since sp_brace_typedef and friends rely on it).

Also bump version to `0.84.0` and add regression tests for the function-pointer-only, mixed plain/function-pointer, and union variants of the bug, plus an enum non-regression guard.

### DIFF
--- a/src/tokenizer/combine.cpp
+++ b/src/tokenizer/combine.cpp
@@ -627,7 +627,9 @@ static bool handle_rvalue_angle_close(Chunk const *prev, Chunk *pc)
       if (  pc->GetLevel() > pc->GetBraceLevel()
          && (  next->Is(E_Token::CT_WORD)
             || next->Is(E_Token::CT_TYPE)
-            || next->Is(E_Token::CT_DC_MEMBER)))   // ::namespace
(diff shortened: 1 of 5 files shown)


levicki:
I am usually working solo and not used to collaboration and PRs, hence beginner mistake, closed and fixed in new PR.

PR #4779, opened 2026-09-02
Fix for #4777: recognize and skip __declspec(...) keyword when detecting variable definitions

### LINKED ISSUE #4777: [BUG]: Uncrustify doesn't align variables after __declspec(...) keyword
As the title says, for this code snippet:
```cpp
#define EXPORT __declspec(dllexport)
extern "C" {
	EXPORT DWORD NvOptimusEnablement1 = 1;
	EXPORT DWORD NvOptimusEnablement2 = 3;
	__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
}
```
I expect (with tab size 8):
```cpp
#define EXPORT __declspec(dllexport)
extern "C" {
	EXPORT DWORD			NvOptimusEnablement1 = 1;
	EXPORT DWORD			NvOptimusEnablement2 = 3;
	__declspec(dllexport) int	AmdPowerXpressRequestHighPerformance = 1;
}
```
Actual result with 0.83.0 is:
```cpp
#define EXPORT __declspec(dllexport)
extern "C" {
	EXPORT DWORD	NvOptimusEnablement1 = 1;
	EXPORT DWORD	NvOptimusEnablement2 = 3;

	__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
}
```
Note that this is two issues caused by the same underlying mistake &mdash; current `__declspec(...)` handling prevents last variable definition from being recognized as a part of `var_def_blk` and:

 - Separates it with newlines (I configured it like that).
 - Doesn't align word after type with the rest.

I already have a fix for this, so I am just documenting the bug.

levicki:
Fixes #4777.

 - Recognized CT_DECLSPEC as a valid statement-start token in fix_symbols()'s variable-definition gate (combine.cpp) so fix_variable_definition() is actually invoked for declarations prefixed with __declspec(...).
 - Made fix_variable_definition()'s scan loop (combine_fix_mark.cpp) skip a leading __declspec(...) specifier via the existing skip_declspec_next() helper before continuing to scan for the real type/name run.
 - Added a matching CT_DECLSPEC skip to newline_var_def_blk()'s own "skip qualifiers" step (src/newlines/var_def_blk.cpp), via the same skip_declspec_next() helper, mirroring its existing CT_QUALIFIER handling.
 - Added 5 regression tests (cpp/declspec_extern_c.cpp, cpp/declspec_single.cpp, cpp/declspec_func_decl.cpp, cpp/declspec_class_namespace.cpp, cpp/declspec_mixed_block.cpp) registered as cpp.test 60180-60184, with a new tests/config/cpp/declspec_var_def.cfg.
 
**Note:** Fix co-authored with Claude

### DIFF
--- a/src/newlines/var_def_blk.cpp
+++ b/src/newlines/var_def_blk.cpp
@@ -12,6 +12,7 @@
 #include "newlines/is_func_call_or_def.h"
 #include "newlines/is_var_def.h"
 #include "newlines/min_after.h"
+#include "tokenizer/combine_skip.h"
 
 
 using namespace uncrustify;
@@ -71,6 +72,14 @@ Chunk *newline_var_def_blk(Chunk *start)
       if (pc->Is(E_Token::CT_QUALIFIER))
       {
          pc = pc->GetNext();
+         continue;
+      }
+
+      // skip '__declspec(...)' specifiers -- like a qualifier, they aren't
+      // themselves part of the variable definition that follows
+      if (pc->Is(E_Token::CT_DECLSPEC))
+      {
+         pc = skip_declspec_next(pc);
          continue;
       }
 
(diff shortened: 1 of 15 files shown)


levicki:
This one it turns out is incomplete. I will redo it.

PR #4781, opened 2026-09-02
Fix for #4780: add VERSIONINFO for WIN32, match official binary for MSVC, semver+hash

### LINKED ISSUE #4780: [MISC]: Windows MSVC build, multiple issues
### 1. Missing `VERSIONINFO` resource
It is customary for Windows executables to be have an embedded [VERSIONINFO resource](https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource) in the executable so users are able to inspect the product name, description, and product / file version, from Windows Explorer and using PowerShell commands without actually running the executable.

### 2. Version string doesn't contain short commit hash
This is making it hard to identify exact source tree from which executable was built. I am aware that current system builds from archives and since those have no `git` tree it uses version fallbacks, but for people who do build from git tree having the short commit hash in the version string would be genuinely useful info.

### 3. CMake MSVC build
 - does not produce by default clean release executable without `.debug` section.
 - does not produce by default statically linked executable like Linux or Windows mingw cross-compiling does.
 - embeds a manifest by default which isn't present in official build and is not necessary

I already have a fix for this ready, I am just documenting the issues.

levicki:
Fixes #4780.

 - Added src/uncrustify.rc with FileDescription/ProductName/ProductVersion for Windows builds, wired into the executable target on WIN32 only.
 - Defined _WIN64 for the resource compiler on 64-bit builds, since only cl.exe auto-defines it, not rc.exe/windres.
 - Derived VER_MAJOR/VER_MINOR/VER_PATCH/VER_STRING from UNCRUSTIFY_VERSION in both the git and NoGitVersionString paths, and failed the CMake configure/build if UNCRUSTIFY_VERSION doesn't match major.minor.patch.
 - Appended the short commit hash to UNCRUSTIFY_VERSION and VER_STRING when git is available.
 - Switched MSVC builds to the static CRT (/MT, /MTd) to match the statically-linked official binaries.
 - Disabled manifest generation for MSVC builds (/MANIFEST:NO) to match the official MinGW-linked binaries, which embed none.
 - Stripped debug information from MSVC Release builds and passed /debug:none plus the undocumented /EMITPOGOPHASEINFO link.exe switch to the Release linker.
 
 **Note:** Fix co-authored with Claude

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,25 @@ endif()
 if(MSVC)
   add_definitions(/D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_WARNINGS /wd4267)
   add_definitions(/utf-8)
+
+  # Match the official MinGW-w64 binaries, which are statically linked
+  # against the CRT (no runtime DLL dependency).
+  foreach(flag_var
+      CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
+      CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO
+      CMAKE_C_FLAGS CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
+      CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELWITHDEBINFO)
+    string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
+  endforeach()
+
+  # Match the official MinGW-w64 binaries, which embed no manifest.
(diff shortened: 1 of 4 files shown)


micheleCTDEAdmin:
@levicki did you also try to cross build Windows executables in linux?

levicki:
> @levicki did you also try to cross build Windows executables in linux?

No, I didn't have access to that, I did try building with mingw which ships with Strawberry Perl on Windows and that one passed for me.

**EDIT:** I would appreciate some input on this.

Presently `scripts/make_version.py` overwrites passed in fallback version with short commit hash when building in repo cloned with `--no-tags --depth=1` which is visible in CI runner output:
```txt
-- Version: '0ab774e'
```
Now for pure automation builds that's obviously not important so I could just put `0.0.0` in version resource and call it a day. The change I made now to `GenerateVersionHeader.cmake` preserves version passed into it and uses that instead.

Now that there's fallback, `FATAL_ERROR` part doesn't make sense anymore.

I welcome suggestions. Maybe you prefer fixing this in `scripts/make_version.py` as it clearly has a fallback for builds out of tarball but not out of shallow git clones?

Issue #4782, opened 2026-09-03
[BUG]: uncrustify classifies ^ as OC_BLOCK_CARET after template closing >


levicki:
When for example C++/CLI `array<int>^ GetData() { ... }` gets fed through uncrustify's current matcher, it:

1. Sees the caret → assumes "block literal starts here."
2. Scans forward for parens+braces → finds `()` and `{ }` → confirms the shape, commits.
3. Walks backward from `(` to the caret, grabbing everything in between — that's just the single token `GetData` — and stuffs it into the "optional RTYPE" slot `(make_type(lbp), PCF_OC_RTYPE, parent CT_OC_BLOCK_EXPR)`.
4. The `()` gets flagged and reparsed as the block's own argument-list parens (`flag_parens(..., CT_OC_BLOCK_EXPR) + fix_fcn_def_params)`, not as a named function's parameter list.

Another case where it breaks is in plain C++ (>=14) formatting:
```
int x = Mask<4> ^ [](){ return 1; }();
```
Changes to:
```
int x = Mask<4> ^ [] (){ return 1; } ();
```
Confirmed via `-p` dump: the `^` gets retyped `OC_BLOCK_CARET`, and the lambda's own parameter-list parens and trailing call parens get the same spurious-space treatment as `GetData ()` does. Reproduces with default config settings.

I know that uncrustify currently doesn't support C++/CLI, but I plan to add support so this is worth fixing first. I will propose a PR shorty.

levicki:
After fixing this issue locally I found that the commit `083e919cdaaf3056bc3419b9ce659752bfa67296` by Scott Bilas has introduced an issue which this bug was masking so after fixing it the following:
```
int y = Mask<4> ^ Mask<8>;
```
Becomes:
```
int y = Mask<4> ^Mask<8>;
```
Because that commit misclassifies `^` as `CT_PTR_TYPE` even though it's a aritmetic xor and part of an assignment expression.

I realize fixing the first issue without fixing this previously hidden and now unmasked second one is not an option so am currently checking what would be the best approach.

levicki:
Devised a fix that handles all cases I could come up with properly. I will build my proposal for handling of C++/CLI off of this fix.

PR #4784, opened 2026-09-03
Fix for #4782: don't mistype ^ as block or ptr

levicki:
- Excluded ^ after CT_ANGLE_CLOSE from Obj-C block-literal detection at both call sites in do_symbol_check() (combine.cpp), which corrupted declarations like array<int>^ GetData().
 - Added angle_close_starts_declaration() (combine.cpp), testing PCF_STMT_START to tell real declarations (array<int>^ Name) apart from value-template expressions (Mask<4> ^ Mask<8>) that look identical at that point.
 - Gated the existing 2015 CT_ANGLE_CLOSE -> CT_PTR_TYPE rule (083e919cda) on that check, so XOR against a template value, literal, macro, call, or lambda now types correctly.
 - Added regression tests cpp/Issue_4782_{decl,xor}.cpp (cpp.test 34967-34968), tests/config/cpp/Issue_4782.cfg.

**Note:** Fix co-authored with Claude.

### DIFF
--- a/src/tokenizer/combine.cpp
+++ b/src/tokenizer/combine.cpp
@@ -98,6 +98,54 @@ static void handle_oc_block_type(Chunk *pc);
 
 
 /**
+ * Issue #4782: test whether the template argument list closed by
+ * 'angle_close' (a CT_ANGLE_CLOSE chunk) genuinely begins its enclosing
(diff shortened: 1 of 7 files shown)


micheleCTDEAdmin:
Can you resolve the conflicts?

levicki:
> Can you resolve the conflicts?

Done.

Issue #4450, opened 2025-01-29
Add option to control lambda body indentation based on closing brace position


1143910315:
I noticed that #3765 modified the indentation behavior of lambdas.
But I expect the following code format:
```c++
// case 1
a([]() {
    b();
});
// case 2
c([]() {
        d();
    }
);
// case 3
auto e = []() {
    f();
};
```
When "});" is located on the same line, the lambda is indented only once.
Please help me, I need an option to control this behavior.
These are the options I used for testing.
```option
input_tab_size=4
output_tab_size=4
indent_columns=4
indent_with_tabs=0
indent_func_call_param=true
indent_paren_close=2
indent_align_assign=false
indent_cpp_lambda_body=true
```
When I set indent_cpp_lambda_body to true, cases 2 and 3 are the code formats I want, while case 1 does not match. When I set indent_cpp_lambda_body to false, case 1 is the code format I want, while cases 2 and 3 do not match.

PR #4785, opened 2026-09-07
Add option to align lambda semicolon

1143910315:
Introduce a new boolean option indent_cpp_lambda_align_semicolon and wire it into indent logic for C++ lambdas. When enabled, the formatter will consider aligning a lambda's trailing semicolon with the lambda body indentation. Added a log entry and updated the condition in indent_text to respect the new option.

### DIFF
--- a/src/indent.cpp
+++ b/src/indent.cpp
@@ -1614,6 +1614,7 @@ void indent_text()
                  && pc->GetParentType() == E_Token::CT_CPP_LAMBDA)
          {
             log_rule_B("indent_cpp_lambda_body");
+            log_rule_B("indent_cpp_lambda_align_semicolon");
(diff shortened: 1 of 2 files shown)


1143910315:
#4450

1143910315:
With indent_cpp_lambda_body=true and indent_cpp_lambda_align_semicolon=true, the formatting result is as shown in the code below:
```cpp
// case 1
a([]() {
    b();
});
// case 2
c([]() {
        d();
    }
);
// case 3
c([]() {
        d();
    }, []() {
        d();
    }
);
// case 4
auto e = []() {
    f();
};
```

micheleCTDEAdmin (changes requested):
Please add some test cases. Moreover when a new option is added, there are a number of files that need to be updated. Look at other options for example and add the required changes

PR #4786, opened 2026-09-08
[v2] Stop recompiling import-sorting regexes for every file and comparison

adamjernst:
Alternate version of #4762 which implements reviewer feedback:

> I think we should fix chunk_sort_str() and strip the leading delimiter, so that it returns 'foo.h' only.

See #4762 for the rationale for this change.


adamjernst:
db40b5e7ba79631c61f51ff0937c8493f051f500 implements the changes to `chunk_sort_str`; the regex cache is layered on top.

